PR #4, opened 2018-11-11
do not traceback on concurent runs

xsuchy:
When you run two concurrent `liquidctl status -v` you get traceback with an error. This will just print the error.

Tested only with python3

### DIFF
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -52,6 +52,7 @@ GNU General Public License for more details.
 import inspect
 import itertools
 import logging
+import sys
 import usb
 
 from docopt import docopt
@@ -119,6 +120,7 @@ def _device_get_status(dev, num):
             print('{:<18}    {:>10}  {:<3}'.format(k, v, u))
     except usb.core.USBError as e:
         LOGGER.error(str(e))
+        sys.exit(2)
     finally:
         dev.disconnect()
     print('')
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -52,6 +52,7 @@ GNU General Public License for more details.
 import inspect
 import itertools
 import logging
+import usb
 
 from docopt import docopt
 
@@ -116,6 +117,8 @@ def _device_get_status(dev, num):
         status = dev.get_status()
         for k, v, u in status:
             print('{:<18}    {:>10}  {:<3}'.format(k, v, u))
+    except usb.core.USBError as e:
+        LOGGER.error(str(e))
     finally:
         dev.disconnect()
     print('')


jonasmalacofilho:
Thanks!

Yes, concurrent runs might see errors like:

```
  File "/usr/lib/python3.7/site-packages/usb/core.py", line 975, in read
    intf, ep = self._ctx.setup_request(self, endpoint)
  File "/usr/lib/python3.7/site-packages/usb/core.py", line 102, in wrapper
    return f(self, *args, **kwargs)
  File "/usr/lib/python3.7/site-packages/usb/core.py", line 216, in setup_request
    self.managed_claim_interface(device, intf)
  File "/usr/lib/python3.7/site-packages/usb/core.py", line 102, in wrapper
    return f(self, *args, **kwargs)
  File "/usr/lib/python3.7/site-packages/usb/core.py", line 167, in managed_claim_interface
    self.backend.claim_interface(self.handle, i)
  File "/usr/lib/python3.7/site-packages/usb/backend/libusb1.py", line 811, in claim_interface
    _check(self.lib.libusb_claim_interface(dev_handle.handle, intf))
  File "/usr/lib/python3.7/site-packages/usb/backend/libusb1.py", line 595, in _check
    raise USBError(_strerror(ret), ret, _libusb_errno[ret])
usb.core.USBError: [Errno 16] Resource busy
```

But while the traceback might not be particularly useful, the `Resource busy` error should not be ignored by most interactive users of the CLI (i.e. why were there concurrent calls in the first place?).  Also, keeping the error exit code is super important for [unattained scripts and system services](https://github.com/jonasmalacofilho/dotfiles/tree/master/liquidctl).

I think the issue you found is more general: _we currently lack more advanced error handling and better error messages,_ for anything other than simple usage mistakes.  But, in its current state, this PR would not help much in either regard, and it also needs to consider all possible USB errors, known[1] or not.

[1] Resource busy from concurrent runs, timeout from uninitialized devices (Smart Device, Grid+ V3 and probably second generation Krakens like the X61), etc.

jonasmalacofilho:
> That happened during my experiments with increased timeout.

Yes, it is to be expected when testing intentional concurrent calls.  Though if it starts occurring frequently for normal users, it could indicate a more severe issue.

> I am not trying to save the world. [...] feel free to close this. I would not mind.

Thanks for clarifying the intention behind the patch, I get it now.

Since you said it only happened with a specially large timeout and during testing, and since accounting for other scenarios would require a more complex solution (i.e. differentiating between other possible USB erros, offering the traceback another way, etc.), I will close this for now and will eventually try to approach the problem another way.

But still... Thanks!

Issue #17, opened 2019-02-11
Windows Tutorial


cdgriffith:
I think it might be nice if the github had a more in depth guide for getting this software working on Windows from the ground up (as most CAM users may not have a lot of Python experience.)

[I wrote a guide here](https://codecalamity.com/replacing-nzxts-cam-software-on-windows-for-kraken/), that you make take any or all of to include for documentation.

jonasmalacofilho:
I saw that on r/NZXT a bit earlier, great stuff!

I would pick some parts of it to include here and there, but if you're up for it, I would very much welcome a pull request for a proper page under the `/docs` folder.

Another thing: I want to set up a CI system to build packages for Windows automatically.  Is [build.bat](https://github.com/cdgriffith/liquidctl/blob/0b6845e540c41e518658421d6d09bbc4442f0fef/build.bat#L1) all I need?  Do you know if I can easily cross-package it from Linux/Travis CI, or I'm probably better off with Windows/Appveyor?

---

P.S. On the post you mention how you had to look up the available modes in the code.  I guess the docs aren't as visible as I had hoped for, but you can find (links to) them in the [_Supported devices_](https://github.com/jonasmalacofilho/liquidctl#supported-devices) section.

cdgriffith:
Will do! 

PyInstaller was honestly quite the pain on Windows with Python 3.7. Good to document how I got it working though: 

1. Download latest development version (as standard is broken for 3.7) `pip install https://github.com/pyinstaller/pyinstaller/tarball/develop`
2. Modify the `venv\Lib\site-packages\PyInstaller\bindepend.py` file at line 881  (based on https://github.com/pyinstaller/pyinstaller/pull/3956 )  to include: 
```
    if is_win and 'VERSION.dll' in dlls:
        pydll = 'python%d%d.dll' % sys.version_info[:2]
        return getfullnameof(pydll)
```
3. Download the Windows 10 SDK https://developer.microsoft.com/en-us/windows/downloads/windows-10-sdk  (based off their notes here: https://pyinstaller.readthedocs.io/en/stable/usage.html#windows) 
4. Download libusb and add the 32 bit dll in same folder as `build.bat`
5. Update `pip` and `setuptools` via `pip install pip setuptools --upgrade -
-ignore-installed`
5. Then should be able to run `build.bat` 

(For 64-bit Python 3.7 need to update the build.bat to point to x64 folder and grab libusb 64 I believe, not tested.)

[...]

PR #18, opened 2019-02-21
Automated Windows and Linux builds

cdgriffith:
I'm too lazy to write out the tutorial for #17 so instead I just automated the PyInstaller builds with [appveyor](https://ci.appveyor.com/project/cdgriffith/liquidctl).  Will require you to sign up there, but it is free for open source / public repos. 

I am able to test the windows build locally and it is working. The linux one compiles but I don't have method to test. 

Note that it is currently pulling in libusb for Windows, if you chose to keep this method for Windows it will need documentation / licensees / references to comply with LGPL.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ install:
 
 build_script:
   - cmd: C:\Python36-x64\Scripts\pyinstaller --add-data MS64\dll\*;. --clean -F liquidctl\cli.py --name liquidctl
-  - sh: pyinstaller liquidctl\cli.py  --clean -F --name liquidctl
+  - sh: sudo pyinstaller liquidctl\cli.py  --clean -F --name liquidctl
 
 test: off
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,10 @@
-version: 1.1.0-{build}
+version: {build}
 
 skip_commits:
   message: /\[skip ci\]/
 
 
 image:
-  - Ubuntu1804
   - Visual Studio 2017
 
 platform: x64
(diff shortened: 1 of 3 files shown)


jonasmalacofilho:
Thanks!

> I'm too lazy to write out the tutorial for #17 so instead [...]

Will close that issue then.

> The linux one compiles but I don't have method to test.

I don't think a standalone egg for Linux is necessary: installing a release pip or even the sources is much very straightforward on Linux, even considering runtime dependencies.  Packages for Debian/Ubuntu and Fedora would be nice, I might look into that.

jonasmalacofilho:
> Note that it is currently pulling in libusb for Windows, if you chose to keep this method for Windows it will need documentation / licensees / references to comply with LGPL.

Can you look into this?

cdgriffith:
The egg is only showing up because I didn't explicitly define what artifacts to save, so it grabbed everything in the `dist` folder (and I did a `setup.py install` to grab requirements) The other file should be standalone and work on Ubuntu / Fedora. Will fix the egg issue. 

Yeah can look into grabbing the required files and putting them into the docs folder, thankfully you're already GPL so no catches, just need to include their license IIRC.

jonasmalacofilho:
> The other file should be standalone and work on Ubuntu / Fedora. Will fix the egg issue.

My bad, I missed that one.  But for now I think it's ok if you only focus on Windows.

I'll leave Linux packages for later.  We typically want to supply a few distro-specific (yet light) packages, and later the different package managers will handle installation and dependency management appropriately.

---

> Yeah can look into grabbing the required files and putting them into the docs folder

Thanks (again).  But instead of dumping into `/docs`, I think the installer should included a modified `LICENSE.txt` with the various copyright notices and license copy (with headings to mark each library/project).

The alternate file could live in `extra/installer-licenses.txt` and just be used to overwrite the stock `LICENSE.txt` contents.

---

One last thing I noticed: could you change the versioning scheme to just `version: {build}`?  Don't want to have it unnecessarily out of sync.

cdgriffith:
Ok, suggestions implemented and done playing, as I don't think the app has a way to view the license currently I am including with the build artifacts. https://ci.appveyor.com/project/cdgriffith/liquidctl/builds/22534703/artifacts

As a heads up I included a way to have it not auto-build if you specify `[skip ci]` in a commit message, so when just updating text files or something you don't have to trigger it.

jonasmalacofilho:
Thank you for the adjustments.

And I'm sorry if it took me a while to merge this, I've been working on three other things for this project.

PR #20, opened 2019-02-27
Add a "fans-all" channel for the NZXT Smart Device

suve:
This changeset adds `fans-all` channel to the NZXT Smart Device, allowing to set the speed of all fans at once to the same value.

I'm a total newbie to Python, so I suppose this could be done better code-wise. Architecture-wise, should the code perhaps check if the fan is present before sending a command?

### DIFF
--- a/liquidctl/driver/nzxt_smart_device.py
+++ b/liquidctl/driver/nzxt_smart_device.py
@@ -148,7 +148,6 @@ class NzxtSmartDeviceDriver(BaseUsbDriver):
         super().__init__(device, description)
         self._speed_channels = {'fan{}'.format(i + 1): (i, _MIN_DUTY, _MAX_DUTY)
                                 for i in range(speed_channel_count)}
-        self._speed_channels['fans-all'] = (-1, _MIN_DUTY, _MAX_DUTY)
         self._color_channels = {'sync': (0)} if color_channel_count else {}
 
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
Thanks.

I did rename the virtual channel to `sync`, for uniformity [with other code](https://github.com/jonasmalacofilho/liquidctl/blob/master/liquidctl/driver/kraken_two.py#L56).

> should the code perhaps check if the fan is present before sending a command

That isn't required by the firmware, and there is at least one valid use case for changing the duty value of a _seemenly_ unconnected channel: controlling a "two pin" fan that doesn't have SENSE (e.g. small VRM or chipset fans).

---

By the way, @suve, I'm working on Linux Kernel hwmon drivers (i.e. support in lm-sensors) for the devices that work under liquidctl.

```
$ sensors
kraken-hid-3-4c
Adapter: HID adapter
fan1:         829 RPM
Pump:        1992 RPM
Coolant:      +27.8°C  

smart_device-hid-3-4d
Adapter: HID adapter
fan1:        1005 RPM
fan2:           0 RPM
fan3:        1553 RPM
```

Do you want me to let you know once the Smart Device hwmon driver is stable enough for you to try?

suve:
Yeah, having those in `lm-sensors` could be useful.

PR #29, opened 2019-05-22
Change default time-per-color on legacy drivers

AlecVercruysse:
the default time was confusing as it looked like fading. Fixes jonasmalacofilho/liquidctl#27

### DIFF
--- a/liquidctl/driver/asetek.py
+++ b/liquidctl/driver/asetek.py
@@ -374,7 +374,7 @@ class LegacyAsetekDriver(CommonAsetekDriver):
             ('Firmware version', firmware, '')
         ]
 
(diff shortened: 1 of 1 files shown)


AlecVercruysse:
I took a look at at all the other defaults for the legacy driver, and they all seemed fine.

jonasmalacofilho:
Thanks for the investigation and fix.

I increased the fading interval a bit more, to make it stand out more with the 1 Hz sampling, but reverted the blinking interval to 1 second; [if I understood the issue correctly] larger default values for that mode seem unnecessary and awkward.

AlecVercruysse:
Those values look good, cheers

PR #86, opened 2020-02-16
Install pyusb from source due to severely outdated PyPi package

EdvinAkerstrom:
Fix access violation errors on windows machines by installing pyusb from gitrepo due to outdated pypi package

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,7 +96,7 @@ setuptools.setup(
         'Documentation': DOC_URL,
         'Changelog': CHANGES_URL,
     },
-    install_requires=['docopt', 'pyusb', 'hidapi'],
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
_I'm guessing that the issue in question is pyusb/pyusb#203; please correct me if I'm wrong._

Thank you, but I'm leaning towards not merging this because of a few reasons:

 - having a dependency on unreleased library versions is frowned upon (or even a no-go) in many package managers
 - the issue is fairly specific (Windows + libusb >= 1.22), and most of the possibly affected users actually rely on the precompiled exe, which is already patched
 - this would need to be reverted as soon as there's a new release of PyUSB
 - I'm the maintainer of PyUSB these days, and a new release is indeed coming (soon™)

Does this make sense?

Issue #118, opened 2020-05-30
Kraken X53, X63 and X73 with Grid+ V3 Fan profile


mburgholte:
First of all, thank you very much for your work. It works very well on my system.
Right now I have a question. Is it possible to enter a profile for my fans on a Grid+ controller depending on the fluid temperature? 

For example:
`liquidctl --product 0x1711 set fan1 speed 20 30 30 50 34 80 40 90 50 100`

Another not so important thing, I know this project is called liquidctl. What if RGB could also be controlled by e.g. RAM memory? It is so nice to see how people are working on this project. Maybe it would be possible?

jonasmalacofilho:
Hi @mburgholte,

At the moment it isn't possible to control Grid+ fans based on the coolant temperature of a Kraken.  More generally, liquidctl currently only exposes the features that are available at the firmware level and that thus can work without software interaction (that is, after being configured... and until the device resets).

What I call software-based cooling profiles is something I started to explore in #72, but haven't yet been able to stabilize into a good API/CLI.  Ideas and (even better) pull requests welcome!

---

Regarding support for other RGB componentes, like RAM, it's not something I'm opposed to, but it also isn't something I'm personally looking into at the moment.

There are other projects more focused on the RGB side of things (like [OpenRGB](https://gitlab.com/CalcProgrammer1/OpenRGB/activity)), and I view cross-platform support (which is a strong requirement in this project) for things like memory or (non-USB) motherboard controllers as something tricky to get right.  For example, OpenRGB requires a custom Super I/O kernel driver on some motherboards.

Still, pull requests welcome.  However, I would also need some help with maintaining the code after the initial development.

mburgholte:
Hi @jonasmalacofilho,
thanks a lot for your explanation. Regarding the Grid+ fans, I am thinking of some kind of helper script. This script uses liquidctl to read the current temperature and then changes the grid+ fan setpoints. It might not be nice, but this is how you could control the fans.

I'll have a look at OpenRGB.

jonasmalacofilho:
That's a perfectly acceptable solution.  In fact, you can start with [`extra/yoda`](https://github.com/jonasmalacofilho/liquidctl/blob/master/extra/yoda) (or the older `extra/krakentwo-poc`).  And please let me know if you come up with a good solution.

[...]

Issue #72, opened 2019-12-19
Software-based fan/pump speed profiles


sebikolo:
How to control pump and fans based on CPU temperature, not Liquid temperature?

jonasmalacofilho:
liquidctl only exposes native device features (through a CLI and a set of APIs).

The supported coolers only have a liquid temperature sensor, and there's no way for them to _directly_ access or support CPU temperature sensors.  
But you can use the liquidctl APIs + a suitable sensor API to create a script that adjusts the speeds according to the algorithm of your choice.

There's already an example of this in [`extra/krakencurve-poc`](https://github.com/jonasmalacofilho/liquidctl/blob/master/extra/krakencurve-poc), which supports fan/pump curves based on any sensor available through the Linux hwmon subsystem.

That example can be adapted to support complex algorithms (e.g. including  hysteresis), access to sensors on platforms other than Linux, or the use of a different cooler.

One final note: as this will be software-based fan/pump speed control, the script will need to be kept running as if it was a service.  This differs from the liquidctl CLI, that manipulates hardware-based features and doesn't have to run continuously.

Please let me know if you have any other questions.

sebikolo:
Thank you for the quick reply. Unfortunately, I'm not a programmer :-(. I'm looking for a ready solution.

jonasmalacofilho:
What OS are you running?

sebikolo:
macOS Catalina 10.15.2

jonasmalacofilho:
I don't have a mac to work on this, but there should be _some_ way of getting the CPU temperature programmatically.  Maybe through the BSD layer, maybe using a library or CLI program.

If you can figure that out and point me to how it works, I should be able to help.  The example script I linked already is functional, it just lacks access to external sensors on macOS.

---

@CaseySJ do you think you can help us here?

CaseySJ:
@jonasmalacofilho 

One of the command line tools we can use is a Ruby app called `iStats`. We can install it via the command `sudo gem install iStats`. Here is some sample CLI output:

```
$ istats --help
 
  - iStats: help ---------------------------------------------------

  istats --help                        This help text
  istats --version                     Print current version

  # Commands
  istats all                           Print all stats
  istats cpu                           Print all CPU stats
  istats cpu [temp | temperature]      Print CPU temperature
  istats fan                           Print all fan stats
  istats fan [speed]                   Print fan speed
  istats battery                       Print all battery stats
  istats battery [health]              Print battery health
[...]

PR #132, opened 2020-06-11
liquidctlfan wrapper script for your liquidctl

mburgholte:
As dicussed in #118 
I tried it with yoda, but it seems to be a little bit slower than sensors.

### DIFF
--- a/extra/contrib/liquidctlfan/liquidctlfan
+++ b/extra/contrib/liquidctlfan/liquidctlfan
@@ -21,6 +21,8 @@ FANT4="100"
 SLTIME="10"
 #Enable Syslog
 SYSLOG="1"
+#Unit Celsius C or Fahrenheit F
+UNIT="C"
 programs=(bc sensors liquidctl logger)
 
 for program in "${programs[@]}"; do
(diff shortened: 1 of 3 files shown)


jonasmalacofilho:
> I tried it with yoda, but it seems to be a little bit slower than sensors.

What you mean by slower?  Slower to respond?  If that's the case, that's intentional, as it uses an exponential moving average as a low-pass filter to avoid random spikes in fan speed and noise.

_(note: the configuration of the filter is incorrectly implemented in yoda, but it shouldn't make a huge practically difference; still, I can fix it if that's the only problem)_

---

Can I also ask you to add a (simple) README to the folder?  Your script is a very nice clean example and I would like more people to immediately get that.

mburgholte:
You're right. Thanks. I fixed it. I'm not sure, where to put a copyright notice..

extra/contrib/liquidctlfan/liquidctlfan
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+# Place this script in /usr/bin or you've to adjust systemd unit file. 
+# Please have a look at cputemp. You may have to adjust something for your system.
```
jonasmalacofilho, line 3: My suggestion would be something simple like:

```bash
#!/bin/bash
#
# liquidctlfan - wrapper script for liquidctl to control your fans
#
# Place this script in /usr/bin or you've to adjust systemd unit file. 
# Please have a look at cputemp. You may have to adjust something for your system.
#
# Copyright (C) 2020  Martin Burgholte
# SPDX-License-Identifier: GPL-3.0-or-later

Name="Liquidctl Fan Control"
...
```

jonasmalacofilho:
> I am thinking about a separate config file?! What do you think?

If you can find a nice balance between generality and simplicity I will be very interested in seeing how you achieved that.

But I don't think it's a must here...

---

By the way, please don't forget to let me know what exactly was slow with yoda, as it's something I'll eventually get back to (more specifically, the liquidctl-integrated version of it for #72).

mburgholte:
> > I am thinking about a separate config file?! What do you think?
> 
> If you can find a nice balance between generality and simplicity I will be very interested in seeing how you achieved that.
> 
> But I don't think it's a must here...
Sometimes it is a nice idea to have a sleep. ;o) 
Now we have all possibilities. And I think it's so easy to use.

> By the way, please don't forget to let me know what exactly was slow with yoda, as it's something I'll eventually get back to (more specifically, the liquidctl-integrated version of it for #72).

I thought I already reported the results. 
I did the following:

- time sensors
sensors  0,00s user 0,00s system 96% cpu 0,005 total

- time yoda
yoda --product 0x1711 show-sensors  0,08s user 0,02s system 2% cpu 3,841 total

jonasmalacofilho:
Oh, I see, and that's to be expected, because yoda (as liquidctl), has to load the Python VM...

But the idea isn't to use yoda as a replacement to sensors, but rather as a replacement to a loop and successful invocations to liquidctl.

Both should be (mostly) equivalent:

```
$ liquidctlfan -p 0x1711 -u c -ct1 50.0 -ct2 60.0 -ct3 70.0 -ct4 80.0 -f0 30 -f1 40 -f2 50 -f3 80 -f4 100 -i 10
$ yoda -m grid control sync with "(49,30),(50,40),(60,50),(70,80),(100,100)" on <name from --show-sensors>
```

Yoda _should_ be lighter on the CPU because it doesn't have to initialize the Python VM, initialize the liquidctl CLI, find the correct device and connect to/disconnect from it... every iteration.  

_(That is, unless psutils, which is what yoda is using to get the hwmon data in Python, is surprising inefficient.)_

But the two commands above are not truly equivalent, because I've not passed `--interval 10`, and will default to 2 s.  In yoda I've started to implement a fancy smoothing algorithm, but it's still not smart enough to deal with very high intervals and will likely introduce too much delay to the speed adjustments.  What I mean by this delay is the time it takes from the CPU temperature to _really_ increase (that is, not considering instantaneous spikes and noise) to the the fans being properly sped up.

---

That said, your script is a work a scripting art!

Issue #142, opened 2020-06-22
CoolIT coolers like Corsair H80i and H110i GT


RobertoBiundo:
The current system has compatibility for H110i GTX using the Asetek driver but the GT version is developed by Coolit.

I don't think it is too hard to adapt the current Coolit platinum driver in order to make room for the 110i GT. I can make the necessary code changes or even create a new driver if needed and submit a Pull Request but I need help with the reverse engineering part of the protocol.

I tried to read the communication using Wireshark on windows but with no success. Can someone post a quick guide on how where the other protocols figure out so that I can follow the same workflow for the 110i GT and add support for it?

H110i GT:
Product Id: 0C04

RobertoBiundo:
I tried already adding to the coolit platinum driver the 110i GT but with no success. It reports firmware as 0.0.0.0 or 15.0.0.0 ( if I set fan count to 0, the 110i GT has 2 fans). My thoughts are that the driver is close to being compatible but maybe the 110i GT manages something in the protocol differently.

jonasmalacofilho:
Hi @RobertoBiundo, great to have you here!

I've been meaning to write a blog post about some of the techniques I use, but I always get stuck on the least interesting (even if probably as important) parts about dealing with the tools.  Since that "proper" post isn't happening any time soon, I'll just go over the main ideas here, and we can explore more details as needed.

But before anything else: as far as I know the H100i GT does indeed use a different protocol.

---

## USB transfers

At a basic level you can view USB traffic as a collection of transfers.  Transfers can be of a few different types, but for the most part we're only interested in control transfers, interrupt transfers and, very occasionally, bulk transfers.  I'll skip over the purposes of each type since we'll merely use whichever ones the protocol mandates.

But our devices at not necessarily always manipulated as USB devices.  In fact, it's rather common that we need to work on a layer further up the abstraction chain, on a Human Interface Device (HID).  USB HIDs are a special type of USB device specified by a corresponding "interface class".¹

Working with HID protocols is almost identical to working with other classes of USB devices: they use control transfers and interrupt transfers, and we capture them with Wireshark.  The distinction does matter in a few places, most notability around the "report ID", but I'll get to that later...

## Wireshark, part one

So, you have captured some USB traffic to your device with Wireshark.  How do you approach that data?

[...]

PR #147, opened 2020-06-28
Added experimental support for H110i GT

RobertoBiundo:
This is a partial solution to #142, adding support for the Corsair H110i GT.
At the moment the driver only supports 
- Fan Control over PWM
  liquidctl set fan speed 30
- Fan and pump RMP read
  liquidctl status
- Temperature read from pump
  liquidctl status
- Firmware version reading and initialization
  liquidctl initialize
- Fan curve set
  liquidctl set fan speed  20 30  30 50  34 80  40 90  50 100

Pending:
- Fan control over RPM ( cannot figure out how to even set it using the liquidctl command )
- LED color read and set ( it will remaining to whatever you left it set up using iCUE or CorsairLink4 )
- Pump curve!. Yes, apparently this is possible ( not even possible with Corsair tools )

### DIFF
--- a/liquidctl/driver/coolit.py
+++ b/liquidctl/driver/coolit.py
@@ -1,0 +1,313 @@
+"""liquidctl driver for Corsair Platinum and PRO XT coolers.
+
+Supported devices
+-----------------
+
+ - Corsair H110i GT
+
+Supported features
+------------------
+
+ - general monitoring
+ - pump speed control
+ - fan speed control
+
+Copyright (C) 2020–2020  Jonas Malaco
+Copyright (C) 2020–2020  each contribution's author
+
+SPDX-License-Identifier: GPL-3.0-or-later
+"""
+
+import itertools
+import logging
+
+from enum import Enum, unique
+
+from liquidctl.driver.usb import UsbHidDriver
+from liquidctl.keyval import RuntimeStorage
+from liquidctl.pmbus import compute_pec
+from liquidctl.util import clamp, fraction_of_byte, u16le_from, normalize_profile
+
+
+LOGGER = logging.getLogger(__name__)
+
+_REPORT_LENGTH = 64
+_MAX_FAN_RPM = 2250
+_PUMP_INDEX = 0x02
+
+_COMMAND_FIRMWARE_ID = 0x01
+_COMMAND_TEMP_READ = 0x0E
+_COMMAND_FAN_SELECT = 0x10
+_COMMAND_FAN_MODE = 0x12
+_COMMAND_FAN_FIXED_PWM = 0x13
+_COMMAND_FAN_FIXED_RPM = 0x14
+_COMMAND_FAN_READ_RPM = 0x16
+_COMMAND_FAN_RPM_TABLE = 0x19
+_COMMAND_FAN_TEMP_TABLE = 0x1A
+
+_OP_CODE_WRITE_ONE_BYTE = 0x06
+_OP_CODE_READ_ONE_BYTE = 0x07
+_OP_CODE_WRITE_TWO_BYTES = 0x08
+_OP_CODE_READ_TWO_BYTES = 0x09
+_OP_CODE_WRITE_THREE_BYTES = 0x0A
+_OP_CODE_READ_THREE_BYTES = 0x0B
+
+_PROFILE_LENGTH = 5
+_CRITICAL_TEMPERATURE = 60
+
+_PUMP_DEFAULT_QUIET = [0x2e, 0x09]
+_PUMP_DEFAULT_EXTREME = [0x86, 0x0b]
+
+@unique
+class _FanMode(Enum):
+    FIXED_DUTY = 0x02
+    FIXED_RPM = 0x04
+    CUSTOM_PROFILE = 0x0E
+
+    @classmethod
+    def _missing_(cls, value):
+        LOGGER.debug("falling back to FIXED_DUTY for _FanMode(%s)", value)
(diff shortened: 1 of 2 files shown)


liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,313 @@
...
+        
+        LOGGER.debug('write %s', buf.hex())
+        self.device.clear_enqueued_reports()
+        self.device.write(buf)
+        buf = bytes(self.device.read(_REPORT_LENGTH))
+        self.device.release()
+        LOGGER.debug('received %s', buf.hex())
+        return buf
+
+    def _send_set_cooling(self):
```
jonasmalacofilho, line 250: This can be simplified, since the protocol doesn't require all cooling settings to be sent together.

In this case sending the commands immediately, instead of saving them to temporary storage (`self._data`) and sending them in groups (what this function does), is cheaper and bit simpler.

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,313 @@
...
+    QUIET = 0x08
+    EXTREME = 0x0C
+
+    @classmethod
+    def _missing_(cls, value):
+        LOGGER.debug("falling back to QUIET for _PumpMode(%s)", value)
+        return _PumpMode.QUIET
+
+
+def _sequence(storage):
```
jonasmalacofilho, line 84: You mentioned that perfectly increasing and cycling sequence numbers might not be required, based on how the protocol looks like.

If this is right and the device responds correctly, It would be cheaper to use a simple cycling iterator initialized with a random value, instead of read/writing to `self._data`.

Or maybe even sending a constant (only randomized at `connect` time) is fine if the device doesn't complain and we end-up not checking the ACKs.

tvidal-net:
Hi guys, is there any progress on this? Is there anything I can do to  help?

jonasmalacofilho:
@RobertoBiundo

If you think discussing the remaining problems may help, let me know. I prefer async text (=issues or email), but if there's a particular topic that may benefit from more synchronous communication, we can try to schedule something.

jonasmalacofilho:
I'm going to close this due to the lack of feedback.

That said, the code appears to be functional, just a bit more complex than necessary.  So it probably just needs a bit of refactoring (besides some tests and documentation), and it's at the very least a good reference.

P.S. I normally expect the contributor to stick around and help maintain the driver afterwards its merged.  Some issues require real world testing, and I don't have this device.  Also, sometimes the code and/or the protocol are too complex for me to reason about what the driver _should_ do (as opposed to simply what it does).

P.P.S. Secondary copy of the patch [here](https://github.com/jonasmalacofilho/liquidctl/commit/60472d84c3b438dc09461a6461e82b46a38c3b6a.patch).

jonasmalacofilho:
Oh, forgot to say...

@RobertoBiundo, let me know if you find a bit more time to work on this (or simply if you intend to work on this in the near future), and how you think I can help. 

Thanks again!

RobertoBiundo:
Im going to work on this in the next few weeks yes.

Unfortunately I did some SSD upgrades on the computer I was developing and
testing on and will need time to reinstall Linux and get it all working
again.

jonasmalacofilho:
@RobertoBiundo 

Ok, awesome.  When the time comes, let me know if you need any help.

liquidctl/driver/coolit.py
sanmai: From the looks of it I doubt this was supposed to be a global sequence, but rather a per-command sequence. I hacked it a bit to always start with `1`, and it still works.

ItsMeAubey:
What is the status of this?

I am desperately looking for a replacement for icue, which hogs 5-10% of my cpu at all times. 

I've tried going through all of the linked PRs/issues but cannot figure out whether or not the h110i gtx is actually supported. liquidctl does not seem to see mine, even after i zadig'd the "integrated usb bridge" that shows up over usb. 

Thank you!

jonasmalacofilho:
> I've tried going through all of the linked PRs/issues but cannot figure out whether or not the h110i gtx is actually supported. liquidctl does not seem to see mine, even after i zadig'd the "integrated usb bridge" that shows up over usb.

Please open a [bug report](https://github.com/liquidctl/liquidctl/issues/new?assignees=&labels=bug&template=bug_report.md) so we can investigate why liquidctl isn't detecting your H110i GTX.

CRCinAU:
Just wondering if anyone got the chance to look at this?

I don't code in Python, so its a bit beyond me - but otherwise, we're still stuck with OpenCorsairLink without this being merged...

Yes, I'm still running a H110i (non GT) - and its still running strong. Quality bit of kit that cools a Ryzen 5950x without issue. Moving to another 280mm AIO would be wasteful for no benefit.

CRCinAU:
Sadly, nearly a year on, this cooler is still going strong but nobody picked this up :(

Serphentas:
@CRCinAU I believe Jonas was waiting new feedback from OP (@RobertoBiundo) but he hasn't made an appearance yet. Are you able to test preliminary code versions so that this PR can be foolproofed ?

CRCinAU:
Sure - if something can be provided, I can start running it to see if it works....

Serphentas:
Great ! Since OP's branch is not available for write, I suggest you fork his repo and use the `master` to continue work from there. We can then talk about the implementation over IM. Do you have Discord ?

CRCinAU:
I do, but I'm also not very competent with python  - its probably my least used language.

PR #179, opened 2020-08-28
Fix user HOME path for macOS

mediocrescientist:
Currently the cache folder for macOS is `~/Library/Caches` but the `~` is not expanded to the current user directory. 

If I open a terminal in my HOME folder and run `liquidctl status` it generates `~/Library/Caches/liquidctl/...` inside my HOME directory, i.e. `/Users/<username>/~/Library/Caches`. If I now change the directory in the terminal to `~/Downloads` and run liquidctl again, it creates a directory `/Users/<username>/Downloads/~/Library/Caches/liquidctl/...`.

To fix this behaviour, `os.path.expanduser(path)` is used to expand the `~` symbol to it's absolute path. Now the cache directory will be in the correct place.

### DIFF
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -23,7 +23,7 @@ def get_runtime_dirs(appname='liquidctl'):
     if sys.platform == 'win32':
         dirs = [os.path.join(os.getenv('TEMP'), appname)]
     elif sys.platform == 'darwin':
-        dirs = [os.path.join('~/Library/Caches', appname)]
+        dirs = [os.path.expanduser(os.path.join('~/Library/Caches', appname))]
     elif sys.platform == 'linux':
         # threat all other platforms as *nix and conform to XDG basedir spec
         dirs = []
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -56,12 +56,10 @@ class _FilesystemBackend:
         self._read_dirs = [os.path.join(x, *key_prefixes) for x in get_runtime_dirs()]
         self._write_dir = self._read_dirs[0]
         os.makedirs(self._write_dir, exist_ok=True)
-        if XDG_RUNTIME_DIR and os.path.commonpath([XDG_RUNTIME_DIR, self._write_dir]):
+        if sys.platform == 'linux':
             # set the sticky bit to prevent removal during cleanup
             os.chmod(self._write_dir, 0o1700)
-            LOGGER.debug('data in %s (within XDG_RUNTIME_DIR)', self._write_dir)
-        else:
-            LOGGER.debug('data in %s', self._write_dir)
+        LOGGER.debug('data in %s', self._write_dir)
 
     def load(self, key):
         for base in self._read_dirs:


mediocrescientist:
Tests are failing because of `os.path.commonpath` in [here](https://github.com/jonasmalacofilho/liquidctl/blob/0ccb9d9887df7597087dd60114a8d4c9894cd133/liquidctl/keyval.py#L59) but I don't really understand the logic behind this. In the tests `XDG_RUNTIME_DIR` is a relative path.

`XDG_RUNTIME_DIR` _must_ always be an absolute path according to [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). `os.path.commonpath` needs all paths to be either absolute or relative, therefore `self._write_dir` should be absolute as well. `os.path.commonpath` always returns something when comparing absolute paths, e.g. the leading `/` in *nix or drive letter like `c:\`. So what is the point to check whether the paths have something in common? 

If I understand it correctly, the idea is to check whether the `self._write_dir` is inside the `XDG_RUNTIME_DIR`. In that case, would it make sense to check whether the common part of two paths is equal to `XDG_RUNTIME_DIR`?
@jonasmalacofilho

jonasmalacofilho:
Wow, thanks for noticing this issue.

---

Does macOS adhere to the XDG base dir spec?  I didn't know/expect that.

The idea of that test is to only set sticky bit in the computed directory is within a `$XDG_RUNTIME_DIR`.  But the test is wrong, the sticky bit must be set even if the environment variable hasn't been: a platform that adheres to the XDG base dir spec may always clean up the contents of the runtime directory, and we want to prevent our files from disappearing (hence the sticky bit).  That said, we still don't want to (attempt to) set the sticky bit on platforms where we aren't using a XDG runtime directory.

On another front, if macOS adheres to the XDG base dir spec, then we should definitely use that instead of custom location shenanigans.  Alternatively, to avoid SSD trashing from drivers that actively use the runtime storage (albeit minimal), we should really prefer directories that are (normally) memory-backed: if there's to XDG_RUNTIME_DIR, then `/tmp/...` or some other memory backend mount is a better location.

(the criticisms are obviously towards myself; thanks for fixing this)

mediocrescientist:
No, as fas as I know macOS does _not_ adhere to the XDG base dir spec and I guess there is no real need to check for this env. variable on macOS. It seems `~/Library/Caches` is the most commonly used place for cache'ing files. I did not find any memory backed directories.

> the sticky bit must be set even if the environment variable hasn't been: a platform that adheres to the XDG spec may clean up the contents of the runtime directory, and we want to prevent our files from disappearing.

Would it then make sense to set the sticky bit always when `sys.platform == 'linux'` and never in other cases? As macOS writes to a cache dir that is never automatically deleted and sticky bit doesn't work in Windows. Changed this now in the PR, give a look. I hope I understood you correctly.

jonasmalacofilho:
> Would it then make sense to set the sticky bit always when sys.platform == 'linux' and never in other cases?

Yeah, that's probably better.  Thanks.

> It seems ~/Library/Caches is the most commonly used place for cache'ing files. I did not find any memory backed directories.

That's where I got from reading Apple's docs, but maybe there's a better way: is `/tmp` user-writable and typically memory-backed?

mediocrescientist:
> That's where I got from reading Apple's docs, but maybe there's a better way: is `/tmp` user-writable and typically memory-backed?
`/tmp` is user writable but is not memory backed. Contents are deleted after every boot and after three days of no activity for the files. Nevertheless the folder seems to be not heavily used as it is practically empty for me. So I would propose to leave it as is because:

* I didn't find any folder that is memory backed by default
* Most of the programs keep their cache in `~/Library/Caches`
* There are some other options like `$TMPDIR`, `/tmp` or even `/var`  but I don't know why any of these would be a better option. Out of these the `$TMPDIR` would be the most logical alternative.

jonasmalacofilho:
That makes sense.  Merging...

Thanks again!

PR #181, opened 2020-08-31
Use liquiddump as a basis to create a Prometheus metrics exporter

alexberryman:
## Changes
I took the idea of `liquiddump` and instead of dumping the stats out as JSON, they are instead exposed as an HTTP endpoint in a format for the Prometheus metric service to collect. Once Prometheus collects this data it is then easy to visualize in tools like Grafana. Each time the metrics web page is requested the `extra.liquidexport.LiquidCollector.collect()` method is called to collect the data from `liquidctl` and create the output.

## New packages required
`pip3 install prometheus-client`

## Example Output
`liquidexport` will host an HTTP endpoint so that Prometheus can scrape the metrics. I picked port `8098` to be the default, but exposed a `--port` option. Opening http://localhost:8098/metrics in your browser will show the basic python Prometheus metrics and then below that you will find the `liquidctl` specific metrics.

`liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="liquid_temperature",unit="°C"} 35.1`
Looking at the specific example line above you can see that it has the following properties:
- The metric name is `liquidctl` to match the name of the main library you built
- The metric then has several labels to help filter and aggregate the data once it is stored in Prometheus
  - `device` is collected from `d.description`
  - `sensor` is a normalized version of the name of each stat that is reported from `d.get_status()` 
  - `unit` is the direct result from `d.get_status()` 
- Finally, the `35.1` is the actual measurement that was created when the page was loaded

```
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 167.0
python_gc_objects_collected_total{generation="1"} 176.0
python_gc_objects_collected_total{generation="2"} 0.0
# HELP python_gc_objects_uncollectable_total Uncollectable object found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 44.0
python_gc_collections_total{generation="1"} 4.0
python_gc_collections_total{generation="2"} 0.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="3",minor="8",patchlevel="5",version="3.8.5"} 1.0
# HELP liquidctl liquidctl exported metrics
# TYPE liquidctl gauge
liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="liquid_temperature",unit="°C"} 35.1
liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="fan_speed",unit="rpm"} 1154.0
liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="pump_speed",unit="rpm"} 2064.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_1_duty",unit="%"} 100.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_1_speed",unit="rpm"} 1290.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_2_duty",unit="%"} 100.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_2_speed",unit="rpm"} 978.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_3_duty",unit="%"} 100.0
liquidctl{device="NZXT Smart Device V2",sensor="fan_3_speed",unit="rpm"} 1321.0
liquidctl{device="NZXT Smart Device V2",sensor="noise_level",unit="dB"} 58.0
```
## Example Dashboard
By graphing this data out in Grafana it makes it easy to visualize the impact and effectiveness of fan curves. In the top left panel of this [liquidexport Prometheus Example dashboard](https://snapshot.raintank.io/dashboard/snapshot/TYi10VjL63wwwOEjLqihGqGXAICsSiDc?orgId=2) I created green/yellow/red horizontal thresholds to represent the following fan profiles defined using `liquidctl`:
- ` ./liquidctl.exe set --match kraken pump speed 40 60 41 80 44 100`
- `./liquidctl.exe set --match kraken fan speed 40 60 41 80 44 100`

Now every time the blue `Liquid Temp` line crosses a threshold at 40°C/41°C/44°C you can see the impact on the `AIO Pump` and `Exhaust - Radiator` fan RPMs.
liquidexport_example_dashboard

## Question for Reviewers
1. Is this something that you want to add to your main repository or should I spin it off to a separate package that simply requires `liquidctl`?
2. Each loop through `extra.liquidexport.LiquidCollector.collect()` I am calling `d.connect()`, gathering the data I need, and then calling `d.disconnect()`. Is this pattern ideal or should I refactor my code to maintain a single long-running connection to the devices? If I keep the connection open for hours/days are there any drawbacks?

### DIFF
--- a/extra/liquidexport
+++ b/extra/liquidexport
@@ -16,19 +16,8 @@ Usage:
   liquidexport --version
 
 Options:
-  --interval <seconds>    Update interval in seconds [default: 2]
-  --hid <module>          Override API for USB HIDs: usb, hid or hidraw
   --legacy-690lc          Use Asetek 690LC in legacy mode (old Krakens)
-  --match <name>          Filter devices by name
-  --vendor <id>           Filter devices by vendor id
-  --product <id>          Filter devices by product id
-  --release <number>      Filter devices by release number
-  --serial <number>       Filter devices by serial number
-  --bus <bus>             Filter devices by bus
-  --address <address>     Filter devices by address in bus
-  --usb-port <port>       Filter devices by USB port in bus
-  --pick <number>         Pick among many results for a given filter
-  --port <number>         Port for the HTTP /metrics endpoint
+  --server-port <number>  Port for the HTTP /metrics endpoint
   -v, --verbose           Output additional information
   -g, --debug             Show debug information on stderr
   --version               Display the version number
@@ -44,19 +33,15 @@ SPDX-License-Identifier: GPL-3.0-or-later
 """
 
 import logging
+import re
 import sys
 import time
-
-import liquidctl.cli as _borrow
 import usb
 
 from docopt import docopt
 from liquidctl.driver import *
-
 from prometheus_client import start_http_server
 from prometheus_client.core import GaugeMetricFamily, REGISTRY
-
-import re
 
 LOGGER = logging.getLogger(__name__)
 
@@ -66,9 +51,8 @@ def gauge_name_sanitize(name):
 
 
 class LiquidCollector(object):
-    def __init__(self, args):
+    def __init__(self):
         self.description = 'liquidctl exported metrics'
-        self.args = args
 
     def collect(self):
         labels = ['device', 'sensor', 'unit']
@@ -95,9 +79,7 @@ class LiquidCollector(object):
 
 if __name__ == '__main__':
     args = docopt(__doc__, version='0.1.0')
-    frwd = _borrow._make_opts(args)
-    devs = list(find_liquidctl_devices(**frwd))
-    update_interval = float(args['--interval'])
+    devs = list(find_liquidctl_devices())
 
     if args['--debug']:
         args['--verbose'] = True
@@ -108,10 +90,10 @@ if __name__ == '__main__':
         logging.basicConfig(level=logging.WARNING, format='%(message)s')
         sys.tracebacklimit = 0
 
-    REGISTRY.register(LiquidCollector(args))
+    REGISTRY.register(LiquidCollector())
 
-    if args['--port']:
-        port = int(args['--port'])
+    if args['--server-port']:
+        port = int(args['--server-port'])
     else:
         port = 8098
 
@@ -120,6 +102,7 @@ if __name__ == '__main__':
 
     try:
         while True:
-            time.sleep(update_interval)
+            # Keep HTTP server alive in a loop
+            time.sleep(2)
     except KeyboardInterrupt:
         LOGGER.info('Canceled by user')
--- a/extra/liquidexport
+++ b/extra/liquidexport
@@ -77,9 +77,23 @@ class LiquidCollector(object):
         yield g
 
 
+def _make_opts(args):
+    opts = {}
+    for arg, val in args.items():
+        if val is not None and arg in _PARSE_ARG:
+            opt = arg.replace('--', '').replace('-', '_')
+            opts[opt] = _PARSE_ARG[arg](val)
+    return opts
+
+
+_PARSE_ARG = {
+    '--legacy-690lc': bool
+}
+
 if __name__ == '__main__':
     args = docopt(__doc__, version='0.1.0')
-    devs = list(find_liquidctl_devices())
+    opts = _make_opts(args)
+    devs = list(find_liquidctl_devices(**opts))
 
     if args['--debug']:
         args['--verbose'] = True
--- a/extra/liquidexport
+++ b/extra/liquidexport
@@ -59,7 +59,6 @@ class LiquidCollector(object):
         g = GaugeMetricFamily('liquidctl', self.description, labels=labels)
         i = InfoMetricFamily('liquidctl', self.description, labels=['device'])
         for d in devs:
-            d.connect()
             try:
                 get_status = d.get_status()
                 for metric in get_status:
@@ -86,7 +85,6 @@ class LiquidCollector(object):
             except usb.core.USBError as err:
                 LOGGER.warning('Failed to read from the device, possibly serving stale data')
                 LOGGER.debug(err, exc_info=True)
-            d.disconnect()
         yield g
         yield i
 
@@ -108,6 +106,9 @@ if __name__ == '__main__':
     args = docopt(__doc__, version='0.1.0')
     opts = _make_opts(args)
     devs = list(find_liquidctl_devices(**opts))
+    for d in devs:
+        LOGGER.info('Initializing %s', d.description)
+        d.connect()
 
     if args['--debug']:
         args['--verbose'] = True
@@ -134,3 +135,10 @@ if __name__ == '__main__':
             time.sleep(2)
     except KeyboardInterrupt:
         LOGGER.info('Canceled by user')
+    finally:
+        for d in devs:
+            try:
+                LOGGER.info('Disconnecting from %s', d.description)
+                d.disconnect()
+            except:
+                LOGGER.exception('Unexpected error when disconnecting')
--- a/extra/liquidexport
+++ b/extra/liquidexport
@@ -12,8 +12,6 @@ liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="liquid_temperatu
 
 Usage:
   liquidexport [options]
-  liquidexport --help
-  liquidexport --version
 
 Options:
   --legacy-690lc          Use Asetek 690LC in legacy mode (old Krakens)
--- a/extra/liquidexport
+++ b/extra/liquidexport
@@ -33,7 +33,6 @@ SPDX-License-Identifier: GPL-3.0-or-later
 """
 
 import logging
-import re
 import sys
 import time
 import usb
@@ -41,7 +40,8 @@ import usb
 from docopt import docopt
 from liquidctl.driver import *
 from prometheus_client import start_http_server
-from prometheus_client.core import GaugeMetricFamily, REGISTRY
+from prometheus_client.core import GaugeMetricFamily, REGISTRY, InfoMetricFamily
+from datetime import timedelta
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
Hi @alexberryman,

That's very interesting.

> 1. Is this something that you want to add to your main repository or should I spin it off to a separate package that simply requires `liquidctl`?

I would very much like to have it here.  That said, it might wise to keep it separate, particularly if you intend to distribute it to Linux distros (which I think that you should).

The main problem is that there are no plans for a liquidctl v2, and having your script packaged with it would restrict you from making breaking changes (for the foreseeable future).

> 2. Each loop through `extra.liquidexport.LiquidCollector.collect()` I am calling `d.connect()`, gathering the data I need, and then calling `d.disconnect()`. Is this pattern ideal or should I refactor my code to maintain a single long-running connection to the devices? If I keep the connection open for hours/days are there any drawbacks?

Performance wise it's better to keep long running connections.  Drivers should already release the device (without disconnecting) before returning from `get_status` (when necessary).

But I think you also need to handle the case where the connection is lost for some reason, and in that scenario it is likely that the device handle is no longer valid.  So to robustly handle this I think you will need to probe the buses again.

Actually, the most complete output would require probing devices at every iteration, to account for newly connected devices.  But that's very costly, and I think that probing only when a connection is lost is more than enough.

Of course, you need to disconnect/connect all handles before/after probing.

extra/liquidexport
```diff
@@ -0,0 +1,127 @@
...
+liquidctl{device="NZXT Kraken X (X42, X52, X62 or X72)",sensor="liquid_temperature",unit="°C"} 33.6
+
+Usage:
+  liquidexport [options]
+  liquidexport --help
+  liquidexport --version
+
+Options:
+  --interval <seconds>    Update interval in seconds [default: 2]
+  --hid <module>          Override API for USB HIDs: usb, hid or hidraw
```
jonasmalacofilho, line 20: It is no longer necessary to forward `--hid` since it is ignored since 1.4.0.

extra/liquidexport
```diff
@@ -0,0 +1,127 @@
...
+  --match <name>          Filter devices by name
+  --vendor <id>           Filter devices by vendor id
+  --product <id>          Filter devices by product id
+  --release <number>      Filter devices by release number
+  --serial <number>       Filter devices by serial number
+  --bus <bus>             Filter devices by bus
+  --address <address>     Filter devices by address in bus
+  --usb-port <port>       Filter devices by USB port in bus
+  --pick <number>         Pick among many results for a given filter
+  --port <number>         Port for the HTTP /metrics endpoint
```
jonasmalacofilho, line 31: Having both `--usb-port` and `--port` may be confusing.

It may be reasonable to remove device filtering altogether (`--legacy-690lc` is still necessary because it affects driver behavior).

Or maybe rename the specific option to `--server-port`?

extra/liquidexport
```diff
@@ -0,0 +1,127 @@
...
+from prometheus_client import start_http_server
+from prometheus_client.core import GaugeMetricFamily, REGISTRY
+
+import re
+
+LOGGER = logging.getLogger(__name__)
+
+
+def gauge_name_sanitize(name):
+    name = re.sub(r'(?<!^)(?=[A-Z])', '_', name).lower()
```
jonasmalacofilho, line 65: This regex substitution will result in weird output for something like:

```python
>>> some_status_item = ('LED accessories', 2, '')  # from Smart Device V1
>>> re.sub(r'(?<!^)(?=[A-Z])', '_', some_status_item[0]).lower()
'l_e_d accessories'
```

Admittedly one could argue that this specific metric should be filtered out, but "LED" and other acronyms are bound to appear in more interesting places.

extra/liquidexport
```diff
@@ -0,0 +1,127 @@
...
+    def collect(self):
+        labels = ['device', 'sensor', 'unit']
+        g = GaugeMetricFamily('liquidctl', self.description, labels=labels)
+        for d in devs:
+            d.connect()
+            try:
+                get_status = d.get_status()
+                for metric in get_status:
+                    # Only allow numerical stats and exclude string values like "firmware='6.3.3'"
+                    if isinstance(metric[1], (int, float, complex)) and not isinstance(metric[1], bool):
```
jonasmalacofilho, line 84: This filter doesn't match `datetime.timedelta`s; we have a couple of those in one driver for PSU uptime (session and ever).

On the other hand, have you considered filtering out all metrics without a unit?

---

If you haven't already seen it, the CLI also implements some [conditional formatting of status values](https://github.com/jonasmalacofilho/liquidctl/blob/de95601ba4cc2d4b59e600d5a691669e6b56e16f/liquidctl/cli.py#L190-L195).

extra/liquidexport
```diff
@@ -0,0 +1,127 @@
...
+                        sanitized_name = gauge_name_sanitize(metric[0])
+                        label_values = [d.description, sanitized_name, metric[2]]
+                        g.add_metric(label_values, value=metric[1])
+                        LOGGER.debug(
+                            '{}: {} as {} with labels {}'.format(d.description, metric, sanitized_name,
```
jonasmalacofilho, line 89: Because the server is not expected to run in debug mode, it makes sense to only use the old-fashioned `%` based format that `logging` natively supports, and that only lazily builds the strings.

extra/liquidexport
alexberryman: Very good point. 

Re-evaluating the name used again I think I can delete the regex and just replace it with two simple steps:
1. Replace spaces with `_`
2. String to lower

```python
def gauge_name_sanitize(name):
    return name.replace(" ", "_").lower()
```

extra/liquidexport
alexberryman: Removing the `--hid` option lead to an error since `liquidctl/cli.py` still cares about `--hid`. 
```python
  File "liquidexport", line 97, in <module>
    frwd = _borrow._make_opts(args)
  File "/mnt/c/Users/alexb/code/play/liquidctl/src/liquidwslenv/lib/python3.8/site-packages/liquidctl/cli.py", line 220, in _make_opts
    if args['--hid']:
KeyError: '--hid'
```

I piggybacked on the same `args` forwarding boilerplate from `liquiddump` which passes them onto `liquidctl.cli`:
```python
import liquidctl.cli as _borrow
if __name__ == '__main__':
    args = docopt(__doc__, version='0.1.0')
    frwd = _borrow._make_opts(args)
    devs = list(find_liquidctl_devices(**frwd))
```

If I remove all device filtering then I can avoid this issue which seems fine for this tool.

extra/liquidexport
alexberryman: I removed the device filtering, but copied over some arg validation to keep `--legacy-690lc` working and flowing into `find_liquidctl_devices(**opts)` which is where I assume it is used by the drivers.

I also renamed the other key to `--server-port` to differentiate it from the USB ports in other scripts.

extra/liquidexport
alexberryman: I added support for the `datetime.timedelta` values. They will be cast into `seconds` and have their `unit` updated to reflect that before the data is used to create a GaugeMetric. I tested this by faking the following timedelta which yields the following metric:
```python
get_status.append(('Current uptime', timedelta(seconds=42), ''))
```
`liquidctl{device="NZXT Smart Device V2",sensor="current_uptime",unit="seconds"} 42.0`

I also added support for `firmware` and other unit-less (non-timedelta) statues as an InfoMetric:
`liquidctl_info{device="NZXT Kraken X (X42, X52, X62 or X72)",firmware_version="6.0.2"} 1.0`

alexberryman:
@jonasmalacofilho I addressed most of your feedback. Thanks for taking the time to look through this!

To quickly summarize the new commits:
- got rid of the regex and used simpler string manipulation for metric names
- removed device filtering but kept the `--legacy-690lc` argument.
- renamed `--port` to `--server-port` even though `--usb-port` was removed
- added support for `datetime.timedelta` by casting it to `seconds` and defining a unit
- swapped to long-running connections, the performance difference is minimal

The remaining unaddressed issue is:
> handle the case where the connection is lost for some reason, and in that scenario it is likely that the device handle is no longer valid. So to robustly handle this I think you will need to probe the buses again.

Would this condition be caught by the `d.get_status()` throwing an `usb.core.USBError` exception? If so my gut reaction is to change the catch-block to something like:
```python
except usb.core.USBError as err:
    LOGGER.warning('Failed to read from the device, possibly serving stale data')
    LOGGER.debug(err, exc_info=True)
    devs = list(find_liquidctl_devices(**opts))
   continue
```

This would rebuild the list of connected devices if one became un-reachable. If this makes sense then I would also go back to calling `d.connect()` and `d.disconnect()` each loop since the performance difference was minimal.

jonasmalacofilho:
> Would this condition be caught by the d.get_status() throwing an usb.core.USBError exception? If so my gut reaction is to change the catch-block to something like [...]

That makes sense, but most devices are using Hidapi and will raise another exception.  Unfortunately I did not pay that much attention to error handling, so some testing will be necessary.

> If this makes sense then I would also go back to calling d.connect() and d.disconnect() each loop since the performance difference was minimal.

The performance difference is indeed minimal.  Additionally, in most cases `connect` / `disconnect` don't involve any transfers to the device; of the top of my head, the exceptions are some Asetek coolers that are _not_ HIDs: old Krakens, some old Corsair coolers too, and the EVGA CLCs.

Either way, while the cost is never high, there may be some blocking where a transfer is required, plus the extra `connect` / `disconnect` calls may be enough to put the CPU into a higher power state, resulting in higher idle temperatures (I have observed this in the past, on my machine).

Still, keeping it simple for now, if marginally to slightly less efficient, would be a perfectly valid decision IMO.

extra/liquidexport
```diff
@@ -0,0 +1,144 @@
...
+                    sample_value = metric[1]
+                    unit = metric[2]
+
+                    if isinstance(sample_value, timedelta):
+                        # cast timedelta into seconds and override the supplied unit
+                        sample_value = sample_value.seconds
+                        unit = 'seconds'
+
+                    if unit != '':
+                        label_values = [d.description, sanitized_name, unit]
```
jonasmalacofilho, line 75: I forgot to mention before, but because these are being logged it's probably desirable to keep the label as stable as possible, so you may want to sanitize and remove any ` (experimental)` suffixes from `d.description`.

---

Additionally, a system can have more than one device of the same model.  This is a particular concern for the devices that, at the USB level, are indistinguishable within a whole family (e.g. Kraken X coolers within a generation).

While each device would ideally have a unique serial number, some don't have serials numbers at all, and others save constant/shared information (like hardware version) in that descriptor. : \

To make matters worse, using bus addresses (or even USB port numbers, in the few cases where they are available) wouldn't result in stable identifiers.  But maybe that's better than nothing?

extra/liquidexport
```diff
@@ -0,0 +1,144 @@
...
+                        g.add_metric(label_values, value=sample_value)
+                        LOGGER.debug(
+                            '%s: %s as GaugeMetric %s labels %s' %
+                            (d.description, metric, sanitized_name, '/'.join(label_values)))
```
jonasmalacofilho, line 79: You can also let the `logging` module apply the `%` operator for you, which is what results in the string being constructed lazily.

```suggestion
                        LOGGER.debug(
                            '%s: %s as GaugeMetric %s labels %s',
                            d.description, metric, sanitized_name, '/'.join(label_values))
```

jonasmalacofilho:
I started to look into packaging this for ArchLinux.  It seems that in Arch and Debian exporters follow the convention `prometheus-<name>-exporter` for both binary and package names.

extra/liquidexport
```diff
@@ -0,0 +1,144 @@
...
+  -v, --verbose           Output additional information
+  -g, --debug             Show debug information on stderr
+  --version               Display the version number
+  --help                  Show this message
+
+Examples:
+  liquidexport
+  liquidexport --product 0xb200
+  liquidexport --interval 0.5
+
```
jonasmalacofilho, line 30: Both options have been removed, and the remaining example is IMO not worth keeping.

Issue #149, opened 2020-07-06
H100i RGB PRO XT Lighting not responding to commands


ShaunTheQuietGamer:
Making new issue as requested, as stated previously, My H100i RGB PRO XT is initialized, the fan control works, but lighting does not respond to commands. logs from one for the commands I tried below. (it does not respond to any of the command variations in its help docs.)
```
shaun  ~  sudo liquidctl set led color off --debug
[DEBUG] liquidctl.cli: running liquidctl v1.3.3 (Arch Linux; liquidctl-git 1.3.3.r238.454cde3-1)
[DEBUG] liquidctl.driver.usb: searching HidapiBus (api=hid, drivers=[CommonSmartDeviceDriver, CoolitPlatinumDriver, CorsairHidPsuDriver, KrakenTwoDriver, KrakenX3Driver, KrakenZ3Driver, SeasonicEDriver, SmartDeviceDriver, SmartDeviceV2Driver])
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:172b
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:172b
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:172b
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:172b
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:150d
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:150d
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1726
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1726
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1726
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1726
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c20
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair H100i PRO XT (experimental)
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1610
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1610
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1610
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1610
[DEBUG] liquidctl.driver.usb: probing drivers for device 1038:1610
[DEBUG] liquidctl.driver.usb: probing drivers for device 0b05:18a3
[...]

PR #187, opened 2020-10-08
Faulty "too many colors message"

MarshallAsch:
When running the command  `liquidctl set led color super-fixed FF0000 00FF00 0000FF` with a Corsair H100i platinum cooler it gave the error message `too many colors, dropping to 1` I believed this to be in error as the documentation and the code for that command allows up to a max of 24 colors to be given.

I have resolved this misleading error message by adding a `maxcolors` check and giving that message when the number given exceed that amount. I am not sure but this may also be an issue for other coolers but I only checked this one. 

Tested using  version 1.4.1 built from source on Fedora31.

### DIFF
--- a/liquidctl/driver/hydro_platinum.py
+++ b/liquidctl/driver/hydro_platinum.py
@@ -130,6 +130,11 @@ class HydroPlatinum(UsbHidDriver):
             ('led', 'fixed'): 1,
             ('led', 'off'): 0,
         }
+        self._maxcolors = {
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
Thanks!

P.S. The use of of separate `_mincolors` and `_maxcolors` smells a bit, but that's the fault of (my)the code you were working with, which was already messy (besides not working properly).  Hopefully this will get refactored for #149.

Issue #186, opened 2020-09-23
Product ID parsing glitch


andrejpodzimek:
I'm using version 1.4.1 [from AUR](https://aur.archlinux.org/packages/liquidctl/) and it doesn't like product IDs in the usual hexadecimal format (used by `lsusb`) when they start with `0`. This might have something to do with integer parsers that expect such values to be octal, so hexadecimal digits can be considered garbage or, in the worst case, misinterpreted.

The device:
```
Bus 003 Device 002: ID 1b1c:0c17 Corsair H115i Platinum
```

The command line:
```bash
liquidctl --vendor=1b1c --product=0c17 set fan2 speed 20 50 40 80 45 100
```

The error message:
```
ValueError: invalid literal for int() with base 0: '0c17'
```

A workaround is to use `0x` to let the parser know it's hexadecimal.
```bash
liquidctl --vendor=0x1b1c --product=0x0c17 set fan2 speed 20 50 40 80 45 100
```

jonasmalacofilho:
Agreed, that would be better.  Also more consistent, since colors don't even support the `0x` prefix.

Actually, both IDs and colors should _optionally_ accept the prefix (either way being parsed as in base 16).

---

This is a good first issue, please consider contributing the improvement.

_(The corresponding parsing functions need doctests showing that both cases [with and without the prefix] are handled appropriately)._

MarshallAsch:
Im going to take on this issue. and I wanted to clarify something. 

the values for the `product` and `colors` should accept hex values optionally with a `0x` prefix. This also applies to the `vendor` flag. Should the `vendor` and `product` be allowed to be given as decimal numbers or some other base number?

jonasmalacofilho:
Great, thanks!

I think we should disallow decimal values for the `--vendor` and `--product` IDs.  But there's not much we can do in the code, this is more of a documentation thing.

MarshallAsch:
Thats what I figured, especially since those values would be given elsewhere to the user as hex values, and what sane person would convert it to decimal to give it back to the program.

jonasmalacofilho:
Improved/fixed with #189.

PR #189, opened 2020-10-09
Product ID parsing fix

MarshallAsch:
Fix for issue #186. 

- Modified the  the `color_from_str()` function to accept hex color values with the prefixes 0x` or `0X` or without the prefix. 
- Also added 3 more doctests to ensure  verify that it behaves correctly.
- Modified the list of supported color options to show this change
- corrected the parsing of the `--vendor`, `--product`, and `--release` flags to only accept hex values either with or without the `0x` or `0X` prefixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ Lighting is controlled in a similar fashion.  The specific documentation for eac
 
 When configuring lighting effects, colors can be specified in different representations and formats:
 
- - as an implicit hexadecimal RGB triple: e.g. `ff7f3f`
+ - as an implicit hexadecimal RGB triple, either with or without the `0x` prefix: e.g. `ff7f3f`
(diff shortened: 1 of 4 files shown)


MarshallAsch:
Did this need any modifications or is there anything I missed when making these changes?

jonasmalacofilho:
I honestly still need to check it in more detail, sorry about that. : )

That said, the man page (`liquidctl.8`) needs to be updated.  Also, there are probably a few other places where the `0x` prefix is suggested for IDs, but it would be better to normalize everything to _not_ use the prefix.

MarshallAsch:
I forgot about the liquidctl.8 file, and I'll take a look around at the other doc files to see where else it is mentioned.

jonasmalacofilho:
Thank you (again)!

Issue #190, opened 2020-10-09
question: Corsair H100i Platinum `super-fixed` option 


MarshallAsch:
When setting the RGB colours for the H100i Platinum cooler my understanding from the docs is that the command  `# liquidctl set led color super-fixed <up to 24 colors>` should set individual LED's to static colors and any unset values will be left off. 

When I tested it with `liquidctl set led color super-fixed FF0000 00FF00 0000FF` the LEDs change colour in a marquee like pattern, and it does not go around every single LED. It Simultaneously is set for the pump, fan1, and fan2. Im not sure if this is the desired behaviour or if it is a bug.

I have attached a video recording (in a zip folder) showing this behaviour. 

[IMG_0345.MOV.zip](https://github.com/jonasmalacofilho/liquidctl/files/5357108/IMG_0345.MOV.zip)

- Firmware version: 1.1.15
- Fedora31
- liquidctl 1.4.1

jonasmalacofilho:
That's certainly not the documented behavior.

And besides the possible/probable implementation error, it's also not how I understood the device/protocol to work.

(I was working with iCue and an emulated cooler – an Arduino with custom firmware – so I could only see what iCue showed me, not what a real device would do).

Hopefully you can help here too? : )

P.S. You may need to do some new traffic captures,<sup>[[1]](https://github.com/jonasmalacofilho/liquidctl/tree/master/docs/developer)</sup> or at least take a look at the ones we already have.<sup>[[2]](https://github.com/jonasmalacofilho/liquidctl-device-data)</sup>

MarshallAsch:
Yup I'd be happy to see if I can figure it out, (and also trying to get my hacktoberfest PRs in here).
 
I have never done any protocol reverse engineering before, so that might be a bit tough to figure out. Im trying to get a windows VM setup on this computer so I can put icue on that. (while Im figuring that out Ill try to fill out some information for [this](https://github.com/jonasmalacofilho/liquidctl/blob/master/docs/developer/capturing-usb-traffic.md#capturing-usb-traffic-on-linux) section of the documentation)

Do you have any tips on getting things setup to identify commands / test to see if sending a certain command actually does what I think it does?

jonasmalacofilho:
> (while Im figuring that out Ill try to fill out some information for this section of the documentation)

[...]

PR #194, opened 2020-10-13
Added a fix to the LED super-fixed bug, but it uses magic strings

MarshallAsch:
This is a fix for #190. It adds 3 magic packets to the initialize function that will enable the LED setting mode on the cooler. 

I still have no idea if sending that command has any other effects, from the traces added in jonasmalacofilho/liquidctl-device-data/pull/1 it seems that icue sends the same message (or atleast nearly the same) message multiple times so it _should_ be okay. It is currently working on my desktop.

Tested with Fedora 31.

### DIFF
--- a/liquidctl/driver/hydro_platinum.py
+++ b/liquidctl/driver/hydro_platinum.py
@@ -164,6 +164,17 @@ class HydroPlatinum(UsbHidDriver):
         Returns a list of `(property, value, unit)` tuples.
         """
 
+        # These hex strings are currently magic values that work but Im not quite sure why.
+        d1 = bytes.fromhex("0101ffffffffffffffffffffffffff7f7f7f7fff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffffffffffffffffffffffffff")
+        d2 = bytes.fromhex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f2021222324252627ffffffffffffffffffffffffffffffffffffffffff")
+        d3 = bytes.fromhex("28292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4fffffffffffffffffffffffffffffffffffffffffff")
+        
+        # Send the magic messages to enable setting the LEDs to statuC values
+        self._send_command(None, 0b001, data=d1)
+        self._send_command(None, 0b010, data=d2)
+        self._send_command(None, 0b011, data=d3)
+
+
         self._data.store('pump_mode', _PumpMode[pump_mode.upper()].value)
         res = self._send_set_cooling()
         fw_version = (res[2] >> 4, res[2] & 0xf, res[3])
@@ -174,6 +185,15 @@ class HydroPlatinum(UsbHidDriver):
 
         Returns a list of `(property, value, unit)` tuples.
         """
+
+
+        d1 = bytes.fromhex("0101ffffffffffffffffffffffffff7f7f7f7fff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffffffffffffffffffffffffff")
+        d2 = bytes.fromhex("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f2021222324252627ffffffffffffffffffffffffffffffffffffffffff")
+        d3 = bytes.fromhex("28292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4fffffffffffffffffffffffffffffffffffffffffff")
+
+        self._send_command(None, 0b001, data=d1)
+        self._send_command(None, 0b010, data=d2)
+        self._send_command(None, 0b011, data=d3)
 
         res = self._send_command(_FEATURE_COOLING, _CMD_GET_STATUS)
         assert len(self._fan_names) == 2, f'cannot yet parse with {len(self._fan_names)} fans'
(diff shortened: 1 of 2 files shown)


MarshallAsch:
I noticed that the test failed, Im not quite sure why, could it have something to do with the way it is being mocked and the initialize function now sending several messages?

liquidctl/driver/hydro_platinum.py
```diff
@@ -164,6 +164,17 @@ def initialize(self, pump_mode='balanced', **kwargs):
         Returns a list of `(property, value, unit)` tuples.
         """
 
+        # These hex strings are currently magic values that work but Im not quite sure why.
```
jonasmalacofilho, line 167: If these aren't needed unless to set LED settings, then they should only be sent in that API.

---

Additionally, I try to avoid sending opaque blobs: even when they fix a current issue, they are unmaintainable.

I'm also worried these in particular could be storing non-volatile state, which together with the opaqueness, could lead us to brick devices: if these do anything to non-volatile state we need to know (at least to some degree) their contract.  For example: if `d2` fails, do we just abort `d3` and (later) restart at `d1` or do we need some unknown `d4` to clear it up?

Again, if these weren't so opaque, it would be easier to reason about.  So, practically, can you try to distill these down to the underlying byte and/or bit flags that are being sent?

Like, are these always constant? Do any of these change when you toggle/vary different settings on iCue?  When are these sent exactly, and is it always the three, in this sequence, and with nothing else in between?  What does the cooler respond to each one, and are the responses the same too? Etc.

liquidctl/driver/hydro_platinum.py
```diff
@@ -175,6 +186,15 @@ def get_status(self, **kwargs):
         Returns a list of `(property, value, unit)` tuples.
         """
 
+
+        d1 = bytes.fromhex("0101ffffffffffffffffffffffffff7f7f7f7fff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffff00ffffffffffffffffffffffffffffff")
```
jonasmalacofilho, line 190: Is this necessary?  By the way, `get_status` needs to be as lean as possible, since it may be used for device monitoring and maybe be feeding into programs that iteratively call it every 1 or 2 seconds.

liquidctl/driver/hydro_platinum.py
```diff
@@ -268,7 +288,7 @@ def set_color(self, channel, mode, colors, unsafe=None, **kwargs):
         data1 = bytes(itertools.chain(*((b, g, r) for r, g, b in expanded[0:20])))
         data2 = bytes(itertools.chain(*((b, g, r) for r, g, b in expanded[20:])))
         self._send_command(_FEATURE_LIGHTING, _CMD_SET_LIGHTING1, data=data1)
-        self._send_command(_FEATURE_LIGHTING, _CMD_SET_LIGHTING2, data=data2)
+        #self._send_command(_FEATURE_LIGHTING, _CMD_SET_LIGHTING2, data=data2)
```
jonasmalacofilho, line 291: This change may have been accidental, but just in case: this second command is needed because there isn't enough room in the first command to set all (possible) 24 LEDs.

jonasmalacofilho (commented):
Regarding the tests, you changed how the protocol works so you would need to adjust the tests accordingly.

The idea is that tests check for the minimum requirements we know are necessary for the driver to work, while the implementation is free to do that in whatever way it's more efficient or easier (for example, it could reuse buffers only caring about the fields that are know to be read for a particular command).  The implementation can also include experimental additional features.

liquidctl/driver/hydro_platinum.py
MarshallAsch: oops this is in the wrong spot, this should not be here. I didn't notice that the file stayed when I switched branches. Ill fix that.

liquidctl/driver/hydro_platinum.py
MarshallAsch: yup this was accidental when I was testing something else. 

I also discovered that there is a possible 3rd message for setting LEDs, in case a different fan is being used that has 16 LEDs instead of 4, (mentioned in the  jonasmalacofilho/liquidctl-device-data repository).

jonasmalacofilho: I missed that part, let me check it again.

liquidctl/driver/hydro_platinum.py
MarshallAsch: > Like, are these always constant?

From the different wireshark captures that I have they are the almost the exact same every time they are sent. The only bytes that I have noticed are different is the first 2 `0101` is sometimes `0000`, but I have not been able to tell why. I will look into it a little more tho. 


> If these aren't needed unless to set LED settings, then they should only be sent in that API.

Thats a good point, I can move this to the `set_color()` and set a persistent value to not send it again if it has already been set.

jonasmalacofilho:
Not knowing what goes on in some many bytes makes me weary, but you have the device and this solves a real problem.  So I'm just going to merge it and, hopefully, we'll improve it later.

---

By the way, payloads `d2` and `d3` are just enumerating all possible byte values up to 0x4f.  Weird.

MarshallAsch:
> By the way, payloads d2 and d3 are just enumerating all possible byte values up to 0x4f. Weird.

I noticed that too, I also noticed from some of the other captures that sometimes it enumerates them slightly differently, (sometimes there were a couple less byte values in that enumeration). perhaps an issue could be opened for those 3 lines incase someone else wants to take a look later to see if they have any ideas about getting more info from it.

jonasmalacofilho:
Knowing that the enumeration sometimes starts from a different value makes me think it could be related to the sequence number.  But I guess we'll only find out more if it eventually breaks.

That said, have you checked whether d2 and d3 are actually necessary?
(sorry if I already asked you this)

MarshallAsch:
I haven't actually checked that yet.

Although I have noticed if I run the command wrong `liquidctl set led ffffff .... ` then it also seems to change the mode, however I didnt think that actually sent a message to the device
.

jonasmalacofilho:
Oh, we're sending these magic packets before validating the channel, mode and number of colors.

Can you fix this in a new PR?

MarshallAsch:
Oops my bad, yup Ill fix that in a new PR.

PR #197, opened 2020-10-14
Status firmware

MarshallAsch:
Added the device firmware version to the `get_status()` response for the Corsair hydro series coolers. I added this because a number of the other coolers include the firmware version in the status message. The tests were also updated accordingly

### DIFF
--- a/liquidctl/driver/hydro_platinum.py
+++ b/liquidctl/driver/hydro_platinum.py
@@ -176,12 +176,15 @@ class HydroPlatinum(UsbHidDriver):
         """
 
         res = self._send_command(_FEATURE_COOLING, _CMD_GET_STATUS)
+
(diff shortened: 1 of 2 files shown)


jonasmalacofilho:
Since this is a newer driver, we opted to only report the (constant during a particular session) firmware version in `initialize`.

The ability to report info from `initialize` is relatively new, and older drivers don't make use of it to ensure backwards compatibility: some scripts and third-party projects (like GKraken) expect to find the firmware version in the `get_status` response.

Despite this, newer drivers like this one usually place session-constant data in `initialize`, keeping `get_status` as lean as possible, both in output size (relevant to reduce noise in the CLI output) and CPU time.

---

I'm not going to take this PR, but it's a valid proposal and I'm going to hacktoberfest-accept it just in case you're doing that.

MarshallAsch:
That makes sense, I didn't realize that.

Issue #204, opened 2020-10-22
Thoughts on a bash completion script


MarshallAsch:
I have recently been looking at bash tab completion and I was wondering if this would be something your interested in having added to this project before I work on that much more. 

Personally I really like tab completion because I don't have to remember the entire command I want to type and it will give hints. 

Im not sure how great my implementation will be, but I think it will be better than no completions.

jonasmalacofilho:
Absolutely!

---

On a related note, in my `.zshrc` I'm using `compdef _gnu_generic liquidctl` to have some completion be automatically generated from the output of `--help`.  But it's limited to options, so neither commands nor device specific details are suggested.

```
$ liquidctl --<TAB>
--address          -- Filter devices by address in bus
--alert-color      -- Color used by the visual high temperature alert
--alert-threshold  -- Threshold temperature for a visual alert (°C)
--bus              -- Filter devices by bus
--debug            -- Show debug information on stderr
--device           -- Select device by listing id
--help             -- Show this message
--hid              -- Deprecated
--legacy-690lc     -- Use Asetek 690LC in legacy mode (old Krakens)
--match            -- Filter devices by description substring
--pick             -- Pick among many results for a given filter
--product          -- Filter devices by product id
--pump-mode        -- Set the pump mode (certain Corsair coolers)
--release          -- Filter devices by release number
--serial           -- Filter devices by serial number
--single-12v-ocp   -- Enable single rail +12V OCP
--speed            -- Abstract animation speed (device/mode specific)
--time-off         -- Time to wait with the LED turned off (seconds)
--time-per-color   -- Time to wait on each color (seconds)
--unsafe           -- Comma-separated bleeding-edge features to enable
--usb-port         -- Filter devices by USB port in bus
--vendor           -- Filter devices by vendor id
--verbose          -- Output additional information
--version          -- Display the version number
```

MarshallAsch:
Thats cool I didn't know it did it automatically, I know that zsh has a far more powerful autocomplete functionality, but I use bash and haven't used it very much. 

In #205, I did not add any device specific completions, but I did go through all of the files in `/doc` to pull out the different channels for each of the devices.  Although I did add the commands in there are a couple things that I did not implement yet.

PR #205, opened 2020-10-22
Bash completion

MarshallAsch:
Closes: #204 
Added a basic bash completion script. 

Tasks that this does / should have:

- [X] all of the flags
- [X] the arguments for most of the flags that take arguments for specifying a device
- [X] each of the commands and the sub commands that are currently implemented
- [X] all of the channels under the set command (note that these are not device specific suggestions rather just all the currently accepted options)
- [x] Remove flags that have already been specified from the suggestions
- [x] allow flags to be given after a command, might not actually want this tab completion behaviour as it will cause a lot of extra options to be listed. It might be better to assume that the flags will be given _before_ the command
- ~make the suggestions device specific, this would require added some extra commands to liquidctl to provide a list~ Do this separately


liquidctl-completion.sh
```diff
@@ -0,0 +1,273 @@
...
+  elif [[ "$is_fan" -eq "0" ]]
+  then
+      _liquidctl_set_led
+  else
+
+    # no trailing space here so that the fan number can be appended
+    compopt -o nospace
+    # possibly use some command here to get a list of all the possible channels from liquidctl
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    COMPREPLY=($(compgen -W "fan fan1 fan2 fan3 fan4 fan5 fan6 led led1 led2 led3 led4 led5 led6 pump sync ring logo external" -- "$cur"))
```
jonasmalacofilho, line 256: I hadn't thought about simply suggesting all possible channel names, regardless of the particular device.  It's a small enough set, and it's better than nothing.  Cool!

liquidctl-completion.sh
```diff
@@ -0,0 +1,273 @@
+#!/usr/bin/env bash
+
+# This is a bash completion script for liquidctl 
+# to enable this completion place this file in /etc/bash_completion.d/ (note this locatio may be system dependant_ 
+# you can also enable this manually by sourcing this file in your ~/.bashrc file `source path/to/this` 
+#
+# Created by: Marshall Asch
```
jonasmalacofilho, line 7: Unless this is super common convention for completion files, I think it's better to have an explicit copyright and license notice.

```
Copyright (C) 2020–2020  Marshall Asch
```

CHANGELOG.md
```diff
@@ -5,13 +5,15 @@
 Changelog since 1.4.1:
 ### Added
  - Add `Modern690Lc.downgrade_to_legacy` (unstable API)
+ - Added a bash completion script for tab completion
```
jonasmalacofilho, line 8: nitpick: "Add completion script for Bash"

CHANGELOG.md
```diff
@@ -5,13 +5,15 @@
...
 ### Changed
  - Accept hexadecimal inputs regardless of a `0x` prefix
  - Warn on faulty temperature readings from Kraken X3 coolers
  - Warn on Hydro Platinum/PRO XT firmware versions that are may be too old
  - Update PyInstaller used for the Windows executable
  - Update PyUSB bundled with the Windows executable
  - Improve the documentation
+ - Updated installation instructions for the manpage
```
jonasmalacofilho, line 16: Not necessary, the previous item already covers this.

liquidctl-completion.sh
MarshallAsch: I really dont like that I hard coded all those values, I was originally thinking of added a new command to liquidctl (perhaps `--list-channels` and `--list-modes`) to get the values for specific devices, but the way the values are stored for each driver did not seem consistent at first glance and it would require a larger refactor then I really wanted to do for this.

liquidctl-completion.sh
MarshallAsch: nope there isnt a convention, I'm just new to "good" file headers, most of the files I have put it on have been for classes.

CHANGELOG.md
MarshallAsch: didnt notice that, Ill remove this in the next revision

jonasmalacofilho:
>  allow flags to be given after a command, might not actually want this tab completion behaviour as it will cause a lot of extra options to be listed. It might be better to assume that the flags will be given before the command

I'm honestly not sure what is better during tab completion, but from the user's perspective certains flags may only appear sensible (to users) after the command that uses them: e.g. `set led color loading --faster`.

liquidctl-completion.sh
jonasmalacofilho: Yes, having a way to query the driver for what it (and the particular device that it's bounded to) support would be better, but also a lot of work.

liquidctl-completion.sh
jonasmalacofilho: The format you arrived at was fine in isolation, but since we're shipping a bunch of stuff together, I think it's easier to just normalize credits/copyright/license info in a single format.

MarshallAsch:
Different groups of flags could also be suggested after the command. I wasn't sure which ones. From a usability perspective I would think that all of the device selection flags, debugging, verbose, should be only at the beginning, not sure about the others tho.

jonasmalacofilho:
I agree with you in general, and I follow a similar rule in all examples.

But in the end it really depends on what the user is doing: in a system with more than one similar device, 

```
liquidctl set sync color fixed f070d5 --match x62
liquidctl set sync color fixed f070d5 --match x53
```

can make sense (for a usability perspective).

Also, I personally only append `--debug` (also, but less often, `--verbose`) at the end of commands.

MarshallAsch:
I have updated the ability to add the flags at the end of the command, but only if the user enters '-' as the last character. I chose this because it makes it easy to get the flag suggestions at the end but wont start filling them in if you press tab too many times by accident. 

I think that I am not going to add functionality into liquidctl to list what the valid options are for things like channels or mode as that would change the API and I think that would be too large of a change for this PR.

MarshallAsch:
This is now complete. It will remove flags that have already been given from the suggested list. But it will not remove both short and long forms of the flag.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 Changelog since 1.4.1:
 ### Added
  - Add `Modern690Lc.downgrade_to_legacy` (unstable API)
+ - Added completion script for bash
```
jonasmalacofilho, line 8: No longer in the right place since 1.4.2; needs to te in a new "[Unreleased]" block.

_(note: the context showed by GitHub around this change is misleading)_

liquidctl-completion.sh
```diff
@@ -0,0 +1,337 @@
+#!/usr/bin/env bash
```
jonasmalacofilho, line 1: Let's move this to `extra/linux`.

README.md
```diff
@@ -148,6 +148,9 @@ $ cd liquidctl
 $ pytest  # optional step
 $ python -m liquidctl.cli <args>...
 # pip install .
+# sudo cp liquidctl-completion.sh /etc/bash_completion.d/	# to install bash completion script
+# sudo cp liquidctl.8 /usr/local/share/man/man8/
+# sudo mandb
```
jonasmalacofilho, line 153: I'm not sure about suggesting this for all manual installs.  These files wouldn't be tracked, would eventually become trash, and could also conflict/interact badly with a later install of a distro package of liquidctl.

Can you place these into a separate code block, suggesting that they are optional steps?

And since we're already taking care of `liquidctl.8`, the same could be done with the udev rules.

liquidctl-completion.sh
```diff
@@ -0,0 +1,337 @@
...
+# you can also enable this manually by sourcing this file in your ~/.bashrc file `source path/to/this` 
+#
+#
+# Copyright (C) 2020-2020 Marshall Asch 
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+
+
+
```
jonasmalacofilho, line 14: Nit: weird extra 3 blank lines.

liquidctl-completion.sh
```diff
@@ -0,0 +1,337 @@
...
+    liquidctl list -v | grep 'Product ID:' | cut -d ':' -f 2 | cut -c4-6 | sort -u
+}
+
+_list_device_options () {
+    liquidctl list | cut -d ' ' -f 3 | cut -d ':' -f 1 | sort -u
+}
+
+_list_match_options () {
+    liquidctl list | cut -d ':' -f 2 | sort -u
+}
```
jonasmalacofilho, line 37: This worked surprisingly well, but I two additional suggestions:

- make everything lowercase (the comparison is insensitive anyway)
- remove parenthesis

For reference, this is the output on my system:

```
$ liquidctl --match <TAB>
1080            (experimental)  Kraken          Smart           (X42,           X72)
Device          FTW             NZXT            (V1)            X52,            
EVGA            GTX             or              X               X62             
```

jonasmalacofilho (commented):
LGTM in general, by I marked a few (very minor) things.

liquidctl-completion.sh
MarshallAsch: This wouldn't be specific to linux tho, and would be installed as part of the dist package

README.md
MarshallAsch: yup I'll move this to an optional step.

liquidctl-completion.sh
jonasmalacofilho: Regarding distribution, everything in `extra` is part of the sdist.

But you're right, this is not specific to Linux (although also not super general, as many users are not running bash, even on *nix platforms).  So how about `extra/bash/liquidctl-completion.sh`?

_(We'll presumably have other completion scripts in the future, and this one is specific to bash.)_

MarshallAsch: how about `extra/completions/liquidctl-completion.sh`? since there likely would be more just bash extras

MarshallAsch:
I think I fixed all of your comments. I did a quick check after the rebase to see if anything got missed and I didnt notice anything.

liquidctl-completion.sh
jonasmalacofilho: I think `bash` has to be somewhere in the path so it wont be confusing when another completion script is added.

But at this point it doesn't matter _that_ much, it's just something to remember to check/fix before 1.5.0.

jonasmalacofilho: Naming things is hard, but also important.  But this pattern (shell completions) is very common, how are other projects handling the multiple shells?  We should the same, just (somewhere) in `extra/`.

README.md
```diff
@@ -149,6 +149,20 @@ $ pytest  # optional step
...
+- install bash completion
+```
+# cp extra/completions/liquidctl-completion.sh /etc/bash_completion.d/	# to install bash completion script
+```
+- install udev rules (on linux)
+```
+# cp  extra/linux/71-liquidctl.rules /etc/udev/rules.d/
+```
 
 _Note: in systems that default to Python 2, use `pip3`, `python3` and `pytest-3`._
```
jonasmalacofilho, line 167: This note should come before the optional steps.

liquidctl-completion.sh
MarshallAsch: > This pattern (shell completions) is very common, how are other projects handling the multiple shells

git: 
`contrib/completion/git-completion.bash`

Most of the completions that I have found have been located in the one `bash-completions` project and not in the individual project that they belong to. (tbh I also didnt look all that hard for individual projects with them)

Issue #155, opened 2020-07-20
Corsair Commander Pro support


hgrahamcs:
Hello, I was looking to attempt to port the commander pro code from [audiohacked/OpenCorsairLink ](https://github.com/audiohacked/OpenCorsairLink) and would like to know where I should start. So far I have been attempting to understand how the driver works in OpenCorsairLink but haven't had much luck yet. When I do eventually figure it out what steps would I have to take to make a new driver for it in this project. 

Thanks.

jonasmalacofilho:
Hi Henry,

Great to have you here!

I wrote [this longer than usual comment](https://github.com/jonasmalacofilho/liquidctl/issues/129#issuecomment-640258429) about porting drivers from OCL.  Even though it's for a different device, the general steps are the same and it should get you started.  Let me know how it goes and if you have any questions...

Also, if you decide you need to do some additional reverse engineering of the protocol, [this other comment](https://github.com/jonasmalacofilho/liquidctl/issues/142#issuecomment-650568291) may also help.

hgrahamcs:
Thanks for the quick response, 

Those both look like great resources. Thanks for the help.

jonasmalacofilho:
Note: the Corsair Lighting Node Core (issue #154) appears to share the same protocol.

hol430:
Hi all, just wondering if there has been any work done on this issue? I'm interested in helping out with this, but don't want to duplicate any existing efforts. I have my own quick and dirty (certainly not fully-featured) driver which was based on some of the work in open corsair link, but figure it will be helpful to more people if it lives in this library.

hol430:
Ok, so I have made a start on this and have a question about how we want `set_fixed_speed()` to work. The commander pro supports setting speed as fixed RPM or as a percentage. I assume we want to encapsulate both of these inside `set_fixed_speed()`, and if so how would we ideally want to expose this to the CLI? Would we want something like `liquidctl --match "Commander pro" set <channel> speed 1250 --rpm`?

I'm still not super familiar with the rest of the project infrastructure so apologies if I've missed a common pattern here.

hol430:
Ah ok, thanks for the detailed explanation, it's good info for me as I'm not really much of a python programmer. Another option which I also don't really like might be to automatically interpret the speed as percent if it's between 0..100, or RPM if it's >100. It doesn't seem very intuitive though and it also doesn't address the issue of the parameter being named `duty`, which doesn't really describe what it does in this instance.

[...]

Issue #195, opened 2020-10-13
Feature: initialize command gets some values to a config file


MarshallAsch:
I have been enjoying using this program, and I know that I can add a system service or set a script to run on boot, but what are your thoughts and createing a config file `.config/liquidctl`, .liquidctl`, etc. that will keep track of my preferred default LED settings, fan speeds, and pump speeds  and will set those when running `liquidctl initialize`?

If you think that this might be a useful addition I can look into adding this.

jonasmalacofilho:
That's a great idea, but let's do a quick design phase before committing to an implementation.

There's the obvious the question of which format to use.  On that, I've been doing a lot of Rust lately, so [TOML](https://github.com/toml-lang/toml) files are something I would suggest.  Even though I acknowledge that they have some issues, they have a spec of some sort and fairly simple for humans and (it seems) programs.  But we should consider other alternatives.

Besides the serialization format, we need to figure out what schema to use.  With TOML I would probably separate between identifying a device (with vendor/product IDs and/or bus name and addresses, and/or serial numbers) and specifying its initialization options (i.e. options to `initialize` itself) and which other actions to perform (like setting LED settings or fan speeds) by default.

Finally, there should be command line arguments and/or environment variables to disable the use of the config file, or to use a different one than that in the default location.

P.S. A general/global section could also solve #67 by allowing the user to set the legacy mode.

MarshallAsch:
I personally really like the [.ini](https://en.wikipedia.org/wiki/INI_file) like configuration files, [TOML](https://en.wikipedia.org/wiki/TOML), seems to be a formally structured variant of that. I generally don't think that json, or yaml make good configuration files they seem to verbose. 

I didn't think about specify the vendor/product IDs and be able to set `initialize` options for those, but that seems like a good idea since CPU coolers isn't the only type of device that this supports (If it was I would say I don't think that makes much sense because what normal person is going to have more than one AIO in their system). 

Ill try to think of some sort of potential format over the next couple days. 

Would adding this in cause any issues with public API changes within the application, I believe that was mentioned somewhere in a different issue.

jonasmalacofilho:
> I generally don't think that json, or yaml make good configuration files they seem to verbose.

Agreed.  I would argue JSON is not really human writable, and I can never get YAML configs right the first time, since things can work one way but not the other, and it's never that well documented (besides the most trivial cases).

[...]

PR #213, opened 2020-11-08
Added initial support for the corsair commander pro

MarshallAsch:
Closes: #155 

I just noticed now that @hol430 has already started an implementation [here](https://github.com/hol430/liquidctl/tree/commanderpro). oops. I don't really want to duplicate all the work that you did, but a PR was never submitted with the change so. You seemed to have some modes and settings in your implementation that I dont have or that I don't quite have a plan for. Another one of the major differences between my implementation and yours is that I think how I implemented it is more simular to the existing project code structure (based off of what is in the hydro_platinum.py driver).

This implementation is based on the protocol description that is found here https://github.com/audiohacked/OpenCorsairLink/issues/70

Things to do / need to be done
- [X] update supported device list in README
- [X] update udev rules
- [X] get firmware version (in initialize)
- [X] get bootloader version (in initialize)
- [X] get the temperature from each connected temp sensor (in get status)
- [X] identify connected temp and fans (in initialize)
- [X] get fan speeds for all DC and PWM controlled fans
- [X] get voltages for the 12v 5v and 3.3v rails
- [ ] figure out how to tell liquidctl what fan types are connected without having to go to windows to change the setting
- [x] figure out how to tell liquidctl what type of devices are connected to the LED controller, how many LED's their are and what order the devices are in
- [x] add support for setting the fan speed to fixed duty
- [x] add support for setting the fan speed to a fan curve
- [ ] add support for changing brightness
- [x] add support for changing led mode
- [x] automated function tests
- [x] usage description
- [x] documented protocol description ( in addition to what is found [here](https://github.com/audiohacked/OpenCorsairLink/issues/70) and [here](https://gitlab.com/CalcProgrammer1/OpenRGB/-/wikis/Corsair-Lighting-Node-Devices)
- [ ] extra things probably.... 
 

I can stop working on this if @hol430 implementation is preferred.


hol430:
Nice work @MarshallAsch - I'm happy to switch to using this version of the driver if you reckon it integrates better with the existing project structure. You can probably fill in a few of the blanks (set fan speed, change brightness, led mode, etc) by copying from my branch although it's not that hard to implement from scratch anyway once you've got the constants set up and you have some example transmissions captured from the windows driver. Agreed that it would be nice if we could report what devices/fan types are connected and their current state as the windows version does - this is something I never got around to doing.

Let me know if there's any areas I can help out/test/whatever, as I'm still interested in finishing this feature (and using it of course :).

MarshallAsch:
Okay thanks @hol430, Im going to keep working on it then. Im currently partway through the set fan speed command. Im not sure if my approach integrates better with the rest of the code base per say, but I feel like it is more consistent (naming things, how things are defined) which makes it a little bit easier to maintain and add things too. 

When you were testing did you notice anything about the commands that were sent being persistent across reboots (or device restarts) or did the commands need to be sent every time (I know the H100i platinum does not save any information about fan curves between reboots)?

jonasmalacofilho:
@MarshallAsch (and @hol430),

Thank you for collaborating on this!

I probably won't be able to add a lot of value here, since my knowledge about this device is limited to this code and to what's in OpenCorsairLink.  So do you have any questions about how to integrate this in liquidctl?  Or any specific areas of the driver/protocol you would like me to look into?

MarshallAsch:
@jonasmalacofilho There is one thing that I was wondering about how I should go about implementing.

For the set fan curve command the fan speeds are set as RPM values rather than as a percentage. (Im not sure if any of the other drives do that as well) and I was thinking of leaving the interface as setting (temp, duty) pairs and then converting the percentage to an rpm value for the fans but this requires knowledge about what the max rpm is for the fan. 

There is also the setting for fan type for the RGB settings for this device and I was thinking of adding a flag called `--fan-type` and `--temp-sensor` for this these commands. 

How this is implemented might also tie into #195, or something with a longer term persistent settings file.

jonasmalacofilho:
No other drivers handle RPM input yet (but the Corsair Platinum/PRO XT coolers can support it).

Between taking in the fan's maximum RPM and duty values, and taking in the desired RPM directly, I think I prefer the latter: it's simpler and, because it's stateless, less prone to user errors.

```
$ liquidctl -mcommander set fan2 speed 1250 --rpm
```

(this was also @hol430's original proposal... there are [some issues](https://github.com/jonasmalacofilho/liquidctl/issues/155#issuecomment-678359454), but I think it's the best alternative so far)

---

On the RGB side, you're talking about the ML vs. LL having different number of LEDs, right?

Do we need to communicate the fan/accessory type to the Commander Pro?  If the protocol doesn't require it, I think we should leave the fan type out of it, and just treat each channel as having an unspecified number of LEDs (bounded only by the maximum possible number of LEDs on that channel).

This patterns leaves it up to the user to pass in the correct number of colors.  Our other drivers work like this.

MarshallAsch:
Okay that makes sense. I'll implement it using the rpm values for the fan profile command. 

For the LED settings everyones documentation for it seems to specify that the value is sent in one of the messages to the device that specifies the fan types that are connected. Altho this value seems to be equal to the number of led's that the device has at first glance. Once I get to the LED portion Ill confirm that behaviour, on one of the forums there was a comment from a corsair employee / mod (Im not sure which they were) that alluded to the device type does not matter just the number of LED's

hol430:
> When you were testing did you notice anything about the commands that were sent being persistent across reboots (or device restarts) or did the commands need to be sent every time (I know the H100i platinum does not save any information about fan curves between reboots)?

Yeah, these are definitely persistent between reboots.

MarshallAsch:
I have added some basic LED control settings.that I think works I cant test the multi led effects because I only have SP RGB fans connected to my commander pro which have one LED on them. 

@hol430 what is your configuration like maybe some of the other effects will work for you. Im going to take another look through your code and the openRGB project over the weekend to see if I missed things.

Its not documented yet and there are not any regression tests to ensure that the messages being sent are correct. I added a couple of flags and changed the behaviour of the `set <channel> color <mode> <color>` command a little bit where the channel isnt an actual LED channel but rather a device number (added a channel flag for the actual channel).

Something I noticed is that the value that is described as the `the device type` seems to have a value that is equal to the number of LEDs in that device. so putting a different value there changes the behaviour a little bit sometimes giving the actual effect that is desired across multiple fans.

BangerTech:
Hi @MarshallAsch 
thanks for your great work. When do you think the commander pro support will be ready for us "normal" users? 
And does this implementation work with the CORSAIR Lighting Node PRO too? 
thanks again and kindest regards!
Mike

MarshallAsch:
Hi @BangerTech the plan for this is for it to be included in the 1.5.0 release that is slated for February. There are something that I cant actually test because of what led devices I have connected to the commander pro. (If you happen to have devices connected to yours that are not just the SP RGB Fans and are willing to try testing some things that would be awesome.) 

This same driver (with a couple of small modifications) _should_ work with the Lighting Node PRO and the Commander Core, again if you would be okay with test on those devices that would be awesome.

BangerTech:
@MarshallAsch thanks for your fast reply ;-)
ya for sure I would love to test. I'm running the LL120 on the Lighting Node Pro on a Mac Environment. 
Right now I´m having the version 1.4.2 installed. How can I test your version? 
Thanks

MarshallAsch:
@BangerTech  I just quickly made a couple changes so it _should_ detect your device but Im not certain. let me know if any of the steps do or dont work.

But to test you will need to clone the repository and run it: (note some of the flag names that I currently have dont make much sense)

```
$ git clone https://github.com/MarshallAsch/liquidctl.git
$ cd liquidctl
$ git checkout commander-pro
$ sudo python3 -m liquidctl.cli list
   (should see your device here, if not stop)
$ sudo python3 -m liquidctl.cli -m node set led1 color fixed ff0000 --led-type 1 
    (should set the first led on the first fan to red everything else should be off)
$ sudo python3 -m liquidctl.cli -m node set led2 color fixed 00ff00 --led-type 1 
    (should set the first led on the first fan to red and the first led on fan2 should be green)
 ```

That is the behaviour I am currently expecting to occur and It assumes that you have the following setup
```
lighting node pro (channel 1) -> rgb hub -> LL fan
                             -> LL fan
```

If instead the first led is red and the second is green then my understanding of one of the fields is a little off (and I know how to fix it)
and if you could also test. 
```
sudo python3 -m liquidctl.cli -m node set led color clear
sudo python3 -m liquidctl.cli -m node set led1 color rainbow --led-type 16
```
That should set the first fan to a rainbow effect (all of the LED's should not be the same color)

**Comment by BangerTech** [comment]
@MarshallAsch hi and thanks for the repository but I get an error after putting in sudo python3.....
liquidctl % sudo python3 -m liquidctl.cli list
Password:
  File "/usr/local/Cellar/python@3.9/3.9.0_2/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/Cellar/python@3.9/3.9.0_2/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/sentshuashack/liquidctl/liquidctl/cli.py", line 78, in <module>
    from docopt import docopt
ModuleNotFoundError: No module named 'docopt'"
even when I just type liquidctl list I get : zsh: command not found: liquidctl 
what do you think is the problem now
i´m running the recent python 3.9

**Comment by MarshallAsch** [comment]
that is a missing dependancy. you will have to install the docopt package `pip3 install docopt`. that should work. (my mac has a weird version mismatch thing with python and pip for some reason)

**Comment by BangerTech** [comment]
@MarshallAsch yeah now I can see the lighting node but I get this error: ERROR: Unexpected error with Corsair Lighting Node Pro(experimental)
TypeError: '<' not supported between instances of 'str' and 'int'

**Comment by BangerTech** [comment]
ah and yes my setup is lightning node pro channel 1 - hub - LLFan1, LLFan2, LLFan3 (channel 2 is not used yet)

**Comment by MarshallAsch** [comment]
Which command is that error from?

**Comment by BangerTech** [comment]
sorry ;-)
it's from sudo python3 -m liquidctl.cli -m node set led1 color fixed ff0000 --led-type 1
and for :sudo python3 -m liquidctl.cli -m node set led color clear ERROR: Unexpected error with Corsair Lighting Node Pro(experimental)
ValueError: Mode "clear" is not valid

**Comment by MarshallAsch** [comment]
Do you want to chat on discord instead so others don't get spammed with notifications?

**Comment by BangerTech** [comment]
sure what's your @

**Comment by MarshallAsch** [comment]
@BangerTech send me an email. I don't really want to post my discord name publicly. you can get my email from my github profile.

**Comment by jonasmalacofilho** [comment]
I guess it's time for liquidctl to have a Discord channel...

**Comment by MarshallAsch** [comment]
Update as a result of testing with @BangerTech yesterday:

- The LED control protocol for the commander pro is the same as that for the lighting node pro
- The lighting node pro does not have the rail voltage information that the commander pro has
- the field that I thought was the device number is actually the start led number, and the `led type` field is actually a number of led's to set the effect for. (ie. if you have a LL fan and want to set fan number 3 to an effect start needs to be 32 and leds needs to be 16) This will be updated
- something currently seems to be broken with setting the brightness
- other then that it works as intended on both devices. Suspect that it will also work with the lighting node core as well if I can get the product ID for that device.

**Comment by MarshallAsch** [comment]
@jonasmalacofilho Im pretty sure this is now ready for a full review. I have added tests for the new module, documented how to use the new module, added udev rules for the devices, updated completion flags, and added a protocol description in the docs section. There are a couple of "extras" that probably could be included but I think that should be left for future improvements.

File: liquidctl/cli.py [hunk 1]
```diff
@@ -28,13 +28,17 @@
   --time-off <value>          Time to wait with the LED turned off (seconds)
   --alert-threshold <number>  Threshold temperature for a visual alert (°C)
   --alert-color <color>       Color used by the visual high temperature alert
+  --direction <string>        If the pattern should move forward or backwords. [default: forward]
```
jonasmalacofilho, line 31: There's a typo in the description: s/backwords/backwards/.

Please also add an issue for all drivers that need to be updated to support this new flag (in additional to the old and silly `backwards-` prefixed modes).

liquidctl/cli.py
```diff
@@ -28,13 +28,17 @@
...
+  --start-led <number>        The first led to start the effect at (0 offset)
+  --num-leds <number>         The number of LED's the effect shoud apply to
```
jonasmalacofilho, line 33: Perhaps `--maximum-leds` instead of `--num-leds`?  I would like to avoid abbreviations here.

liquidctl/driver/commander_pro.py
```diff
@@ -0,0 +1,514 @@
...
+        may default to 100% duty.
+
+        Up to six (temperature, duty) pairs can be supplied in `profile`,
+        with temperatures in Celsius and duty values in percentage.  The last
+        point should set the fan to 100% duty cycle, or be omitted; in the
+        latter case the fan will be set to max out at 60°C.
+
+
+        there are 6 points on the fan curve, temp, rpm pairs
+        Send one message per fan that is being set
```
jonasmalacofilho, line 340: Apparently some old/informal comment left behind.

liquidctl/cli.py
```diff
@@ -28,13 +28,17 @@
...
 
 Other device options:
   --single-12v-ocp            Enable single rail +12V OCP
   --pump-mode <mode>          Set the pump mode (certain Corsair coolers)
   --legacy-690lc              Use Asetek 690LC in legacy mode (old Krakens)
   --non-volatile              Store on non-volatile controller memory
   --unsafe <features>         Comma-separated bleeding-edge features to enable
+  --temp-sensor <number>      The tempature sensor number for the Commander Pro
```
jonasmalacofilho, line 41: - typo in description: s/tempature/temperature/
- avoid abbreviation: `--temperature-sensor`

liquidctl/driver/commander_pro.py
```diff
@@ -0,0 +1,514 @@
...
+    else:
+        return 'UNKNOWN'
+
+class CommanderPro(UsbHidDriver):
+    """Corsair Commander Pro LED and fan hub"""
+
+    SUPPORTED_DEVICES = [
+        (0x1B1C, 0x0C10, None, 'Corsair Commander Pro (experimental)',
+            {'fan_count': 6, 'temp_probs': 4, 'led_channels': 2}),
+        (0x1B1C, 0x0C0B, None, 'Corsair Lighting Node Pro(experimental)',
```
jonasmalacofilho, line 118: Missing space before parenthesis.

liquidctl/driver/commander_pro.py
```diff
@@ -0,0 +1,514 @@
...
+
+        colors = list(colors)
+        expanded = colors[:3]
+        c = itertools.chain(*((r, g, b) for r, g, b in expanded))
+        colors = list(c)
+
+        direction = kwargs.get('direction', 'forward').lower()
+        speed = kwargs.get('speed', 'medium').lower()
+        start_led = kwargs.get('start_led', 1)
+        num_leds = kwargs.get('num_leds', 1);
```
jonasmalacofilho, line 448: These should be exposed _directly_ as function parameters.  This makes it easier to understand which parameters are expected by a function, as well as helps inspection of the APIs using the Python REPL or pydoc.

liquidctl/driver/commander_pro.py
```diff
@@ -0,0 +1,514 @@
...
+        # need to figure out how to find out what the max rpm is for the given fan
+
+        if self._fan_count == 0:
+            raise NotSupportedByDevice()
+
+        profile = list(profile)
+        profile = _prepare_profile(profile)
+
+        # fan_type = kwargs['fan_type'] # need to make sure this is set
+        temp_sensor = kwargs.get('temp_sensor', 1) # need to make sure this is set and in range 1-4 or ext
```
jonasmalacofilho, line 354: Same issue as mentioned bellow (make this an explicit parameter of the function).

jonasmalacofilho (changes requested):
At a glance almost everything looks good, so I'm delegating the specifics of the Commander Pro to you and basically only focusing on changes the (may) affect other parts of liquidctl.

README.md
```diff
@@ -82,6 +82,7 @@ The following devices are supported by this version of liquidctl.  See each guid
 | AIO liquid cooler | [NZXT Kraken Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_E_</sup> |
 | DDR4 DRAM | [Corsair Vengeance RGB](docs/ddr4-guide.md) | SMBus | <sup>_EUXN_</sup> |
 | DDR4 DRAM | [DIMMs with a standard temperature sensor](docs/ddr4-guide.md) | SMBus | <sup>_EUXN_</sup> |
+| Fan/LED controller | [Corsair Commander Pro](docs/corsair-commander.md) | <sup>_EN_</sup> |
```
jonasmalacofilho, line 85: Missing value for column "bus" ("USB  HID").

README.md
```diff
@@ -82,6 +82,7 @@ The following devices are supported by this version of liquidctl.  See each guid
...
 | Fan/LED controller | [NZXT Grid+ V3](docs/nzxt-smart-device-v1-guide.md) | USB HID | |
 | Fan/LED controller | [NZXT HUE 2, HUE 2 Ambient](docs/nzxt-hue2-guide.md) | USB HID | |
 | Fan/LED controller | [NZXT RGB & Fan Controller](docs/nzxt-hue2-guide.md) | USB HID | |
```
jonasmalacofilho, line 88: Also, missing entry for the Lighting Node Pro.

jonasmalacofilho (changes requested):
Hm, we need to normalize the casing of "Pro" (and, later, "Core").

I honestly think "Pro" is better (manufacturers love to shout), but this means that other drivers need to be adjusted to match this (in all documentation and references, as well as the device's `description`).

jonasmalacofilho (changes requested):
One more thing, the man page needs to be updated with the new options, as well as entries for the Commander Pro and Lighting Node Pro.

liquidctl/cli.py
MarshallAsch: created #253 for this.

liquidctl/cli.py
MarshallAsch: sure I can change it to that.

liquidctl/driver/commander_pro.py
MarshallAsch: removed.

README.md
MarshallAsch: ah, I had thought that was some sort of replacement column for `<sup>_EN_</sup>` and was confused by it. But I just added that in.

MarshallAsch:
I think I have responded to all the comments now and fixed everything that needed fixing.

MarshallAsch:
> Hm, we need to normalize the casing of "Pro" (and, later, "Core").

I think all of my uses of "Pro" are now in the same case, I did not check the other drivers though

liquidctl/cli.py
```diff
@@ -11,43 +11,43 @@
...
-  --legacy-690lc              Use Asetek 690LC in legacy mode (old Krakens)
-  --non-volatile              Store on non-volatile controller memory
-  --unsafe <features>         Comma-separated bleeding-edge features to enable
-  --temp-sensor <number>      The tempature sensor number for the Commander Pro 
+  --single-12v-ocp              Enable single rail +12V OCP
+  --pump-mode <mode>            Set the pump mode (certain Corsair coolers)
+  --legacy-690lc                Use Asetek 690LC in legacy mode (old Krakens)
+  --non-volatile                Store on non-volatile controller memory
+  --unsafe <features>           Comma-separated bleeding-edge features to enable
+  --temperature-sensor <number> The temperature sensor number for the Commander Pro
```
jonasmalacofilho, line 41: docopt requires two spaces between an option and its description.   Even if this still somehow parses (correctly), let's follow the spec.

liquidctl.8
```diff
@@ -152,6 +161,9 @@ Set the pump mode.
 .B \-\-legacy\-690lc
 Use Asetek 690LC in legacy mode (old Krakens).
 .TP
+.B \-\-non\-volatile
```
jonasmalacofilho, line 164: `--temperature-sensor` is missing.

liquidctl.8
```diff
@@ -195,6 +207,37 @@ Internal data used by some drivers.
 .
 .SH DEVICE SPECIFICS
 .
+.SS Corsair Commander Pro
+.SS Corsair Lighting Node Pro
+Cooling channels: \fIfan\fR, \fIfan[1\-5]\fR. (Commander Pro only)
```
jonasmalacofilho, line 212: I missed this before: what does `fan` mean on a Commander Pro?

It it syncs all fans, then we have a convention of calling that (pseudo) channel "sync".

liquidctl.8
```diff
@@ -195,6 +207,37 @@ Internal data used by some drivers.
...
+\fIrainbow2\fR	0
+.TE
+.PP
+When applicable the animation speed can be set with
+.BI \-\-speed= value ,
+where the allowed values are: \fIfast\fR, \fImedium\fR, \fIslow\fR.
+.PP
+When applicable the animation direction can be set with
+.BI \-\-direction= value ,
+where the allowed values are: \fIforward\fR or \fIbackward\fR.
```
jonasmalacofilho, line 240: `--temperature-sensor`, `--start-led` and `--maximum-leds` should also be mentioned in these notes.

You can also omit the second (or later) "When applicable".

jonasmalacofilho (changes requested):
Just a few more minor things.

liquidctl.8
MarshallAsch: hmm. Some of the drivers, (ie the H100i uses the `fan` channel to control both fans. I can change this to sync if you'd prefer.

liquidctl/cli.py
```diff
@@ -11,39 +11,43 @@
...
-  --pump-mode <mode>          Set the pump mode (certain Corsair coolers)
-  --legacy-690lc              Use Asetek 690LC in legacy mode (old Krakens)
-  --non-volatile              Store on non-volatile controller memory
-  --unsafe <features>         Comma-separated bleeding-edge features to enable
+  --single-12v-ocp               Enable single rail +12V OCP
+  --pump-mode <mode>             Set the pump mode (certain Corsair coolers)
+  --legacy-690lc                 Use Asetek 690LC in legacy mode (old Krakens)
+  --non-volatile                 Store on non-volatile controller memory
+  --unsafe <features>            Comma-separated bleeding-edge features to enable
+  --temperature-sensor <number>  The temperature sensor number for the Commander Pro
```
grzechu92, line 41: Use of `--temperature-sensor 2` seems to work correctly (no output), but it's falling back to first sensor. Isn't there any mismatch of parameter names? Because line 107 in this file says `--temp-sensor`, but while trying to use `--temp-sensor` I'm getting help output (seems like some kind of validation when parameter name isn't correct).

Tested on c94d931 build created by CI

grzechu92:
```
PS C:\liquidctl\dev\commander> ./liquidctl --match commander set fan1 speed 0 1 30 1500 --temp-sensor 2
Usage:
  liquidctl [options] list
  liquidctl [options] initialize [all]
  liquidctl [options] status
  liquidctl [options] set <channel> speed (<temperature> <percentage>) ...
  liquidctl [options] set <channel> speed <percentage>
  liquidctl --help
  liquidctl --version

Comment (MarshallAsch) [reply liquidctl/cli.py:41]: Good catch. I missed that change.

Comment (MarshallAsch) [reply liquidctl/cli.py:41]: I think the latest commit has a fix for that.

Comment (grzechu92) [reply liquidctl/cli.py:41]: Works as expected, thanks for extremely fast fix :) I decided to buy Commander Pro when I've noticed that liquidctl integration is in progress, great job! :)

Comment (MarshallAsch) [reply liquidctl/cli.py:41]: good, Im happy it worked as anticipated for you.

File: liquidctl/driver/commander_pro.py [hunk 15]
```diff
@@ -0,0 +1,499 @@
...
+_LED_DIRECTION_BACKWARD = 0x00
+
+_FAN_MODE_DISCONNECTED = 0x00
+_FAN_MODE_DC = 0x01
+_FAN_MODE_PWM = 0x02
+
+
+_PROFILE_LENGTH = 6
+_CRITICAL_TEMPERATURE = 60
+_MAX_FAN_RPM = 5000             # I have no idea if this is a good value or not
```
grzechu92, line 69: Could this be parameterised, made disableable, or ignored when explicitly defining value over 60*C?
I've found usecase when I'd like to set curve with highest point starting at 70*C and having only 1200 out of 1800 rpm because fan is getting too noisy at max rpm. With current code executing:
`./liquidctl --match commander set fan1 speed 40 500 70 1200 100 1250 --temperature-sensor 2`
It will max out the 1800 rpm fan few degrees over 40*C

MarshallAsch: After thinking about this a little bit we decided to add an `--unsafe=high_tempature` flag that will increase the maximum temperature to 100 degrees. In most cases the 60 degrees will be left as a maximum ambient temperature. as an alternative you could add an extra temperature point at 1 degree bellow that max if you want to reduce the fan speed.

If you have any further issues with that it would probably be better to open a new issue for this to be addressed, but for now Im going to mark this as resolved.

Issue #224, opened 2020-11-15
Add support for Pascal graphics cards from EVGA


jonasmalacofilho:
If we're going to start supporting SMBus devices, I might just as well support my own hardware.

In the case of Pascal-era cards, some EVGA cards (e.g. 1080 FTW) don't use the same mechanism as the FE/reference cards.

PR #230, opened 2020-11-22
Add support for SMBus and EVGA GTX 1080 FTW

jonasmalacofilho:
Closes: #224

### DIFF
--- a/docs/developer/style-guide.md
+++ b/docs/developer/style-guide.md
@@ -1,0 +1,120 @@
+# Style guide
+
+> This is not the code; this is just a tribute.
+
+A style guide is meant to make the code more pleasant to look at or modify.
+
+## General guidelines
+
+This section has yet to be written, but for a start...
+
+Read [PEP 8], then immediately watch Raymond Hettinger's [Beyond PEP 8] talk.
+Write code somewhere between those lines.
+
+In this repository, newer drivers are usually better examples than older ones.
+Experience with the domain helps to write better code...  Who would have
+thought that, right?
+
+Try to keep lines around 80-ish columns wide; in some modules 100-ish columns
+may be more suited, but definitively try to avoid going beyond that.
+
+Be consistent within a given module.  Try to be consistent between similar
+modules.
+
+[PEP 8]: https://pep8.org/
+[Beyond PEP 8]: https://www.youtube.com/watch?v=wf-BqAjZb8M
+
+## Use of automatic formatters
+
+Pull requests are welcome.
+
+For a suggestion of a formatter and associated configuration to use, not just
+to fill this section.
+
+## Writing messages
+
+Liquidctl generates a substantial amount of human readable messages, either
+intended for direct consumption by CLI users, or for _a posteriori_ debugging
+by developers.
+
+While discussing the format of these messages deviates from the usual tab ×
+spaces wars we have learned to love [citation needed], it is very important to
+keep the messages clear, objective and consistent.
+
+### Errors
+
+Raise exceptions, preferably `liquidctl.error.*` or Python's
+[built-in exceptions].
+
+[built-in exceptions]: https://docs.python.org/3/library/exceptions.html
+
+### Warnings
+[warnings]: #warnings
+
+Warnings are used to convey information that are deemed always relevant.  They
+should be used to alert of unexpected conditions, degraded states, and other
+high-importance issues.
+
+By default the CLI outputs these to `stderr` for all users.  Since the messages
+are visible to non-developers, the should follow the simplified guidelines
+bellow:
+
+- start with a lowercase letter
+- string values that the user is expected to type, or that absolutely need to
+  delimited: use single quotes
+- other values: no quotes or backticks
+- command line commands or options: no quotes or backticks
+
+Deprecated guidelines for warnings:
+
+- outputting values as `key=value` pairs
+
+### Info
+
+Information messages are normally hidden, but can be enabled with the
+`--verbose` or `--debug` flags.  They should be used to display complementary
+information that may interest any user, like additional `status` items or
+computed `temp × rpm` curves.
+
+Since these messages are also targeted to non-developers, they should follow
+the same simplified guidelines used for [warnings].
+
+### Debug
+
+Debug messages are only intended for developers or users interested in dealing
+with the internals of the program.  In the CLI these are normally disabled and
+require the `--debug` flag.
+
+Since they are intended for internal use, they follow relaxed guidelines:
+
+- start with a lowercase letter
+- use `key=value`, `` `expression` ``, `'string'`, or `value`, whichever is clearer
+
+### A quick refresher on `logging`
+
+Get a logger at the start of the module.
+
+```py
+import logging
+...
+_LOGGER = logging.getLogger(__name__)
+```
+
+Prefer old-style %-formatting, since this is evaluated lazyly by `logging`, and
+the message will only be formated if the logging level is enabled.
+
+```py
+_LOGGER.warning('value %d, expected %d', current, expected)
+```
+
+When writing informational or debug messages, pay attention to the cost of
+computing each value.  A classic example is hex formatting some bytes, which
+can be expensive; this case can be solved by using
+`liquidctl.util.LazyHexRepr`, and other similar wrapper types can be
+constructed for other scenarios.
+
+```py
+from liquidctl.util import LazyHexRepr
+...
+_LOGGER.debug('buffer: %r', LazyHexRepr(some_bytes))
+```
--- a/docs/nvidia-guide.md
+++ b/docs/nvidia-guide.md
@@ -1,0 +1,78 @@
+# NVIDIA graphics cards
+_Driver API and source code available in [`liquidctl.driver.nvidia`](../liquidctl/driver/nvidia.py)._
+
+Support for these cards in only available on Linux.
+
+Additional requirements must also be met:
+
+- optional Python dependency `smbus` is available
+- `i2c-dev` kernel module has been loaded
+- specific unsafe features have been opted in
+- r/w permissions to card-specific `/dev/i2c-*` devices
+
+---
+
+Jump to a specific card:
+
+* _Series 10/Pascal:_
+    - [EVGA GTX 1080 FTW](#evga-gtx-1080-ftw)
+<!-- - [EVGA GTX 1070 FTW](#evga-gtx-1070-1080-ftw) -->
+<!-- - [EVGA GTX 1080 FTW](#evga-gtx-1070-1080-ftw) -->
+
+
+## EVGA GTX 1080 FTW
+<!-- EVGA GTX 1070/1080 FTW -->
+
+Experimental.  Only RGB lighting supported.
+
+Unsafe features:
+
+- `smbus`: enable SMBus support; SMBus devices may not tolerate writes or reads
+  they do not expect
+- `evga_pascal`: enable access to the specific graphics cards
+
+### Initialization
+
+Not required for this device.
+
+### Retrieving the current RGB lighting mode and color
+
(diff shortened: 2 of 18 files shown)


docs/nvidia-guide.md
```diff
@@ -0,0 +1,78 @@
+# NVIDIA graphics cards
+_Driver API and source code available in [`liquidctl.driver.nvidia`](../liquidctl/driver/nvidia.py)._
+
+Support for these cards in only available on Linux.
+
+Additional requirements must also be met:
+
+- optional Python dependency `smbus` is available
+- `i2c-dev` kernel module has been loaded
```
jonasmalacofilho, line 9: I should prepare and include a "load the `i2c-dev` module during boot" configuration file for inclusion in distro packages.

docs/nvidia-guide.md
```diff
@@ -0,0 +1,78 @@
...
+- specific unsafe features have been opted in
+- r/w permissions to card-specific `/dev/i2c-*` devices
```
jonasmalacofilho, line 11: I also forgot to add a new rule for the EVGA card in [`extra/linux/71-liquidctl.rules`](https://github.com/jonasmalacofilho/liquidctl/blob/master/extra/linux/71-liquidctl.rules).

jonasmalacofilho:
@MarshallAsch 

I know there are a few changes that shouldn't really be part of this PR (despite being inspired by it), sorry about that!  Feel free to ignore those files, but please let me know by adding a "Ignored." comment to the top of each ignored file.

docs/developer/style-guide.md
```diff
@@ -0,0 +1,120 @@
...
+Prefer old-style %-formatting, since this is evaluated lazyly by `logging`, and
+the message will only be formated if the logging level is enabled.
+
+```py
+_LOGGER.warning('value %d, expected %d', current, expected)
+```
+
+When writing informational or debug messages, pay attention to the cost of
+computing each value.  A classic example is hex formatting some bytes, which
+can be expensive; this case can be solved by using
```
MarshallAsch, line 112: Interesting, I didn't think about that but it makes sense.

docs/nvidia-guide.md
MarshallAsch: That device is also included PR #228 (I know it should have probably been here but, not a big deal)

docs/nvidia-guide.md
```diff
@@ -0,0 +1,78 @@
...
+
+---
+
+Jump to a specific card:
+
+* _Series 10/Pascal:_
+    - [EVGA GTX 1080 FTW](#evga-gtx-1080-ftw)
+<!-- - [EVGA GTX 1070 FTW](#evga-gtx-1070-1080-ftw) -->
+<!-- - [EVGA GTX 1080 FTW](#evga-gtx-1070-1080-ftw) -->
```
MarshallAsch, line 20: Is this line a copy paste error or are those commented out lines different?

liquidctl/cli.py
```diff
@@ -33,6 +33,7 @@
   --single-12v-ocp            Enable single rail +12V OCP
   --pump-mode <mode>          Set the pump mode (certain Corsair coolers)
   --legacy-690lc              Use Asetek 690LC in legacy mode (old Krakens)
+  --non-volatile              Store on non-volatile controller memory
```
MarshallAsch, line 36: We will need to remember to add this to the bash completion script in #205

liquidctl/driver/nvidia.py
```diff
@@ -0,0 +1,171 @@
...
+        RAINBOW = (0b010, 0)
+        BREATHING = (0b101, 1)
+
+        def __str__(self):
+            return self.name.capitalize()
+
+    @classmethod
+    def probe(cls, smbus, vendor=None, product=None, address=None, match=None,
+              release=None, serial=None, **kwargs):
+        GTX_1080_FTW = 0x6286
```
MarshallAsch, line 53: How come this value is here and not at the top with the other constants?

liquidctl/driver/nvidia.py
```diff
@@ -0,0 +1,171 @@
...
+        def __new__(cls, value, required_colors):
+            obj = bytes.__new__(cls, [value])
+            obj._value_ = value
+            obj.required_colors = required_colors
+            return obj
+
+        OFF = (0b000, 0)
+        FIXED = (0b001, 1)
+        RAINBOW = (0b010, 0)
+        BREATHING = (0b101, 1)
```
MarshallAsch, line 45: Are binary values preferred for these constants to hex values?

liquidctl/driver/nvidia.py
```diff
@@ -0,0 +1,171 @@
+"""liquidctl drivers for NVIDIA graphics cards.
```
MarshallAsch, line 1: You sort of mentioned this in one of your messages but why is this driver called `nvidia`? Its intention is not to store the drivers for all NVIDIA graphics cards in this file is it?

liquidctl/driver/nvidia.py
```diff
@@ -0,0 +1,171 @@
...
+
+        It is possible to store them in non-volatile controller memory by
+        passing `non_volatile=True`.  But as this memory has some unknown yet
+        limited maximum number of write cycles, volatile settings are
+        preferable, if the use case allows for them.
+        """
+
+        check_unsafe('smbus', 'evga_pascal', error=True, **kwargs)
+
+        channel = 'led'
```
MarshallAsch, line 132: should the channel name that is given be checked against the value 'led'? Right now channel can be set to anything and is ignored.

liquidctl/driver/smbus.py
```diff
@@ -0,0 +1,275 @@
...
+
+        def open(self):
+            """Open the I²C bus."""
+            if not self._smbus:
+                try:
+                    self._smbus = SMBus(self._number)
+                except FileNotFoundError:
+                    if Path('/sys/class/i2c-dev').exists():
+                        raise
+                    raise OSError('kernel module i2c-dev not loaded') from None
```
MarshallAsch, line 102: why is this error `from None` I'm not all too familiar with that syntax. From what I Googled doesn't that hide the original exception?

liquidctl/driver/smbus.py
```diff
@@ -0,0 +1,275 @@
...
+        return list(devs)
+
+    def __init__(self, smbus, description, vendor_id=None, product_id=None,
+                 address=None, **kwargs):
+        assert vendor_id and product_id and address is not None
+
+        self._smbus = smbus
+        self._description = description
+        self._vendor_id = vendor_id
+        self._product_id = product_id
```
MarshallAsch, line 220: Should a note be added somewhere that these id's are actually the subsystem vendor and the subsystem product to remove some confusion when people are trying to make other drivers later?

docs/developer/style-guide.md
```diff
@@ -0,0 +1,120 @@
...
+- use `key=value`, `` `expression` ``, `'string'`, or `value`, whichever is clearer
+
+### A quick refresher on `logging`
+
+Get a logger at the start of the module.
+
+```py
+import logging
+...
+_LOGGER = logging.getLogger(__name__)
```
MarshallAsch, line 100: I think this recommended format of the logged name needs to be re-factored in a few of the other drivers as well, but that can be a separate PR.

liquidctl/driver/smbus.py
jonasmalacofilho: I also only found out about this yesterday, despite it being documented in the introduction of [Built-in Exceptions](https://docs.python.org/3/library/exceptions.html).

TL;DR: when using `from None` the original exception is kept in `__context__`, but is otherwise hidden; this avoids the display of double tracebacks.

liquidctl/driver/nvidia.py
jonasmalacofilho: We could emit a warning here, but since it's unlikely to ever become a hard error (we're not going to find/expose new LED channels), I don't think it's necessary.

The reason why it wouldn't be a hard error is that I tend to prefer automatically correcting user input when that's possible/safe.  It's the same idea with when the user supplies too many colors.

liquidctl/driver/nvidia.py
jonasmalacofilho: The basic issue is that I don't think it's sensible to have a bunch of single-driver modules, especially when there are significant similarities between the drivers.  _(Module-hell is also not idiomatic in Python)._

Eventually the we'll need to have one module for each "generation", so each module has one to ten drivers (each board partner has a few variants of each GPU, sometimes with different protocols).  But since we're going to be much more strict with adding these drivers, because of the increased associated risks, I'm betting it's going to take a long time until we have more than ten NVIDIA drivers in total.

For now I think we should keep all of them in a single `nvidia` module.  When that starts to get too big we'll split it into `pascal`, `turing`, `ampere`, etc., and at that point `nvidia` can simply re-export all drivers in a single namespace, which is more convenient for API users and also retains backwards compatibility.

liquidctl/driver/nvidia.py
jonasmalacofilho: I had noticed a pattern and though maybe some additional modes could be uncovered.  But that didn't turn out to be the case, so I'll switch to hex for simplicity.

liquidctl/driver/nvidia.py
jonasmalacofilho: I guess it's a personal preference: the value is not needed outside of this function, so I decided to keep it local.

docs/nvidia-guide.md
jonasmalacofilho: It's intentional.  In this example of adding support for the 1070 FTW,

```
- [EVGA GTX 1080 FTW](#evga-gtx-1080-ftw)
```

is replaced with

```
- [EVGA GTX 1070 FTW](#evga-gtx-1070-1080-ftw)
- [EVGA GTX 1080 FTW](#evga-gtx-1070-1080-ftw)
```

The idea is to have both cards of the same "family" (protocol/driver) linking to the same "EVGA GTX 1070/1080 subsection".

liquidctl/driver/smbus.py
jonasmalacofilho: Added.

docs/developer/style-guide.md
jonasmalacofilho: I think most are already doing that, but the logger instance is implicitly public (`LOGGER` instead of `_LOGGER`).

It's arguable whether we can change that now, but we can try...

More generally, the drivers are not consistent in coding-style, and that should be tackled sometime.

jonasmalacofilho:
Thanks for the review!

For my part merging is currently blocked by:

- [ ] missing tests
- [ ] missing udev rules (will use yours, thanks!)
- [ ] missing `i2c-dev` loading shenanigans

liquidctl/driver/nvidia.py
MarshallAsch: That makes sense.

liquidctl/driver/smbus.py
MarshallAsch: Would that be something that should be refactored elsewhere in other modules?

---
I will test that out a little to see how it shows expresses itself to learn that.

liquidctl/driver/nvidia.py
```diff
@@ -0,0 +1,171 @@
+"""liquidctl drivers for NVIDIA graphics cards.
+
+Copyright (C) 2020–2020  Jonas Malaco and contributors
+SPDX-License-Identifier: GPL-3.0-or-later
+"""
+
+from enum import Enum, unique
+import logging
+import re
```
MarshallAsch, line 9: unused import.

liquidctl/driver/smbus.py
jonasmalacofilho: Yes, and we should also switch to custom exceptions when applicable.

Issue #222, opened 2020-11-14
Add Support for the Asus ROG STRIX RTX2080TI O11G


MarshallAsch:
This device does not use a `usb` interface, but it uses a `i2c/smbus` interface. 
As per discussions with @jonasmalacofilho this is a good starting point for adding in `i2c/smbus` devices.
The protocol for this device is described in the [openRGB project](https://gitlab.com/CalcProgrammer1/OpenRGB/-/wikis/ASUS-Aura-GPU) with a c++ implementation [here](https://gitlab.com/CalcProgrammer1/OpenRGB/-/blob/master/Controllers/AuraGPUController).

txf-:
I'll just throw out my 2 cents here, and ask if it wouldn't make more sense just to support the openRGB SDK? If you're going to be adding smbus support just to modify RGB values on graphics cards, isn't that expanding the scope of liquidctl, replicating functionality present in an existing well maintained project.

Or is the goal to have the same functionality implemented purely in python?

MarshallAsch:
@txf- I hadn't considered using the openRGB SDK to do some of the work for this project. (Although we are aware of that project). My main reasons for originally selecting this project over that project, from a user perspective, was wanting a cli to do the things rather then a GUI (as far as I know openRGB is just an SDK and a GUI) and because openRGB does not run on any of my systems for some reason, and after a quick attempt with running the python SDK I got the same unsuccessful result. Those are my reasons, Im not sure about @jonasmalacofilho's reasons though.

txf-:
Yeah, makes sense. I hope I didn't come across as being too pushy.

MarshallAsch:
All good, asking why we are "reinventing the wheel" is a perfectly valid question.

txf-:
I'm liable to need both projects installed, one to control cooling and another to do the lighting of various systems. I don't want to end up in a position where I need x number of different programs running to handle all that stuff, something ostensibly both programs exist to remedy.

Maybe I'm only worrying about a hypothetical.

MarshallAsch:
> I don't want to end up in a position where I need x number of different programs running to handle all that stuff, something ostensibly both programs exist to remedy.

That makes a lot of sense Its why I have been adding support for my devices to this project

What devices do you currently have / which program do you use to control them?

If you prefer to use the liquidctl program you can add an issue for a new device on here

txf-:
will do. Thanks.

jonasmalacofilho:
@txf- 

[...]

PR #233, opened 2020-11-26
ASUS Strix 2080 TI OC

MarshallAsch:
Closes: #222 

Added the ASUS Strix 2080 Ti OC driver that is currently working for me along with documentation on how to use the device.
- added the ASUS 2080ti rev A. and the EGVA 1080 FTW to the udev rule list

This is the same as #228 that accidentally got closed when the destination branch got deleted.

Steps to do before this can be merged:

- [x] Move driver over to the `nvidia.py` driver
- [x] Create automated tests.


MarshallAsch:
@jonasmalacofilho This is now ready for review, all of the tasks that I still had to complete should be resolved.

jonasmalacofilho:
@MarshallAsch reviewing...

Note: I'm making some small fixes myself, please don't make new changes for the time being.

jonasmalacofilho (approved):

jonasmalacofilho:
Please take a look and see if you spot any problems with the changes in 7899412.

---

Apart from that, I also removed the assertions on volatile==true in the test cases (8ee1705).  I'll adjust that accordingly (and, if necessary, reintroduce them) once I better understand when volatile changes persist (see discord).

The corresponding docs have `FIXME` for the same reason.

docs/nvidia-guide.md
```diff
@@ -85,10 +88,10 @@ Experimental. Only RGB lighting supported.
 
 Unsafe features:
 
-- `smbus`: enable SMBus support; SMBus devices may not tolerate writes or reads they do not expect
+- `smbus`: enable SMBus support; SMBus devices may not tolerate writes or reads
+  they do not expect
 - `rog_turing`: enable access to the specific graphics cars
```
MarshallAsch, line 93: spelling. `cars` -> `cards`.

extra/linux/71-liquidctl.rules
```diff
@@ -17,15 +17,14 @@
...
-# NVIDIA ASUS Strix 2080 Ti OC GPU
-KERNEL=="i2c-[0-9]*", ATTRS{vendor}=="0x10de", ATTRS{subsystem_vendor}=="0x1043", ATTRS{device}=="0x1e07", ATTRS{subsystem_device}=="0x866a", TAG+="uaccess"
-
-# NVIDIA EVGA GTX 1080 FTW
-KERNEL=="i2c-[0-9]*", ATTRS{vendor}=="0x10de", ATTRS{subsystem_vendor}=="0x3842", ATTRS{device}=="0x1b80", ATTRS{subsystem_device}=="0x6286", TAG+="uaccess"
-
 # Asetek 690LC (also EVGA CLC or NZXT Kraken)
 SUBSYSTEMS=="usb", ATTRS{idVendor}=="2433", ATTRS{idProduct}=="b200", TAG+="uaccess"
 
+# ASUS Strix 2080 Ti OC
```
MarshallAsch, line 23: Should the usb and the i2c rules be intermixed or would it be better to have sections, for USB and for i2c? To make it easier if someone only wants some rules and not all.

liquidctl/driver/nvidia.py
```diff
@@ -1,7 +1,6 @@
 """liquidctl drivers for NVIDIA graphics cards.
-liquidctl drivers for ASUS Strix RTX 2080 Ti OC graphics card.
 
-Copyright (C) 2020–2020  Jonas Malacoi, Marshall Asch and contributors
+Copyright (C) 2020–2020  Jonas Malaco, Marshall Asch and contributors
```
MarshallAsch, line 3: oops, vim mistake, it seems I was already in insert mode.

liquidctl/driver/nvidia.py
```diff
@@ -41,24 +51,16 @@ class EvgaPascal(SmbusDriver):
...
-        def __new__(cls, value, required_colors):
-            obj = bytes.__new__(cls, [value])
-            obj._value_ = value
-            obj.required_colors = required_colors
-            return obj
-
+    class Mode(_ModeEnum):
         OFF = (0x00, 0)
         FIXED = (0x01, 1)
         RAINBOW = (0x02, 0)
```
MarshallAsch, line 57: I noticed for your GPU there is no mode `0x03` or `0x04`, is it possible that those exist and are a fade / flash mode? It seems odd that they just skipped a couple numbers

liquidctl/driver/nvidia.py
```diff
@@ -238,32 +230,34 @@ def probe(cls, smbus, vendor=None, product=None, address=None, match=None,
...
+                        val2 = smbus.read_byte_data(address, 0x21)
+                    except:
+                        pass
+                    smbus.close()
+
+                    if val1 << 8 | val2 == ASUS_GPU_MAGIC_VALUE:
+                        selected_address = address
+                        break
+            else:
+                selected_address = cls._SENTINEL_ADDRESS
```
MarshallAsch, line 252: What is the reason for `_SENTINEL_ADDRESS`? I get that it is a default, but not sure why it has that value, and does that name have any significance?

liquidctl/driver/nvidia.py
jonasmalacofilho: I though so too, but there doesn't seem to be nothing there.  Trying to fill in this gap is also why before I had these encoded in bits.

liquidctl/driver/nvidia.py
jonasmalacofilho: I described it in the commit message: 7899412.

TL; DR. To prevent a card opened without `unsafe` to later be used with `unsafe`, and thus access an address we didn't actually validate.

The sentinel is checked for in `get_status` and `set_color`, and it also has a large value so that even if we forget, the lower level APIs will still probably refuse it.

extra/linux/71-liquidctl.rules
jonasmalacofilho: You're right, that would be better.

Issue #137, opened 2020-06-19
Support for Corsair PRO (sixth-generation Asetek) coolers


jonasmalacofilho:
Implement support for the Corsair PRO coolers, likely by porting the corresponding OCL code.

These were marketed as:

* Corsair H100i PRO (`1b1c:0c15`)
* Corsair H115i PRO (`1b1c:0c13`)
* Corsair H150i PRO (`1b1c:0c12`)

Don't confuse with PRO XT or Platinum coolers.  The latter keyword appears in the `iProduct` descriptor, but that doesn't reflect how the device was sold, its hardware platform or the protocol in use.

Pull requests are encouraged!  For a starting point, see [#129 (comment)](https://github.com/jonasmalacofilho/liquidctl/issues/129#issuecomment-640258429).

---

Replaces: #106 ("H115i Pro Support")
Replaces: #136 ("Cannot bind H100i to usbhid on Linux")

jlask:
I just wanted to report where I am stuck with my Corsair H100i Pro and liquidctl.
I updated the master branch code to support it (mimic the H115i) and it now lists it.

int@mint-Z390-AORUS-PRO-WIFI:~/tools/liquidctl$ sudo liquidctl list
Device ID 0: Corsair Hydro H100i Pro

However if I try to initialize it, etc, I always get this error:
[https://pastebin.com/znx2Kkw1](url)

BTW - It does work fine on OCL.

jonasmalacofilho:
This makes sense.  If you check OCL you'll see that the sixth-generation coolers use a different driver than the fifth-generation ones.

Additionally, you have a mistake in one of the `LOGGER.debug` calls you added.

jlask:
I've made progress converting OCL over to liquidctl for my H100i Pro (sixth-generation cooler).
I have the led, pump mode, and temperature alerts all working now.
I have not testing any fan control with OCL or liquidctl so that still needs to be worked on.
I'm also not up to speed on all of the duty/speed settings so that could take awhile.

I'm finding that there may be enough differences in this model where it might be better to create a new "asetekpro.py" driver but for now I have logic in the "asetek.py" driver.
One difference is that there is only one mode i am aware of to setting temperature monitoring instead of setting a high threshold for each led mode.
The pump modes are more like coolit devices, where it has Quiet, Balanced, and Performance.

I am also finding that I do not seem to need to "initialize" the device, no need for the "begin transaction".
Do you see any issues with not using those commands?

[...]

PR #234, opened 2020-11-27
Initial support for Corsair PRO coolers

andytank123:
Implementing initial support for Corsair PRO coolers #137 

This has been testing on the H115i PRO thanks to @joaqim, I will test the change on a H100i PRO once I have access to the hardware.

### DIFF
--- a/liquidctl/driver/asetek_pro.py
+++ b/liquidctl/driver/asetek_pro.py
@@ -33,24 +33,26 @@ _READ_LENGTH = 32
 _READ_TIMEOUT = 2000
 _WRITE_TIMEOUT = 2000
 
-_PRO_WRITE_ENDPOINT = 0x1
-_PRO_READ_ENDPOINT = 0x81
-_PRO_CMD_READ_FIRMWARE = 0xaa
-_PRO_CMD_READ_AIO_TEMP = 0xa9
-_PRO_CMD_READ_FAN_SPEED = 0x41
-_PRO_CMD_WRITE_FAN_CURVE = 0x40
-_PRO_CMD_WRITE_FAN_SPEED = 0x42
-_PRO_CMD_READ_PUMP_SPEED = 0x31
-_PRO_CMD_READ_PUMP_MODE = 0x33
-_PRO_CMD_WRITE_PUMP_MODE = 0x32
-_PRO_CMD_WRITE_COLOR_LIST = 0x56
-_PRO_PUMP_MODES = [
+_WRITE_ENDPOINT = 0x1
+_READ_ENDPOINT = 0x81
+_CMD_READ_FIRMWARE = 0xaa
+_CMD_READ_AIO_TEMP = 0xa9
+_CMD_READ_FAN_SPEED = 0x41
+_CMD_WRITE_FAN_CURVE = 0x40
+_CMD_WRITE_FAN_SPEED = 0x42
+_CMD_READ_PUMP_SPEED = 0x31
+_CMD_READ_PUMP_MODE = 0x33
+_CMD_WRITE_PUMP_MODE = 0x32
+_CMD_WRITE_COLOR_LIST = 0x56
+_PUMP_MODES = [
     'quiet',
     'balanced',
     'performance'
     ]
 
-_PRO_SPEEDS = {
+_COLOR_SPEEDS = ['slower', 'normal', 'faster']
+
+_COLOR_SPEEDS_VALUES = {
     'rainbow': [
         0x30, # slow
         0x18, # medium
@@ -73,7 +75,7 @@ _PRO_SPEEDS = {
     ]
 }
 
-_PRO_COLOR_CHANGE_MODES = {
+_COLOR_CHANGE_MODES = {
     'rainbow': [0x55, 0x01],
     'alert': [],
     'shitf': [0x55, 0x01],
@@ -86,17 +88,21 @@ _PRO_COLOR_CHANGE_MODES = {
 class CorsairAsetekProDriver(_CommonAsetekDriver):
     """liquidctl driver for Corsair-branded sixth generation Asetek coolers."""
     SUPPORTED_DEVICES = [
-        (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {}),
-        (0x1b1c, 0x0c13, None, 'Corsair Hydro H115i Pro', {}),
-        (0x1b1c, 0x0c14, None, 'Corsair Hydro H110i Pro', {}),
-        (0x1b1c, 0x0c15, None, 'Corsair Hydro H100i Pro', {}),
-        (0x1b1c, 0x0c16, None, 'Corsair Hydro H80i Pro', {}),
+        (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {'fan_count': 3}),
+        (0x1b1c, 0x0c13, None, 'Corsair Hydro H115i Pro', {'fan_count': 2}),
+        (0x1b1c, 0x0c15, None, 'Corsair Hydro H100i Pro', {'fan_count': 2})
     ]
+
+    def __init__(self, device, description, fan_count=0, **kwargs):
+        super().__init__(device, description, **kwargs)
+        self._data = None
+        self._fan_count = fan_count
+
 
     def _write(self, data):
         """Write data to the AIO and log"""
         _LOGGER.debug('write %s', LazyHexRepr(data))
-        self.device.write(_PRO_WRITE_ENDPOINT, data, _WRITE_TIMEOUT)
+        self.device.write(_WRITE_ENDPOINT, data, _WRITE_TIMEOUT)
 
     # Not calling this (or begin transaction) since they do not seem to be needed for H100i Pro
     def _end_transaction_and_read(self, length=_READ_LENGTH):
@@ -107,15 +113,14 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         seem to work fine without this complexity; we currently try the same
         approach.
         """
-        msg = self.device.read(_PRO_READ_ENDPOINT, length, _READ_TIMEOUT)
+        msg = self.device.read(_READ_ENDPOINT, length, _READ_TIMEOUT)
         _LOGGER.debug('received %s', LazyHexRepr(msg))
         self.device.release()
         return msg
 
     def _write_color_change_speed(self, mode, speed):
         """Send the speed to cycle colors on the RGB pump"""
-        speed = clamp(speed, 1, 3)
-        self._write([0x53, _PRO_SPEEDS[mode][int(speed)-1]])
+        self._write([0x53, _COLOR_SPEEDS_VALUES[mode][_COLOR_SPEEDS.index(speed)]])
         self._end_transaction_and_read(3)
 
     def _write_color(self, colors):
@@ -123,70 +128,84 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         value used for cycle speed"""
         setColors = list(itertools.chain(*colors))
         colorCount = int(len(setColors) / 3)
-        self._write([_PRO_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
+        self._write([_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
         self._end_transaction_and_read(3)
 
     def _get_fan_speeds(self):
         """Read the RPM speed of the fans. Try to get the speeds for 3 fans as that is
         how many the H150i PRO supports"""
         speeds = []
-        for i in range(3):
-            self._write([_PRO_CMD_READ_FAN_SPEED, i])
+        for i in range(self._fan_count):
+            self._write([_CMD_READ_FAN_SPEED, i])
             msg = self._end_transaction_and_read(6)
             if msg[0] != 0x41 or msg[1] != 0x12 or msg[2] != 0x34 or msg[3] != i:
                 _LOGGER.debug('Unable to get speed for fan id %d', i)
+                speeds.append(-1)
                 continue
             speeds.append((i, (msg[4] << 8) + msg[5]))
         return speeds
 
     def _get_fan_indexes(self, channel):
         if len(channel) > 3:
-            return [int(channel[3:]) - 1]
-        return [reading[0] for reading in self._get_fan_speeds()]
+            channel_num = int(channel[3:])
+            if channel_num > self._fan_count:
+                raise ValueError(f'Unknown channel: {channel}')
+            return [channel_num - 1]
+        return range(self._fan_count)
 
     def initialize(self, pump_mode='balanced', **kwargs):
         """Initialize the device."""
         _LOGGER.debug('Pro configure device...')
         _LOGGER.debug('pump_mode %s', pump_mode)
-        self._write([_PRO_CMD_WRITE_PUMP_MODE, _PRO_PUMP_MODES.index(pump_mode.lower())])
+        self._write([_CMD_WRITE_PUMP_MODE, _PUMP_MODES.index(pump_mode.lower())])
         self._end_transaction_and_read(5)
 
     def get_status(self, **kwargs):
         """Get a status report.
         Returns a list of `(property, value, unit)` tuples.
         """
-        status = []
-        self._write([_PRO_CMD_READ_AIO_TEMP])
+        # Liquid temperature
+        self._write([_CMD_READ_AIO_TEMP])
         msg = self._end_transaction_and_read(6)
         aio_temp = msg[3] + msg[4]/10
-        status.append(('Liquid temperature', aio_temp, '°C'))
         speeds = self._get_fan_speeds()
-        for reading in speeds:
-            fan_num = reading[0] + 1
-            status.append((f'Fan {fan_num} speed', reading[1], 'rpm'))
-        self._write([_PRO_CMD_READ_PUMP_MODE])
+        # Pump mode
+        self._write([_CMD_READ_PUMP_MODE])
         msg = self._end_transaction_and_read(4)
         pump_mode = msg[3]
-        status.append(('Pump mode', _PRO_PUMP_MODES[pump_mode], ""))
-        self._write([_PRO_CMD_READ_PUMP_SPEED])
+        # Pump Speed
+        self._write([_CMD_READ_PUMP_SPEED])
         msg = self._end_transaction_and_read(5)
         pump_speed = (msg[3] << 8) + msg[4]
-        status.append(('Pump speed', pump_speed, 'rpm'))
-        self._write([_PRO_CMD_READ_FIRMWARE])
+        # Firmware
+        self._write([_CMD_READ_FIRMWARE])
         msg = self._end_transaction_and_read(7)
         firmware = '{}.{}.{}.{}'.format(*tuple(msg[3:7]))
-        status.append(('Firmware version', firmware, ''))
-        return status
+        status = [('Liquid temperature', aio_temp, '°C')]
+        for i in range(len(speeds)):
+            speed = speeds[i] if speeds[i] >= 0 else 'Error'
+            status.append((f'Fan {i+1} speed', speed, 'rpm'))
+        return status + [
+            ('Pump mode', _PUMP_MODES[pump_mode], ""),
+            ('Pump speed', pump_speed, 'rpm'),
+            ('Firmware version', firmware, '')
+        ]
 
-    def set_color(self, channel, mode, colors, time_per_color=1, time_off=None,
-                  alert_threshold=_HIGH_TEMPERATURE, alert_color=[255, 0, 0],
-                  speed=3, temps=['35', '45', '55'], **kwargs):
+    def set_color(self, channel, mode, colors, speed='faster',
+                  temps=['35', '45', '55'], **kwargs):
         """Set the color mode for a specific channel."""
-        if mode not in _PRO_COLOR_CHANGE_MODES:
-            raise KeyError(f'Unknown lighting mode {mode}')
+        speed = speed.lower()
+        if mode not in _COLOR_CHANGE_MODES:
+            valid = ', '.join(_COLOR_CHANGE_MODES.keys())
+            raise KeyError(f'Unknown lighting mode {mode}, should be one of {valid}')
+        if speed not in _COLOR_SPEEDS:
+            valid = ', '.join(_COLOR_SPEEDS)
+            raise KeyError(f'Unknown speed {speed}, should be one of {valid}')
+        if channel != 'pump':
+            raise KeyError('Cosair PRO only supports setting the color on the pump')
         colors = list(colors)
         _LOGGER.debug('color count = %d', len(colors))
-        _LOGGER.debug('color change speed %d', int(speed))
+        _LOGGER.debug('color change speed %s', speed)
         if mode == 'alert':
             self._write([0x5f, int(temps[0]), 0x00, int(temps[1]), 0x00, int(temps[2]), 0x00] + colors[0] + colors[1] + colors[2])
             self._end_transaction_and_read(6)
@@ -199,7 +218,7 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
             self._write_color(colors)
         if mode != 'fixed':
             self._write_color_change_speed(mode, speed)
-        self._write(_PRO_COLOR_CHANGE_MODES[mode])
+        self._write(_COLOR_CHANGE_MODES[mode])
         self._end_transaction_and_read(3)
 
     def set_speed_profile(self, channel, profile, **kwargs):
@@ -208,7 +227,7 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         if channel == 'pump':
             raise NotSupportedByDevice()
         if not channel.startswith('fan'):
-            raise KeyError(f'Unknown channel: {channel}')
+            raise ValueError(f'Unknown channel: {channel}')
         adjusted = self._prepare_profile(profile, 0, 100, _MAX_PROFILE_POINTS)
         for temp, duty in adjusted:
             _LOGGER.info('setting %s PWM point: (%i°C, %i%%), device interpolated',
@@ -217,7 +236,7 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         # Need to write curve for each fan in channel
         for i in self._get_fan_indexes(channel):
             _LOGGER.info('setting speed for fan %d', i+1)
-            self._write([_PRO_CMD_WRITE_FAN_CURVE, i] + temps + duties)
+            self._write([_CMD_WRITE_FAN_CURVE, i] + temps + duties)
             self._end_transaction_and_read(32)
 
     def set_fixed_speed(self, channel, duty, **kwargs):
@@ -228,7 +247,7 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
             # Need to write curve for each fan in channel
             for i in self._get_fan_indexes(channel):
                 _LOGGER.info('setting speed for fan %d to %d', i+1, duty)
-                self._write([_PRO_CMD_WRITE_FAN_SPEED, i, duty])
+                self._write([_CMD_WRITE_FAN_SPEED, i, duty])
                 self._end_transaction_and_read(32)
         elif channel == 'pump':
             raise NotSupportedByDevice()
--- a/liquidctl/driver/asetek_pro.py
+++ b/liquidctl/driver/asetek_pro.py
@@ -101,7 +101,6 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
 
     def _write(self, data):
         """Write data to the AIO and log"""
-        _LOGGER.debug('write %s', LazyHexRepr(data))
         self.device.write(_WRITE_ENDPOINT, data, _WRITE_TIMEOUT)
 
     # Not calling this (or begin transaction) since they do not seem to be needed for H100i Pro
@@ -114,7 +113,6 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         approach.
         """
         msg = self.device.read(_READ_ENDPOINT, length, _READ_TIMEOUT)
-        _LOGGER.debug('received %s', LazyHexRepr(msg))
         self.device.release()
         return msg
 
--- a/liquidctl/driver/asetek_pro.py
+++ b/liquidctl/driver/asetek_pro.py
@@ -20,10 +20,10 @@ from liquidctl.driver.asetek import Modern690Lc
 from liquidctl.driver.usb import UsbDriver
 from liquidctl.error import NotSupportedByDevice
 from liquidctl.keyval import RuntimeStorage
-from liquidctl.util import clamp
+from liquidctl.util import clamp, LazyHexRepr
 
 
-LOGGER = logging.getLogger(__name__)
+_LOGGER = logging.getLogger(__name__)
 
 _MAX_PROFILE_POINTS = 7
 _CRITICAL_TEMPERATURE = 60
@@ -41,6 +41,7 @@ _PRO_CMD_WRITE_FAN_SPEED = 0x40
 _PRO_CMD_READ_PUMP_SPEED = 0x31
 _PRO_CMD_READ_PUMP_MODE = 0x33
 _PRO_CMD_WRITE_PUMP_MODE = 0x32
+_PRO_CMD_WRITE_COLOR_LIST = 0x56
 _PRO_PUMP_MODES = [
     'Quiet',
     'Balanced',
@@ -70,6 +71,15 @@ _PRO_SPEEDS = {
     ]
 }
 
+_PRO_COLOR_CHANGE_MODES = {
+    'rainbow': [0x55, 0x01],
+    'alert': [],
+    'shitf': [0x55, 0x01],
+    'pulse': [0x52, 0x01],
+    'blinking': [0x58, 0x01],
+    'fixed': [0x55, 0x01]
+}
+
 
 class CorsairAsetekProDriver(Modern690Lc):
     """liquidctl driver for Corsair-branded sixth generation Asetek coolers."""
@@ -83,7 +93,7 @@ class CorsairAsetekProDriver(Modern690Lc):
 
     def _write(self, data):
         """Write data to the AIO and log"""
-        LOGGER.debug('write %s', ' '.join(format(i, '02x') for i in data))
+        _LOGGER.debug('write %s', LazyHexRepr(data))
         self.device.write(_PRO_WRITE_ENDPOINT, data, _WRITE_TIMEOUT)
 
     # Not calling this (or begin transaction) since they do not seem to be needed for H100i Pro
@@ -96,7 +106,7 @@ class CorsairAsetekProDriver(Modern690Lc):
         approach.
         """
         msg = self.device.read(_PRO_READ_ENDPOINT, length, _READ_TIMEOUT)
-        LOGGER.debug('received %s', ' '.join(format(i, '02x') for i in msg))
+        _LOGGER.debug('received %s', LazyHexRepr(msg))
         self.device.release()
         return msg
 
@@ -114,38 +124,36 @@ class CorsairAsetekProDriver(Modern690Lc):
         colorCount = 0
         setColors = []
         for color in colors:
-            LOGGER.debug('color %s', color)
             colorCount = colorCount + 1
             setColors = setColors + color
-        self._write([0x56, colorCount] + setColors)
+        self._write([_PRO_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
         self._end_transaction_and_read(3)
 
     def _get_fan_speeds(self):
-        """Read the RPM speed of the fans. This Tries to get the speeds for 3 fans as that is
-        how many the H150i PRO has"""
+        """Read the RPM speed of the fans. Try to get the speeds for 3 fans as that is
+        how many the H150i PRO supports"""
         speeds = []
         for i in [0, 1, 2]:
             self._write([_PRO_CMD_READ_FAN_SPEED, i])
             msg = self._end_transaction_and_read(6)
             if msg[0] != 0x41 or msg[1] != 0x12 or msg[2] != 0x34 or msg[3] != i:
-                LOGGER.debug('Unable to get speed for fan id {}'.format(i))
+                _LOGGER.debug('Unable to get speed for fan id %d', i)
                 continue
-            speeds.append((msg[4] << 8) + msg[5])
+            speeds.append((i, (msg[4] << 8) + msg[5]))
         return speeds
 
     def _get_fan_indexes(self, channel):
         if len(channel) > 3:
             return [int(channel[3:]) - 1]
-        return range(len(self._get_fan_speeds()))
+        return [reading[0] for reading in self._get_fan_speeds()]
 
     def initialize(self, pump_mode=None, **kwargs):
         """Initialize the device."""
-        LOGGER.debug('Pro configure device...')
+        _LOGGER.debug('Pro configure device...')
         if pump_mode == None:
             return
         else:
-            LOGGER.debug('pump_mode %s write value %s', pump_mode,
-                format(_PRO_PUMP_MODES.index(pump_mode),'02x'))
+            _LOGGER.debug('pump_mode %s', pump_mode)
             self._write([_PRO_CMD_WRITE_PUMP_MODE, _PRO_PUMP_MODES.index(pump_mode)])
             self._end_transaction_and_read(5)
 
@@ -157,27 +165,22 @@ class CorsairAsetekProDriver(Modern690Lc):
         ### Can some Pro devices have more than one temp sensor?   OCL seems to code for that?
         self._write([_PRO_CMD_READ_AIO_TEMP])
         msg = self._end_transaction_and_read(6)
-        #LOGGER.debug('aio temp msg %s', ' '.join(format(i, '02x') for i in msg))
         aio_temp = msg[3] + msg[4]/10
         status.append(('Liquid temperature', aio_temp, '°C'))
         speeds = self._get_fan_speeds()
-        for i in range(len(speeds)):
-            status.append(('Fan {} speed'.format(i+1), speeds[i] | 0, 'rpm'))
+        for reading in speeds:
+            status.append(('Fan {} speed'.format(reading[0]), reading[1] | 0, 'rpm'))
         self._write([_PRO_CMD_READ_PUMP_MODE])
         msg = self._end_transaction_and_read(4)
-        #LOGGER.debug('pump mode msg %s', ' '.join(format(i, '02x') for i in msg))
         pump_mode = msg[3]
         status.append(('Pump mode', _PRO_PUMP_MODES[pump_mode], ""))
         self._write([_PRO_CMD_READ_PUMP_SPEED])
         msg = self._end_transaction_and_read(5)
-        #LOGGER.debug('pump speed msg %s', ' '.join(format(i, '02x') for i in msg))
         pump_speed = (msg[3] << 8) + msg[4]
         status.append(('Pump speed', pump_speed | 0, 'rpm'))
         self._write([_PRO_CMD_READ_FIRMWARE])
         msg = self._end_transaction_and_read(7)
-        #LOGGER.debug('firmware id msg %s', ' '.join(format(i, '02x') for i in msg))
-        firmware = '{}.{}.{}.{}'.format(format(msg[3], '01x'), format(msg[4], '01x'),
-            format(msg[5], '01x'),format(msg[6], '01x'))
+        firmware = '{}.{}.{}.{}'.format(*tuple(msg[3:7]))
         status.append(('Firmware version', firmware, ''))
         return status
 
@@ -185,44 +188,25 @@ class CorsairAsetekProDriver(Modern690Lc):
                   alert_threshold=_HIGH_TEMPERATURE, alert_color=[255, 0, 0],
                   speed=3, temps=['35', '45', '55'], **kwargs):
         """Set the color mode for a specific channel."""
-        LOGGER.debug('kwargs %s', kwargs)
+        if mode not in _PRO_COLOR_CHANGE_MODES:
+            raise KeyError('Unknown lighting mode {}'.format(mode))
         colors = list(colors)
-        LOGGER.debug('color count = %d', len(colors))
-        LOGGER.debug('speed %s', int(speed))
-        if mode == 'rainbow':
-            self._write_color_change_speed(mode, speed)
-            self._write([0x55, 0x01])
-            self._end_transaction_and_read(3)
-        elif mode == 'shift':
-            self._write_color(colors)
-            self._write_color_change_speed(mode, speed)
-            self._write([0x55, 0x01])
-            self._end_transaction_and_read(3)
-        elif mode == 'pulse':
-            self._write_color(colors)
-            self._write_color_change_speed(mode, speed)
-            self._write([0x52, 0x01])
-            self._end_transaction_and_read(3)
-        elif mode == 'blinking':
-            self._write_color(colors)
-            self._write_color_change_speed(mode, speed)
-            self._write([0x58, 0x01])
-            self._end_transaction_and_read(3)
-        elif mode == 'fixed':
-            self._write_color([colors[0], colors[0]])
-            self._write([0x55, 0x01])
-            self._end_transaction_and_read(3)
-        elif mode == 'alert':
-            temps = list(temps.split(' '))
-            LOGGER.debug('temps count = %d', len(temps))
-            LOGGER.debug('asetek temps %s %s %s', temps[0], temps[1], temps[2])
-            LOGGER.debug('color count = %d', len(colors))
+        _LOGGER.debug('color count = %d', len(colors))
+        _LOGGER.debug('color change speed %d', int(speed))
+        if mode == 'alert':
             self._write([0x5f, int(temps[0]), 0x00, int(temps[1]), 0x00, int(temps[2]), 0x00] + colors[0] + colors[1] + colors[2])
             self._end_transaction_and_read(6)
             self._write([0x5E, 0x01])
             self._end_transaction_and_read(3)
-        else:
-            raise KeyError('Unknown lighting mode {}'.format(mode))
+            return
+        if mode == 'fixed':
+            colors = [colors[0], colors[0]]
+        if mode != 'rainbow':
+            self._write_color(colors)
+        if mode != 'fixed':
+            self._write_color_change_speed(mode, speed)
+        self._write(_PRO_COLOR_CHANGE_MODES[mode])
+        self._end_transaction_and_read(3)
 
     def set_speed_profile(self, channel, profile, **kwargs):
         """Set channel to follow a speed duty profile."""
@@ -230,13 +214,14 @@ class CorsairAsetekProDriver(Modern690Lc):
             raise ValueError('set_speed_profile not implemented for the pump on CorsairAsetekPro')
         adjusted = self._prepare_profile(profile, 0, 100, _MAX_PROFILE_POINTS)
         for temp, duty in adjusted:
-            LOGGER.info('setting %s PWM point: (%i°C, %i%%), device interpolated',
+            _LOGGER.info('setting %s PWM point: (%i°C, %i%%), device interpolated',
                         channel, temp, duty)
         temps, duties = map(list, zip(*adjusted))
         # Need to write curve for each fan
         # Assume if we read n values from _get_fan_speed the fans are numbered [0, n)
         # This will have problems on a 2 fan system when the first fan has failed
         for i in self._get_fan_indexes(channel):
+            _LOGGER.info('setting speed for fan %d', i+1)
             self._write([_PRO_CMD_WRITE_FAN_SPEED, i] + temps + duties)
             self._end_transaction_and_read(32)
 
@@ -246,7 +231,7 @@ class CorsairAsetekProDriver(Modern690Lc):
             # Setting a fix speed is supported, however it is done by sending RPM or PWM
             # values. I'm not 100% sure how this maps to %s used by liquidclt so set a flat
             # speed profile instead.
-            LOGGER.info('using a flat profile to set %s to a fixed duty', channel)
+            _LOGGER.info('using a flat profile to set %s to a fixed duty', channel)
             self.set_speed_profile(channel, [(0, duty), (_CRITICAL_TEMPERATURE - 1, duty)])
             return
         # Fixed speed pump is not supported, so guess a curve from the duty value given
--- a/liquidctl/driver/asetek_pro.py
+++ b/liquidctl/driver/asetek_pro.py
@@ -78,7 +78,7 @@ _COLOR_SPEEDS_VALUES = {
 _COLOR_CHANGE_MODES = {
     'rainbow': [0x55, 0x01],
     'alert': [],
-    'shitf': [0x55, 0x01],
+    'shift': [0x55, 0x01],
     'pulse': [0x52, 0x01],
     'blinking': [0x58, 0x01],
     'fixed': [0x55, 0x01]
@@ -140,7 +140,7 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
                 _LOGGER.debug('Unable to get speed for fan id %d', i)
                 speeds.append(-1)
                 continue
-            speeds.append((i, (msg[4] << 8) + msg[5]))
+            speeds.append((msg[4] << 8) + msg[5])
         return speeds
 
     def _get_fan_indexes(self, channel):
@@ -155,6 +155,10 @@ class CorsairAsetekProDriver(_CommonAsetekDriver):
         """Initialize the device."""
         _LOGGER.debug('Pro configure device...')
         _LOGGER.debug('pump_mode %s', pump_mode)
+        pump_mode = pump_mode.lower()
+        if pump_mode not in _PUMP_MODES:
+            valid = ", ".join(_PUMP_MODES)
+            raise KeyError(f'Unknown pump mode {pump_mode}, should be one of {valid}')
         self._write([_CMD_WRITE_PUMP_MODE, _PUMP_MODES.index(pump_mode.lower())])
         self._end_transaction_and_read(5)
 
--- a/liquidctl/driver/asetek_pro.py
+++ b/liquidctl/driver/asetek_pro.py
@@ -12,11 +12,12 @@ and Chris Griffith.
 SPDX-License-Identifier: GPL-3.0-or-later
 """
 
+import itertools
 import logging
 
 import usb
 
-from liquidctl.driver.asetek import Modern690Lc
+from liquidctl.driver.asetek import _CommonAsetekDriver
 from liquidctl.driver.usb import UsbDriver
 from liquidctl.error import NotSupportedByDevice
 from liquidctl.keyval import RuntimeStorage
@@ -37,15 +38,16 @@ _PRO_READ_ENDPOINT = 0x81
 _PRO_CMD_READ_FIRMWARE = 0xaa
 _PRO_CMD_READ_AIO_TEMP = 0xa9
 _PRO_CMD_READ_FAN_SPEED = 0x41
-_PRO_CMD_WRITE_FAN_SPEED = 0x40
+_PRO_CMD_WRITE_FAN_CURVE = 0x40
+_PRO_CMD_WRITE_FAN_SPEED = 0x42
 _PRO_CMD_READ_PUMP_SPEED = 0x31
 _PRO_CMD_READ_PUMP_MODE = 0x33
 _PRO_CMD_WRITE_PUMP_MODE = 0x32
 _PRO_CMD_WRITE_COLOR_LIST = 0x56
 _PRO_PUMP_MODES = [
-    'Quiet',
-    'Balanced',
-    'Performance'
+    'quiet',
+    'balanced',
+    'performance'
     ]
 
 _PRO_SPEEDS = {
@@ -81,7 +83,7 @@ _PRO_COLOR_CHANGE_MODES = {
 }
 
 
-class CorsairAsetekProDriver(Modern690Lc):
+class CorsairAsetekProDriver(_CommonAsetekDriver):
     """liquidctl driver for Corsair-branded sixth generation Asetek coolers."""
     SUPPORTED_DEVICES = [
         (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {}),
@@ -112,20 +114,15 @@ class CorsairAsetekProDriver(Modern690Lc):
 
     def _write_color_change_speed(self, mode, speed):
         """Send the speed to cycle colors on the RGB pump"""
-        if speed < 1 or speed > 3:
-            raise ValueError('Unsupported speed ({}), only 1-3 supported'.format(speed))
-
+        speed = clamp(speed, 1, 3)
         self._write([0x53, _PRO_SPEEDS[mode][int(speed)-1]])
         self._end_transaction_and_read(3)
 
     def _write_color(self, colors):
         """Write a list of colors to cycle between, the cycle pattern comes from the
         value used for cycle speed"""
-        colorCount = 0
-        setColors = []
-        for color in colors:
-            colorCount = colorCount + 1
-            setColors = setColors + color
+        setColors = list(itertools.chain(*colors))
+        colorCount = int(len(setColors) / 3)
         self._write([_PRO_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
         self._end_transaction_and_read(3)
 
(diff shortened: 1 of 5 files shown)


tests/test_asetek_pro.py
```diff
@@ -0,0 +1,66 @@
+import unittest
```
MarshallAsch, line 1: Can you add more tests to confirm that the messages being sent to the device are actually the ones that are expected so that if there is a refactor we can ensure that the protocol implementation has not changed accidentally.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+import usb
+
+from liquidctl.driver.asetek import Modern690Lc
+from liquidctl.driver.usb import UsbDriver
+from liquidctl.error import NotSupportedByDevice
+from liquidctl.keyval import RuntimeStorage
+from liquidctl.util import clamp
+
+
+LOGGER = logging.getLogger(__name__)
```
MarshallAsch, line 26: Can this logger be renamed to `_LOGGER` to conform with the projects [style guide](docs/developer/style-guide.md)

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+_PRO_CMD_WRITE_PUMP_MODE = 0x32
+_PRO_PUMP_MODES = [
+    'Quiet',
+    'Balanced',
+    'Performance'
+    ]
+
+_PRO_SPEEDS = {
+    'rainbow': [
+        0x30, # slow
```
MarshallAsch, line 52: Are other speed values allowed or are these specific values the only ones allowed?

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+        (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {}),
+        (0x1b1c, 0x0c13, None, 'Corsair Hydro H115i Pro', {}),
+        (0x1b1c, 0x0c14, None, 'Corsair Hydro H110i Pro', {}),
+        (0x1b1c, 0x0c15, None, 'Corsair Hydro H100i Pro', {}),
+        (0x1b1c, 0x0c16, None, 'Corsair Hydro H80i Pro', {}),
+    ]
+
+    def _write(self, data):
+        """Write data to the AIO and log"""
+        LOGGER.debug('write %s', ' '.join(format(i, '02x') for i in data))
```
MarshallAsch, line 86: See lazy hex strings in the style guide.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+        self.device.release()
+        return msg
+
+    def _write_color_change_speed(self, mode, speed):
+        """Send the speed to cycle colors on the RGB pump"""
+        if speed < 1 or speed > 3:
+            raise ValueError('Unsupported speed ({}), only 1-3 supported'.format(speed))
+
+        self._write([0x53, _PRO_SPEEDS[mode][int(speed)-1]])
+        self._end_transaction_and_read(3)
```
MarshallAsch, line 109: If this is unused code can it be removed?

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def _write_color(self, colors):
+        """Write a list of colors to cycle between, the cycle pattern comes from the
+        value used for cycle speed"""
+        colorCount = 0
+        setColors = []
+        for color in colors:
+            LOGGER.debug('color %s', color)
+            colorCount = colorCount + 1
+            setColors = setColors + color
```
MarshallAsch, line 119: Is this just a loop to move the colors from one list to another? Can it be replaced with:

`setColors = colors` and `colorCount = len(colors)`

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+        self._write([0x56, colorCount] + setColors)
```
MarshallAsch, line 120: `0x56` magic number? what does it mean, can a defined constant be used instead?

liquidctl/driver/asetek_pro.py
MarshallAsch: This function does seem to be called, maybe the comment is just out of date?

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+        status.append(('Pump mode', _PRO_PUMP_MODES[pump_mode], ""))
+        self._write([_PRO_CMD_READ_PUMP_SPEED])
+        msg = self._end_transaction_and_read(5)
+        #LOGGER.debug('pump speed msg %s', ' '.join(format(i, '02x') for i in msg))
+        pump_speed = (msg[3] << 8) + msg[4]
+        status.append(('Pump speed', pump_speed | 0, 'rpm'))
+        self._write([_PRO_CMD_READ_FIRMWARE])
+        msg = self._end_transaction_and_read(7)
+        #LOGGER.debug('firmware id msg %s', ' '.join(format(i, '02x') for i in msg))
+        firmware = '{}.{}.{}.{}'.format(format(msg[3], '01x'), format(msg[4], '01x'),
```
MarshallAsch, line 179: Are you sure there are 4 sections to this firmware version? There are normally just 3 parts `x.y.z` in a firmware version.

MarshallAsch, line 179: see for an example of a potentially cleaner way of writing the firmware version string from one of the other drivers:

https://github.com/jonasmalacofilho/liquidctl/blob/90d7ec50d7235c58b6ee0d16e602763ed2a1bb9a/liquidctl/driver/hydro_platinum.py#L175-L178

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+            temps = list(temps.split(' '))
+            LOGGER.debug('temps count = %d', len(temps))
+            LOGGER.debug('asetek temps %s %s %s', temps[0], temps[1], temps[2])
+            LOGGER.debug('color count = %d', len(colors))
+            self._write([0x5f, int(temps[0]), 0x00, int(temps[1]), 0x00, int(temps[2]), 0x00] + colors[0] + colors[1] + colors[2])
+            self._end_transaction_and_read(6)
+            self._write([0x5E, 0x01])
+            self._end_transaction_and_read(3)
+        else:
+            raise KeyError('Unknown lighting mode {}'.format(mode))
```
MarshallAsch, line 225: Can this be refactored to have less duplicate code?

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,271 @@
...
+            duty = 0
+        elif duty < 80:
+            duty = 1
+        else:
+            duty = 2
+        self._write([_PRO_CMD_WRITE_PUMP_MODE, duty])
+        self._end_transaction_and_read(5)
+        # OCL Reads the mode after setting it, so do the same
+        self._write([_PRO_CMD_READ_PUMP_MODE])
+        self._end_transaction_and_read(4)
```
MarshallAsch, line 265: I think this is dead code? it is after a return statement

MarshallAsch (changes requested):
This is  good first PR, there are some changes that need to be made before it can get added. 

There are a bunch of comments, but  they are mostly style changes, or when some logging statements get printed. It looks like there is some testing code and debug statements that have been left in. Some of the log statements could probably be removed or be converted to `warning` or `debug` messages.

liquidctl/driver/asetek_pro.py
joaqim: That return statement is only met if channel is 'fan1', 'fan2', etc.
It's a if condition for 'fan' or 'pump' mode when changing speeds.

MarshallAsch: My bad, the side by side diff made  it look like the `return` was not indented.

tests/test_asetek_pro.py
andytank123: I've added some more unit test

liquidctl/driver/asetek_pro.py
andytank123: Done

liquidctl/driver/asetek_pro.py
andytank123: The values in this data structure are the only ones I believe are allowed.

liquidctl/driver/asetek_pro.py
andytank123: Done

liquidctl/driver/asetek_pro.py
andytank123: This code should be called from the `set_color` function.

liquidctl/driver/asetek_pro.py
andytank123: This code is removing the nesting from a list of lists and turning it into a single non nested list
E.g. `[[1, 2, 3],  [4, 5, 6]]` becomes `[1, 2, 3, 4, 5, 6]`

liquidctl/driver/asetek_pro.py
andytank123: I've moved this to a constant

liquidctl/driver/asetek_pro.py
andytank123: I think Asetek firmware has 4 parts. This matches both the OCL driver and the existing Asetek (non pro) driver.

andytank123: Done

liquidctl/driver/asetek_pro.py
andytank123: Done

MarshallAsch (commented):
Looks a lot  cleaner now. good job. Ill look at the tests and ill look at it in some more detail a little later.

MarshallAsch:
@andytank123 is there a list of other tasks that need to be done on your side before this is ready to be merged or is this version right now a done and your confident that it works properly?

andytank123:
I'm happy for this to be merged in it's current state, and am confident that it replicates the behavior of the OCL driver. However unfortunately I don't currently have access to a Corsair PRO device to do actual testing. I've ordered one as part of a new PC, but until It's delivered and working, I can't run live testing on the change. @joaqim has done some testing this to make sure this works.

cluxter:
I own a `Corsair Hydro Series H150i PRO RGB` hardware.
I just tested this PR.
Here are the results:
```
# liquidctl list --verbose
Device ID 0: Corsair Hydro H150i Pro
├── Vendor ID: 0x1b1c
├── Product ID: 0x0c12
├── Release number: 0x0100
├── Serial number: 7289_2.0
├── Bus: usb1
├── Address: 4
├── Port: 5.2
└── Driver: CorsairAsetekProDriver
```
--> Device is properly recognized: OK.

```
# liquidctl initialize
```
--> No error reported on init: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       31.7  °C
├── Fan 0 speed               550  rpm
├── Fan 1 speed               529  rpm
├── Fan 2 speed               526  rpm
├── Pump mode               Quiet  
├── Pump speed               1080  rpm
└── Firmware version      1.0.4.0
```
--> Status looks good and consistent, including firmware version: OK.

After 1 minute:
```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       32.0  °C
├── Fan 0 speed               567  rpm
├── Fan 1 speed               557  rpm
├── Fan 2 speed               549  rpm
├── Pump mode               Quiet  
├── Pump speed               1050  rpm
└── Firmware version      1.0.4.0
```
--> Status still looking good and consistent: OK.

```
# liquidctl set pump speed 50
```
--> No error reported: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature        31.0  °C
├── Fan 0 speed                521  rpm
├── Fan 1 speed                516  rpm
├── Fan 2 speed                512  rpm
├── Pump mode             Balanced  
├── Pump speed                2130  rpm
└── Firmware version       1.0.4.0
```
--> Pump speed changed and Pump mode adjusted accordingly automatically: OK.

```
# liquidctl set pump speed 100
```
--> No error reported and I can now really hear the pump working: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature           31.0  °C
├── Fan 0 speed                   511  rpm
├── Fan 1 speed                   503  rpm
├── Fan 2 speed                   494  rpm
├── Pump mode             Performance  
├── Pump speed                   2910  rpm
└── Firmware version          1.0.4.0
```
---> Status looking good and consistent, Pump speed value is now higher at what seems to be its maximum value and Pump mode adjusted accordingly automatically: OK.

```
# liquidctl set pump speed 0
```
--> Testing danger here... No error reported, PC is not on fire yet: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       31.9  °C
├── Fan 0 speed               557  rpm
├── Fan 1 speed               547  rpm
├── Fan 2 speed               540  rpm
├── Pump mode               Quiet  
├── Pump speed               1050  rpm
└── Firmware version      1.0.4.0
```
--> Pump speed was set to a minimal value of 1050 RPM and not 0 RPM, which is good, otherwise my CPU would probably be dead by now. Pump mode changed accordingly: OK.

```
# liquidctl set fan0 speed 100
ERROR: Unexpected error with Corsair Hydro H150i Pro
OverflowError: unsigned byte integer is less than minimum
```
--> `fan0` doesn't exist here: ERROR.

```
# liquidctl set fan1 speed 100
```
--> No error reported here.

After 1 min:
```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       31.0  °C
├── Fan 0 speed              1680  rpm
├── Fan 1 speed               402  rpm
├── Fan 2 speed               402  rpm
├── Pump mode               Quiet  
├── Pump speed               1050  rpm
└── Firmware version      1.0.4.0
```
--> There is a mismatch between what is displayed as `Fan 0` and the number we use is the argument: NOT OK, SHOULD BE FIXED.

Also, please note that this change takes about 1~2 seconds before it's taken into account, and it takes a long time for the fan to ramp up to 100%. This is something that also happened when I used another fan driver (possibly `OpenCorsairLink`, but maybe another similar project) and used the RPM measurement to set the fan speed. It was also possible to set the fan speed using PWM measurement, and in that case the fan speed was way more stable and ramped up way faster to the desired speed. So it looks like `liquidctl` is using RPM and not PWM, because the fan speed takes a while before reaching the desired value and several `liquidctl status` commands show a pretty variable fan speed in a matter of seconds. **Can someone confirm this?**

```
# liquidctl set fan speed 100
```
--> No error reported: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       30.0  °C
├── Fan 0 speed              1614  rpm
├── Fan 1 speed              1620  rpm
├── Fan 2 speed              1616  rpm
├── Pump mode               Quiet  
├── Pump speed               1020  rpm
└── Firmware version      1.0.4.0
```
--> All fans were set to the same speed: OK.

```
# liquidctl set fan speed 0
```
--> No error reported: OK.

```
# liquidctl status
Corsair Hydro H150i Pro
├── Liquid temperature       29.5  °C
├── Fan 0 speed               0  rpm
├── Fan 1 speed               0  rpm
├── Fan 2 speed               0  rpm
├── Pump mode               Quiet  
├── Pump speed               1020  rpm
└── Firmware version      1.0.4.0
```
--> All fans were turned off: OK (but again: it takes about 1 min to reach this level, which is not optimal vs. what you could achieve with `OpenCorsairLink`).

I wasn't able to set any color or set any Pump mode manually --> NOT OK, TODO.

cluxter:
Here is the list of the commands that I used above in one script so you can easily reproduce my tests (run it as `root`):
```
#!/bin/bash
# liquidctl list --verbose
# liquidctl initialize
# liquidctl status
# sleep 60
# liquidctl status
# liquidctl set pump speed 50
# sleep 60
# liquidctl status
# liquidctl set pump speed 100
# sleep 60
# liquidctl status
# liquidctl set pump speed 0
# sleep 60
# liquidctl status
# liquidctl set fan0 speed 100
# sleep 60
# liquidctl status
# liquidctl set fan1 speed 100
# sleep 60
# liquidctl status
# liquidctl set fan speed 100
# sleep 60
# liquidctl status
# sleep 60
# liquidctl set fan speed 0
# sleep 60
# liquidctl status
```

andytank123:
@cluxter Thank you for your help testing this. A few things to note from your testing.

1. It is not possible to set the pump to run at a constant speed, it only accepts pump mode commands, so I've mapped % values into different modes. I'll add the ability to set the pump mode on the command line as well. 
2. The status output for fans should start at 1 not 0, so it matches the fan set command. I'll fix this in the code.
3. I'll look into changing the way the driver sets a constant fan speed. It is currently always using the 'set fan speed on a temperature curve' feature. Just a flat curve when using a constant temperature. If using set PWM feature gives better results, I'll move to using that.
4. Can you please give me some examples of the color set commands your using and the output returned by liquidctl as well as what it looks like on the pump.

jonasmalacofilho:
@andytank123 

First, thank you for working on this!

> It is not possible to set the pump to run at a constant speed, it only accepts pump mode commands, so I've mapped % values into different modes. I'll add the ability to set the pump mode on the command line as well.

In that case it may be preferable to use the `---pump-mode` option, like we do in the [Hydro Platinum driver](https://github.com/jonasmalacofilho/liquidctl/blob/45a89de753ec17d756f1cf82cbdb75bb9666bb64/liquidctl/driver/hydro_platinum.py#L156-L178).  This is sort of a hack (see #235), but is probably better than mapping percentages to modes.

> The status output for fans should start at 1 not 0, so it matches the fan set command. I'll fix this in the code.

Actually, I would prefer if you changed things the other way around.  In the other drivers we're already counting from 1 in user-facing strings.

> I'll look into changing the way the driver sets a constant fan speed. It is currently always using the 'set fan speed on a temperature curve' feature. Just a flat curve when using a constant temperature. If using set PWM feature gives better results, I'll move to using that.

I think there are two different things to consider here.

First, it may indeed be possible to optimize how constant speeds are sent.  But you should also know that some older Asetek coolers have firmware quirks (well, bugs really) and switching between fixed speeds and profiles, or setting different channels to different modes of operation, can result in unpredictable behavior.  This could have improved, but will need to be specifically tested for.

Second, and if I correctly undestood @cluxter, it may be possible to configure the fan speed at the protocol level in two different ways: duty (percentage, PWM) and rpm (perhaps the firmware implements a PID controller, or some rudimentary alternative), and the first method may react faster.  Is this something the driver could take advantage of?

_(Please excuse me if this last part didn't make any sense, I still need to take a fresh look at your code, as well as the corresponding implementation in OCL)._

jonasmalacofilho:
@cluxter 

Thank you for the thorough testing and detailed reporting!

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def _write_color(self, colors):
+        """Write a list of colors to cycle between, the cycle pattern comes from the
+        value used for cycle speed"""
+        colorCount = 0
+        setColors = []
+        for color in colors:
+            colorCount = colorCount + 1
+            setColors = setColors + color
+        self._write([_PRO_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
```
jonasmalacofilho, line 129: Assuming `colors` is indeed a list, this entire chunk should be equivalent to:

```
self._write([_PRO_CMD_WRITE_COLOR_LIST, len(colors)] + colors)
```

That said, this would try to write a list of lists, which is probably not the intended behavior (and could explain why this part of the driver isn't working).  Instead it may be necessary to use `itertools.chain`.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+            _LOGGER.debug('pump_mode %s', pump_mode)
+            self._write([_PRO_CMD_WRITE_PUMP_MODE, _PRO_PUMP_MODES.index(pump_mode)])
+            self._end_transaction_and_read(5)
+
+    def get_status(self, **kwargs):
+        """Get a status report.
+        Returns a list of `(property, value, unit)` tuples.
+        """
+        status = []
+        ### Can some Pro devices have more than one temp sensor?   OCL seems to code for that?
```
jonasmalacofilho, line 165: OCL has a generic device model; as far as I know all of these coolers has exactly one sensor.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+            colorCount = colorCount + 1
+            setColors = setColors + color
+        self._write([_PRO_CMD_WRITE_COLOR_LIST, colorCount] + setColors)
+        self._end_transaction_and_read(3)
+
+    def _get_fan_speeds(self):
+        """Read the RPM speed of the fans. Try to get the speeds for 3 fans as that is
+        how many the H150i PRO supports"""
+        speeds = []
+        for i in [0, 1, 2]:
```
jonasmalacofilho, line 136: We should already know (since the time the driver was instanced) the maximum number of fans the specific device we're bound to is able to monitor and control, based on its model.  We should use that value here, instead of a constant.

Also, in Python it's more idiomatic to write:

```
for i in range(3):
```

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def _get_fan_indexes(self, channel):
+        if len(channel) > 3:
+            return [int(channel[3:]) - 1]
+        return [reading[0] for reading in self._get_fan_speeds()]
+
+    def initialize(self, pump_mode=None, **kwargs):
+        """Initialize the device."""
+        _LOGGER.debug('Pro configure device...')
+        if pump_mode == None:
```
jonasmalacofilho, line 153: The cooler should always be in a know state after `initialize`.  We need a default value here, and I recommend `'balanced'`, to homogeneity with the Hydro Platinum driver.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+            return
+        else:
+            _LOGGER.debug('pump_mode %s', pump_mode)
+            self._write([_PRO_CMD_WRITE_PUMP_MODE, _PRO_PUMP_MODES.index(pump_mode)])
```
jonasmalacofilho, line 157: The implicit comparison for `pump_mode` must be case insensitive.

I would also prefer to see all lowercase values in `_PRO_PUMP_MODES`.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        status = []
+        ### Can some Pro devices have more than one temp sensor?   OCL seems to code for that?
+        self._write([_PRO_CMD_READ_AIO_TEMP])
+        msg = self._end_transaction_and_read(6)
+        aio_temp = msg[3] + msg[4]/10
+        status.append(('Liquid temperature', aio_temp, '°C'))
+        speeds = self._get_fan_speeds()
+        for reading in speeds:
+            status.append(('Fan {} speed'.format(reading[0]), reading[1] | 0, 'rpm'))
```
jonasmalacofilho, line 172: I see you're expecting `reading[1]` to be `None` sometimes.  Why would that happen?

By the way, you can use `f` strings now.  It's possible that the driver you based this on is from when liquidctl supported Python < 3.6.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        self._write([_PRO_CMD_READ_PUMP_MODE])
+        msg = self._end_transaction_and_read(4)
+        pump_mode = msg[3]
+        status.append(('Pump mode', _PRO_PUMP_MODES[pump_mode], ""))
+        self._write([_PRO_CMD_READ_PUMP_SPEED])
+        msg = self._end_transaction_and_read(5)
+        pump_speed = (msg[3] << 8) + msg[4]
+        status.append(('Pump speed', pump_speed | 0, 'rpm'))
```
jonasmalacofilho, line 180: Same question as before regarding `pump_speed | 0`.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        self._write([_PRO_CMD_READ_FIRMWARE])
+        msg = self._end_transaction_and_read(7)
+        firmware = '{}.{}.{}.{}'.format(*tuple(msg[3:7]))
+        status.append(('Firmware version', firmware, ''))
+        return status
```
jonasmalacofilho, line 185: This function could probably be easier (or at least more pleasant) to follow if you computed the variables in one section (maybe in small blocks, or using a higher level abstraction) and assembled the result in another.

Regarding the abstraction, I'm seeing a common pattern throughout the driver of write a command, perhaps with an argument, and read a response with a certain number of bytes.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+_LOGGER = logging.getLogger(__name__)
+
+_MAX_PROFILE_POINTS = 7
+_CRITICAL_TEMPERATURE = 60
+_HIGH_TEMPERATURE = 45
+_READ_LENGTH = 32
+_READ_TIMEOUT = 2000
+_WRITE_TIMEOUT = 2000
+
+_PRO_WRITE_ENDPOINT = 0x1
```
jonasmalacofilho, line 35: Private constants don't need a prefix based on this module's name; `_WRITE_ENDPOINT` saves a few columns and keystrokes, and is just as clear.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        msg = self._end_transaction_and_read(5)
+        pump_speed = (msg[3] << 8) + msg[4]
+        status.append(('Pump speed', pump_speed | 0, 'rpm'))
+        self._write([_PRO_CMD_READ_FIRMWARE])
+        msg = self._end_transaction_and_read(7)
+        firmware = '{}.{}.{}.{}'.format(*tuple(msg[3:7]))
+        status.append(('Firmware version', firmware, ''))
+        return status
+
+    def set_color(self, channel, mode, colors, time_per_color=1, time_off=None,
```
jonasmalacofilho, line 187: I'm going to skip this method for now.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        else:
+            duty = 2
+        self._write([_PRO_CMD_WRITE_PUMP_MODE, duty])
+        self._end_transaction_and_read(5)
+        # OCL Reads the mode after setting it, so do the same
+        self._write([_PRO_CMD_READ_PUMP_MODE])
+        self._end_transaction_and_read(4)
+
+    @classmethod
+    def probe(cls, handle, legacy_690lc=False, **kwargs):
```
jonasmalacofilho, line 253: This is only necessary because this class extends `Modern690Lc`.

But as far as I can see we only use  `_prepare_profile` from the super class, so we can instead inherit from `_CommonAsetekDriver`, which would make it clearer that this *isn't* a Modern 690Lc, as well as allow us to drop this `probe` override.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        # This will have problems on a 2 fan system when the first fan has failed
+        for i in self._get_fan_indexes(channel):
+            _LOGGER.info('setting speed for fan %d', i+1)
+            self._write([_PRO_CMD_WRITE_FAN_SPEED, i] + temps + duties)
+            self._end_transaction_and_read(32)
+
+    def set_fixed_speed(self, channel, duty, **kwargs):
+        """Set channel to a fixed speed duty."""
+        if channel.startswith('fan'):
+            # Setting a fix speed is supported, however it is done by sending RPM or PWM
```
jonasmalacofilho, line 231: PWM/duty makes more sense, since we don't know which fans are connected and thus cannot translate an input duty (=percentage) to a specific speed (=rpm).

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        # Arbitrary values are arbitrary.
+        if duty < 50:
+            duty = 0
+        elif duty < 80:
+            duty = 1
+        else:
+            duty = 2
+        self._write([_PRO_CMD_WRITE_PUMP_MODE, duty])
+        self._end_transaction_and_read(5)
+        # OCL Reads the mode after setting it, so do the same
```
jonasmalacofilho, line 248: From the rest of the driver it seems that this device is a lot more sensible than the previous generation.  Doing these extra "reads" (by which I mean a command + a read) is probably no longer necessary.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+    def set_fixed_speed(self, channel, duty, **kwargs):
+        """Set channel to a fixed speed duty."""
+        if channel.startswith('fan'):
+            # Setting a fix speed is supported, however it is done by sending RPM or PWM
+            # values. I'm not 100% sure how this maps to %s used by liquidclt so set a flat
+            # speed profile instead.
+            _LOGGER.info('using a flat profile to set %s to a fixed duty', channel)
+            self.set_speed_profile(channel, [(0, duty), (_CRITICAL_TEMPERATURE - 1, duty)])
+            return
+        # Fixed speed pump is not supported, so guess a curve from the duty value given
```
jonasmalacofilho, line 237: If the hardware only supports pump modes, then that's all we should support too, through `initialize --pump-mode <mode>`.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+            self._write_color(colors)
+        if mode != 'fixed':
+            self._write_color_change_speed(mode, speed)
+        self._write(_PRO_COLOR_CHANGE_MODES[mode])
+        self._end_transaction_and_read(3)
+
+    def set_speed_profile(self, channel, profile, **kwargs):
+        """Set channel to follow a speed duty profile."""
+        if channel == 'pump':
+            raise ValueError('set_speed_profile not implemented for the pump on CorsairAsetekPro')
```
jonasmalacofilho, line 214: We have a specific expection class for this type of error:

```
from liquidctl.error import NotSupportedByDevice

raise NotSupportedByDevice()
```

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        adjusted = self._prepare_profile(profile, 0, 100, _MAX_PROFILE_POINTS)
```
jonasmalacofilho, line 215: The channel name should probably be validated before computing the profile.  Any comparisons should be case insensitive, but if the channel is neither `fan` nor a valid `fan[0-9]` value, you should raise a `ValueError`.

Note: not all liquidctl errors are specified yet, so for now and if convenient you can also let a `KeyError` or `IndexError` bubble up.  Nevertheless a customized `ValueError` message would be preferable.

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        for temp, duty in adjusted:
+            _LOGGER.info('setting %s PWM point: (%i°C, %i%%), device interpolated',
+                        channel, temp, duty)
+        temps, duties = map(list, zip(*adjusted))
+        # Need to write curve for each fan
+        # Assume if we read n values from _get_fan_speed the fans are numbered [0, n)
+        # This will have problems on a 2 fan system when the first fan has failed
```
jonasmalacofilho, line 222: Hm, this seems like something that needs to be handled.

tests/test_asetek_pro.py
jonasmalacofilho: It seems much better, thanks.

But I'm also going skip reviewing the tests for now.

liquidctl/driver/asetek_pro.py
jonasmalacofilho: Hm, I noticed this before: it would be better to take the `speed` parameter as a string (`'slow', 'normal', 'fast'`).

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+    'fixed': [0x55, 0x01]
+}
+
+
+class CorsairAsetekProDriver(Modern690Lc):
+    """liquidctl driver for Corsair-branded sixth generation Asetek coolers."""
+    SUPPORTED_DEVICES = [
+        (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {}),
+        (0x1b1c, 0x0c13, None, 'Corsair Hydro H115i Pro', {}),
+        (0x1b1c, 0x0c14, None, 'Corsair Hydro H110i Pro', {}),
```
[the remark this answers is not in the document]
andytank123: This and the `H80i Pro` come from the OCL driver. I'm happy to remove them if you think they don't exist.

liquidctl/driver/asetek_pro.py
andytank123: I've moved this to use `itertools.chain` as it's neater, but it's functional behavior hasn't changed.

liquidctl/driver/asetek_pro.py
andytank123: This was handled by my last change, out of date comment now removed.

liquidctl/driver/asetek_pro.py
andytank123: `initialize --pump-mode <mode` already works. I'll replace this with a `NotSupportedByDevice` error

liquidctl/driver/asetek_pro.py
andytank123: Updated

liquidctl/driver/asetek_pro.py
andytank123: I've used a `KeyError` with a custom message for now

liquidctl/driver/asetek_pro.py
andytank123: I think your right, as `--pump-mode` in the `initialize` function doesn't do this.  I've also removed this and am just setting pump-mode in the `initialize` function, due to another comment

liquidctl/driver/asetek_pro.py
andytank123: This now just uses the set fixed PWM feature. This should also help with the latency @cluxter was seeing.

liquidctl/driver/asetek_pro.py
andytank123: I've changed the class inherited from, should I remove the `_` from the start of `_commonAsetekDriver` as I'm inheriting from it?

liquidctl/driver/asetek_pro.py
andytank123: It should never be `None`, I've removed the ` | 0`

liquidctl/driver/asetek_pro.py
andytank123: It should never be `None`, I've removed the ` | 0`. I've also moved to `f` strings

liquidctl/driver/asetek_pro.py
andytank123: Done

liquidctl/driver/asetek_pro.py
andytank123: Done

liquidctl/driver/asetek_pro.py
andytank123: Sorry, I'm not sure how to get the exact device during when the drive is created. Is the device id passed into the `initialize` function?

liquidctl/driver/asetek_pro.py
andytank123: Removed this comment

liquidctl/driver/asetek_pro.py
jonasmalacofilho: That's better, but:
- `foo` raises `KeyError`
- `fanfoo` raises `ValueError` (from `int(...)`)

More importantly, and I missed this before, `fan4` (or `fan99`) will *not* raise, and result in unsound writes.

jonasmalacofilho: Regarding your question about making `_CommonAsetekDriver` public, I see your point: inheriting from a private class in another module is, err, wrong.

But I think the proper solution is to clean up, document, test and move `prepare_profile` to `liquidctl.util`, and stop inheriting from any `asetek` classes (instead inheriting from `liquidctl.usb.UsbDriver` directly).

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def set_fixed_speed(self, channel, duty, **kwargs):
+        """Set channel to a fixed speed duty."""
+        channel = channel.lower()
+        if channel.startswith('fan'):
+            duty = clamp(duty, 0, 100)
+            # Need to write curve for each fan in channel
+            for i in self._get_fan_indexes(channel):
+                _LOGGER.info('setting speed for fan %d to %d', i+1, duty)
+                self._write([_PRO_CMD_WRITE_FAN_SPEED, i, duty])
```
jonasmalacofilho, line 231: Agreed, this will probably help.

@cluxter, can you check that *all* fan channels are set to sensible speeds after each step of the following sequence?

```
set fan speed 100
set fan speed 0 0 50 100
set fan speed 100
set fan1 speed 0 0 50 100
set fan2 speed 20
```

liquidctl/driver/asetek_pro.py
MarshallAsch: I still find it suspicious that it only accepts the 3 speeds, @andytank123 did you try other speed values (related to one of my earlier comments). I know for one of my devices (I dont remember which one, it may been the `H100i Platinum` icue listed fast normal and slow, but I could actually send any value.

jonasmalacofilho: It would make sense if the parameters for each speed value are timings for parts of the animations.  Perhaps in tens of seconds?

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,256 @@
...
+        self.device.write(_PRO_WRITE_ENDPOINT, data, _WRITE_TIMEOUT)
+
+    # Not calling this (or begin transaction) since they do not seem to be needed for H100i Pro
+    def _end_transaction_and_read(self, length=_READ_LENGTH):
+        """End the transaction by reading from the device.
+        According to the official documentation, as well as Craig's open-source
+        implementation (libSiUSBXp), it should be necessary to check the queue
+        size and read data in chunks.  However, leviathan and its derivatives
+        seem to work fine without this complexity; we currently try the same
+        approach.
```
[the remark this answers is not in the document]
cluxter: In Windows 10, when I go into the Device Manager, there is "USB bus controller > Corsair USBXp Driver". In it, there is a "Driver" tab, with a "Driver details" button. When I click on it, there are:
```
C:\WINDOWS\System32\Drivers\SiLib.sys
C:\WINDOWS\System32\Drivers\SiUSBXp.sys
```
Note: `iCUE` is installed and running.
Let me know if there is another way to check the driver being used.

Also, here is the output of `lsusb -v -d 1b1c:0c12` from Linux:
```
$ lsusb -v -d 1b1c:0c12

Bus 001 Device 004: ID 1b1c:0c12 Corsair H150i Platinum
Couldn't open device, some information will be missing
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 
  bDeviceSubClass         0 
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x1b1c Corsair
  idProduct          0x0c12 
  bcdDevice            1.00
  iManufacturer           1 Corsair
  iProduct                2 H150i Platinum
  iSerial                 3 7289_2.0
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0020
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0xc0
      Self Powered
    MaxPower               64mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      0 
      bInterfaceProtocol      0 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
```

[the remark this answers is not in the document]
jonasmalacofilho: > In Windows 10, when I go into the Device Manager, there is "USB bus controller > Corsair USBXp Driver". In it, there is a "Driver" tab, with a "Driver details" button. When I click on it, there are:
> 
> ```
> C:\WINDOWS\System32\Drivers\SiLib.sys
> C:\WINDOWS\System32\Drivers\SiUSBXp.sys
> ```

This already tells me everything I needed to know about the driver on Windows, thanks!

liquidctl/driver/asetek_pro.py
cluxter: They all work very well!!
I also set this:
```
set fan speed 30 0 40 100
```
and ran a full kernel compilation. Fans got faster and faster up to their max when temperature went up to 40°C (and more, since the pump was in "quiet" mode), and slower and slower when I stopped the compilation and the temperature was declining, down to ~200 RPM when temperature reached 30.4°C again.

Also, now the fans start almost instantaneously! I tested going from speed 0 to speed 100, and it takes up to 3 seconds for the fans to be started, which is pretty good in my opinion (I have no idea if we could even improve this?). But once they are started, they reach their speed very very quickly, about 2 seconds if not less. It's awesome!
I noticed that they don't necessarily start at the same time, so maybe this could be improved too, even if this is good enough for the moment. As you can see, I like optimized stuff, so this is something I would enjoy working on once all the features are fully implemented (since I know Python).

I tried to set the pump mode on something other than "quiet" but didn't succeed. From what I see from the modifications, it looks intentional for the moment, so I assume you'll implement it later on.

I also tried:
```
set ring color fixed ff0000
set ring color fixed 00ff00
```
and it worked.

However:
```
set fan color fixed ff0000
```
or
```
set fan1 color fixed ff0000
```
didn't work, without any error though. I think it's my setup, I will check if I'm at least able to change the fan colors on Windows and let you know.

jonasmalacofilho:
@cluxter 

Thanks!

Regarding the pump modes, can you show us the output of:

```
liquidctl initialize --debug
liquidctl initialize --pump-mode quiet --debug
liquidctl initialize --pump-mode performance --debug
```

cluxter:
It actually works, I just used the wrong command before, my bad.
Here are the outputs:

```
# liquidctl initialize --debug
[DEBUG] liquidctl.cli: running liquidctl v1.4.2 (879144f6477c)
[DEBUG] liquidctl.driver.smbus: skipping LinuxI2c, smbus package not available
[DEBUG] liquidctl.driver.usb: searching HidapiBus (CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver)
[DEBUG] liquidctl.driver.usb: found HID device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found HID device 0b05:1872
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (Hydro690Lc, Legacy690Lc, Modern690Lc, _CommonAsetekDriver, CorsairAsetekProDriver)
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found USB device 04a9:221b
[DEBUG] liquidctl.driver.usb: found USB device 046a:0023
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 8087:0029
[DEBUG] liquidctl.driver.usb: found USB device 0b05:1872
[DEBUG] liquidctl.driver.usb: found USB device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:0c12
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: found USB device 05e3:0610
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.cli: device: Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: selected interface: 0
[DEBUG] liquidctl.driver.asetek: set clear to send = True
[DEBUG] liquidctl.driver.usb: sending control transfer with (64, 2, 2), {}
[DEBUG] liquidctl.driver.asetek_pro: Pro configure device...
[DEBUG] liquidctl.driver.asetek_pro: pump_mode balanced
[DEBUG] liquidctl.driver.asetek_pro: write 32:01
[DEBUG] liquidctl.driver.usb: writting 2 bytes: 32:01
[DEBUG] liquidctl.driver.usb: read 3 bytes: 32:12:34
[DEBUG] liquidctl.driver.asetek_pro: received 32:12:34
[DEBUG] liquidctl.driver.usb: ensure interface is released
[DEBUG] liquidctl.driver.usb: ensure interface is released
```

```
# liquidctl initialize --pump-mode quiet --debug
[DEBUG] liquidctl.cli: running liquidctl v1.4.2 (879144f6477c)
[DEBUG] liquidctl.driver.smbus: skipping LinuxI2c, smbus package not available
[DEBUG] liquidctl.driver.usb: searching HidapiBus (CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver)
[DEBUG] liquidctl.driver.usb: found HID device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found HID device 0b05:1872
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (Hydro690Lc, Legacy690Lc, Modern690Lc, _CommonAsetekDriver, CorsairAsetekProDriver)
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found USB device 04a9:221b
[DEBUG] liquidctl.driver.usb: found USB device 046a:0023
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 8087:0029
[DEBUG] liquidctl.driver.usb: found USB device 0b05:1872
[DEBUG] liquidctl.driver.usb: found USB device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:0c12
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: found USB device 05e3:0610
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.cli: device: Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: selected interface: 0
[DEBUG] liquidctl.driver.asetek: set clear to send = True
[DEBUG] liquidctl.driver.usb: sending control transfer with (64, 2, 2), {}
[DEBUG] liquidctl.driver.asetek_pro: Pro configure device...
[DEBUG] liquidctl.driver.asetek_pro: pump_mode quiet
[DEBUG] liquidctl.driver.asetek_pro: write 32:00
[DEBUG] liquidctl.driver.usb: writting 2 bytes: 32:00
[DEBUG] liquidctl.driver.usb: read 3 bytes: 32:12:34
[DEBUG] liquidctl.driver.asetek_pro: received 32:12:34
[DEBUG] liquidctl.driver.usb: ensure interface is released
[DEBUG] liquidctl.driver.usb: ensure interface is released
```

```
# liquidctl initialize --pump-mode performance --debug
[DEBUG] liquidctl.cli: running liquidctl v1.4.2 (879144f6477c)
[DEBUG] liquidctl.driver.smbus: skipping LinuxI2c, smbus package not available
[DEBUG] liquidctl.driver.usb: searching HidapiBus (CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver)
[DEBUG] liquidctl.driver.usb: found HID device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found HID device 0b05:1872
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 046a:0023
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found HID device 04d9:fc06
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (Hydro690Lc, Legacy690Lc, Modern690Lc, _CommonAsetekDriver, CorsairAsetekProDriver)
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 04d9:fc06
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 174c:3074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:1b49
[DEBUG] liquidctl.driver.usb: found USB device 04a9:221b
[DEBUG] liquidctl.driver.usb: found USB device 046a:0023
[DEBUG] liquidctl.driver.usb: found USB device 174c:2074
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0003
[DEBUG] liquidctl.driver.usb: found USB device 8087:0029
[DEBUG] liquidctl.driver.usb: found USB device 0b05:1872
[DEBUG] liquidctl.driver.usb: found USB device 0b05:18f3
[DEBUG] liquidctl.driver.usb: found USB device 1b1c:0c12
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: found USB device 05e3:0610
[DEBUG] liquidctl.driver.usb: found USB device 1d6b:0002
[DEBUG] liquidctl.cli: device: Corsair Hydro H150i Pro
[DEBUG] liquidctl.driver.usb: selected interface: 0
[DEBUG] liquidctl.driver.asetek: set clear to send = True
[DEBUG] liquidctl.driver.usb: sending control transfer with (64, 2, 2), {}
[DEBUG] liquidctl.driver.asetek_pro: Pro configure device...
[DEBUG] liquidctl.driver.asetek_pro: pump_mode performance
[DEBUG] liquidctl.driver.asetek_pro: write 32:02
[DEBUG] liquidctl.driver.usb: writting 2 bytes: 32:02
[DEBUG] liquidctl.driver.usb: read 3 bytes: 32:12:34
[DEBUG] liquidctl.driver.asetek_pro: received 32:12:34
[DEBUG] liquidctl.driver.usb: ensure interface is released
[DEBUG] liquidctl.driver.usb: ensure interface is released
```

The pump mode is working properly, I can see it from the monitoring + from the sounds that it makes + from the evolution of the temperature.

jonasmalacofilho:
@cluxter oh, good to know, thanks again! : )

cluxter:
Is it possible to set the pump mode according to a level of temperature? Like with the fans, so the pump could adjust its mode automatically.

jonasmalacofilho:
@cluxter not in the firmware (that I'm aware), but you could write a simple script to do that with the liquidctl APIs.  That said, do you really want to switch pump speed × temperature curves (aka pump "modes") based on the temperature?

cluxter:
Yes this is what I want. Here is my logic:
```
set fan1 speed 25 0 35 100
set fan2 speed 25 0 32 100
set fan3 speed 25 0 30 100
```
I do this because I noticed that liquid temperature tends to be between 25 and 40, and I want to keep the temperature pretty low (ideally not over 30), but I also don't want to unnecessarily wear out the fans too fast, so there is no reason to put it on max at 25°C.

Also, fan3 is where the hot liquid comes in, fan2 is in the middle and fan1 is on the coldest end, so I put more pressure on fan3, a little less on fan2 and again less on fan1.

But I noticed that the pump speed is actually what has the most impact on temperature. So I would like to have something like this, on top of the fan speeds:

```
set pump mode 0 quiet 30 balanced 40 performance
```
The idea being: I follow the logic about keeping the temperature below 30, but I let the fans handle what's between 30 and 40 because usually it doesn't reach 40. If it does, it means that the CPU is handling a very heavy task, like a compilation on all the cores for at least 5 minutes, and in this case I want to do everything I can to cool it down because who knows how long this will last, and I don't want to take any risk. But I never want to handle all this manually, it's the computer's job.

So yes, I could write a script to get this feature. However, do you have a particular reason not to include such a feature in `liquidctl`? I assume I could write this part if need be.

jonasmalacofilho:
We don't directly implement dynamic/software control like what you're suggesting because doing that in a generic way always comes with significant tradeoffs, especially when you consider that we support many different devices and platforms.

The idea is to expose all hardware/firmware capabilities through the APIs (and, as much as possible, through the CLI), and allow power users to build their own complex control algorithms on top of them.  How hard that is in practice depends on the specific device the user is interested in, as well as how complex algorithm their are trying to implement is.

That said, we're more than happy to take and distribute these device or feature specific scripts in `extra/contrib`.

---

But given that 'performance' is a curve already, why not just have the pump in that mode all the time?

cluxter:
I think I understand your logic, but what I don't get is: we are doing this dynamic control on fans already. So how does it differ from the pump? If we are setting fan speeds according to temperature, why couldn't we set the pump modes according to some temperature thresholds? (no curve needed here) To me, the pump is kind of like a fan on which you can only set modes and not speeds.
___
I don't want to have the pump on the same mode all the time for several reasons:
1) `performance` mode is louder and I want to reduce noise as much as possible. I work in a quiet environment and I can really hear the difference between `performance` and the two other modes. Even if, reducing noise pollution is something I like, even when it's a little noise, because all noises add up, so I try to minimize the noise sources as much as possible;
2) `performance` mode consumes more energy. Probably not a lot, but it's always nice to optimize things when possible in my opinion. Same logic than 1);
3) the more you solicit your software, the faster it gets obsolete. I try to make hardware last: better for my budget, the planet, and my time since changing one part is time consuming.

liquidctl/driver/asetek_pro.py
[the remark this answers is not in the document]
andytank123: I've removed the devices not on the Corsair product pages

liquidctl/driver/asetek_pro.py
andytank123: I've changed this to raise `ValueError`s. I've also added a check to make sure the fan specified exists (so `fan99` should now error).

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,240 @@
...
+        """Send the speed to cycle colors on the RGB pump"""
+        speed = clamp(speed, 1, 3)
+        self._write([0x53, _PRO_SPEEDS[mode][int(speed)-1]])
+        self._end_transaction_and_read(3)
+
+    def _write_color(self, colors):
+        """Write a list of colors to cycle between, the cycle pattern comes from the
+        value used for cycle speed"""
+        setColors = list(itertools.chain(*colors))
+        colorCount = int(len(setColors) / 3)
```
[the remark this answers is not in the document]
andytank123: Done

liquidctl/driver/asetek_pro.py
```diff
@@ -0,0 +1,240 @@
...
+    'pulse': [0x52, 0x01],
+    'blinking': [0x58, 0x01],
+    'fixed': [0x55, 0x01]
+}
+
+
+class CorsairAsetekProDriver(_CommonAsetekDriver):
+    """liquidctl driver for Corsair-branded sixth generation Asetek coolers."""
+    SUPPORTED_DEVICES = [
+        (0x1b1c, 0x0c12, None, 'Corsair Hydro H150i Pro', {}),
```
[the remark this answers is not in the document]
andytank123: Added

jonasmalacofilho:
> we are doing this dynamic control on fans already

We aren't, the firmware supports custom fan curves.

Which is why it's working at all, considering that the liquidctl CLI is not a daemon continuously pushing updated fan settings.  Rather, it just hands the curve off to the firmware and exists.

> To me, the pump is kind of like a fan on which you can only set modes and not speeds.

I agree, but Corsair apparently has decided it was better to differentiate the two, simplifying (aka limiting) the control over the pump.

> `performance` mode is louder and I want to reduce noise as much as possible

Oh, that's too bad.  It seems that instead of (or in addition to) making the pump curve more aggressive in 'performance', Corsair also chose to offset it.  I agree that this is less than ideal, but it's a firmware issue.

---

But hey, I can help you write the script...  It's not that it doesn't have merit, it just doesn't belong in the main liquidctl CLI.

liquidctl/driver/asetek_pro.py
andytank123: The only RGB the Corsair PRO comes with is on the pump. If you've added your own RGB fans, I'm not sure if the driver should control them. 
I've changed the set color function so the only channel that works is now `pump` to make this clear.

jonasmalacofilho:
@andytank123 

You can also remove all logging of writes and reads, since that's already handled by `device.write`/`device.read`.

cluxter:
@jonasmalacofilho 

> We aren't, the firmware supports custom fan curves.
Oooh I understand now.

> Which is why it's working at all, considering that the liquidctl CLI is not a daemon continuously pushing updated fan settings. Rather, it just hands the curve off to the firmware and exists.
It totally makes sense now. Thanks!
___
> But hey, I can help you write the script... It's not that it doesn't have merit, it just doesn't belong in the main liquidctl CLI.
Absolutely, I totally agree with this approach.
I will write a simple script on my own for the moment and I will add it to `extra/contrib` once I will be satisfied with what I wrote. Then maybe you can review it and guide me in order to improve it. Thank you very much for your dedication and your help on all this!

PS: I still have to test fan colors, I'll keep you posted on this soon.

liquidctl/driver/asetek_pro.py
jonasmalacofilho: Is there an RGB header on the pump for non-stock RGB fans?

andytank123: No, the pump has 2 or 3 (depending on model) 4-pin fan headers. However it doesn't have any RGB headers.

andytank123:
I've finally got my Corsair H100i PRO installed and setup, this has allowed me to test the driver and confirm it works. If there are any more changes you would like me to make, please let me know.

jonasmalacofilho:
Thank you, I'll take a look as soon as I can.

Can you also try to [capture](https://github.com/jonasmalacofilho/liquidctl/blob/master/docs/developer/capturing-usb-traffic.md) some of the traffic between iCue and the device, and submit the data to [liquidctl-device-data](https://github.com/jonasmalacofilho/liquidctl-device-data)?

---

It would also be ideal if a somewhat formal description of the protocol was added to `docs/developer/protocol/`.  But I acknowledge that this is a new thing my part, and it's not strictly required yet.

Still, the benefit is that it gives future maintainers, as well as those working on kernel drivers, a high level view of how the protocol works, rather than relying on "hey, this specific implementation is know to work".

P.S. Right now I'm writing one of these descriptions for the cooler I use, and that may be a better reference for you than the only other document currently in that folder, which has to do with SMBus and DDR4 modules.  And I'll probably only get to your PR when I'm done with this document anyway.

jonasmalacofilho:
Thank you again @andytank123 ,

There were a few issues that I figured (in December) were too small to justify requesting additional changes...

Unfortunately, doing those myself was a rabbit hole, and I ended up noticing more small issues.  That's ok, but I'm going to close this PR, as by now I have spent substantial effort on changes that are not merge-friend.

Once I'm done for the weekend I'll open a new issue to track eventual remaining problems or things that still need to be tested.  Also, I also *really* need those traffics captures (and/or comprehensive documentation of the protocol).

---

P.S. I started to make the changes in December, and had planned to finish them this weekend, (just) in time for the pre-release freeze.  But considering that my changes are large enough to require testing on real hardware, that there are open questions regarding how the device works, that there is no documentation written yet, that new CLI options may be necessary, and that the unit tests need to be ported to pytest (and perhaps also extended), this simply wont make it for 1.5.0.

P.P.S. For those that were anxious to see this driver in 1.5.0: drivers for devices which the maintainers don't have access to, and that don't come from long-term (and known to stick around) contributors, require substantial scrutiny, even on the small or cosmetic things, because by taking in code we implicitly demonstrate our intent to maintain it.  And maintaining code we cannot personally (really) test is not that easy...

PR #254, opened 2020-12-30
added systemd sample service

### LINKED ISSUE #139: attempt to create systemd service
I am trying  to create a service so liquidctl starts and applies my custom curve at boot, and not too sure if i'm doing this correctly. Help is very much appreciated:
/etc/udev/rules.d/71-liquidctl.rules:
```
# Corsair H115i Platinum
SUBSYSTEMS=="usb", ATTRS{idVendor}=="1b1c", ATTRS{idProduct}=="0c17", TAG+="uaccess"
SUBSYSTEMS=="usb", ATTRS{idVendor}=="1b1c", ATTRS{idProduct}=="0c17",TAG+="systemd"
```

/etc/systemd/system/liquidcfg.service;
```
[Service]
Type=oneshot
ExecStart=liquidctl initialize --pump-mode balanced
ExecStart=liquidctl set fan speed 27 0 29 15 30 30 32 40 34 60 36 85 38 100 
After=sys-devices-pci0000:00-0000:00:01.2-0000:20:00.0-0000:21:08.0-0000:2a:00.3-usb3-3\x2d6-3\x2d6.4.device
[Install]
WantedBy=default.target 
```
I'm almost sure that the After clause is not correct, i think the second udev rule that creates de unit file is wrong

MarshallAsch:
closes #139 

Added a systemd service sample file. I'm pretty sure that this should resolve the discussion that was going on in that issue.

### DIFF
--- a/docs/linux/making-systemd-units-wait-for-devices.md
+++ b/docs/linux/making-systemd-units-wait-for-devices.md
@@ -30,6 +30,16 @@ After=dev-clc120.device
 ...
 ```
 
+### Alternate Format
+
+Instead of adding a `SYMLINK` to the udev rules the `TAG+="systemd"` field can be added instead.
+The device dependancy can be specified using the format `dev-bus-usb-<bus>-<device>.device` you can get the bus
+and the device number by using the `lsusb` command.
+The service can be set to wait for multiple devices to be ready by adding more devices to the list separated by spaces.
+
+An example systemd configuration file can be found [here](../../extra/linux/liquidcfg.service), in that example there
+are two devices that get initialized and then the fan speed gets set.
+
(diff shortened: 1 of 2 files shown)


jonasmalacofilho:
I didn't know about the default `dev-bus-...`, nice!

I reword this a bit and added it in 588a04f5bbd8c0eda775321a8bd61dc9efdeceae.  I left the example unit out, because I don't feel it should be in extra, and also because IMO it doesn't really add that much on top of other examples in the docs.

---

This won't close #139 because what I want to mention somewhere is that just as a service unit can depend on one or more devices, a device (unit) can be made to run a service automagically.

Issue #253, opened 2020-12-30
Update drivers to use the new direction flags


MarshallAsch:
Re: https://github.com/liquidctl/liquidctl/pull/213#discussion_r549895314

Change the mode strings for these devices to use the direction flag to specify forwards or backwards

Devices that need updating:

- [x] kraken x2 m2
- [x] kraken x3 z3
- [x] nzxt hue2
- [x] nzxt smart device v1

PR #255, opened 2020-12-31
direction flag change

MarshallAsch:
Closes: #253

Re: https://github.com/liquidctl/liquidctl/pull/213#discussion_r549895314

Change the mode strings for these devices to use the direction flag to specify forwards or backwards

This should be merged after #213


MarshallAsch:
I added back the deprecated `backward` modes for backwards compatibility purposes.

Issue #231, opened 2020-11-23
Code style and formatting refactor


MarshallAsch:
As of the new style guide (that is scheduled to be added in #230) a code refactor is in order. 

Things to clean up
- [x] check and fix the formatting of log messages
- [x] check and fix the formatting of exception messages
- [x] check how values are entered into log messages (format strings vs lazy formatting) 
- [x] how unsafe flags are handled in drivers (but may need to be revisited in #232)
- [x] clean up code style to make the drivers more consistent
- [x] check where exceptions are thrown that they make sense and include a `from None` where it is relevant
- [x] `LOGGER` vs. `_LOGGER` consistency in the drivers`
- [x] trailing whitespace on lines and `flake8` linter warnings / errors 
- [x]  small README fix - re: https://github.com/jonasmalacofilho/liquidctl/pull/205#discussion_r529479623

check formatted string generation  "[PEP 498](https://www.python.org/dev/peps/pep-498/)  encourages the use of f-strings most of the time."

PR #256, opened 2021-01-02
Migrate to pytest

MarshallAsch:
In partial completion of #231. 

This changes all of the tests to use the python pytest module instead of the UnitTest module. Aside from the `hydro platinum` moving where the `RuntimeStorage` module gets set all of the tests and the code should be the same. 

One of the `hydro platinum`  test case to ensure that the `RuntimeStorage`  is in the correct spot was removed.

### DIFF
--- a/liquidctl/driver/hydro_platinum.py
+++ b/liquidctl/driver/hydro_platinum.py
@@ -137,20 +137,24 @@ class HydroPlatinum(UsbHidDriver):
             ('led', 'off'): 0,
         }
 
+        # the following fields are only initialized in connect()
+        self._data = None
+        self._sequence = None
+
+    def connect(self, **kwargs):
+        """Connect to the device."""
+        super().connect(**kwargs)
+
         ids = f'vid{self.vendor_id:04x}_pid{self.product_id:04x}'
         # must use the HID path because there is no serial number; however,
         # these can be quite long on Windows and macOS, so only take the
         # numbers, since they are likely the only parts that vary between two
         # devices of the same model
         loc = 'loc' + '_'.join(re.findall(r'\d+', self.address))
-        self._data = RuntimeStorage(key_prefixes=[ids, loc])
 
-        # the following fields are only initialized in connect()
-        self._sequence = None
-
-    def connect(self, **kwargs):
-        """Connect to the device."""
-        super().connect(**kwargs)
+        if self._data is  None:
+            self._data = RuntimeStorage(key_prefixes=[ids, loc])
+        
         self._sequence = _sequence(self._data)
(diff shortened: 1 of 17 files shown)


tests/test_hydro_platinum.py
```diff
@@ -34,182 +46,200 @@ def read(self, length):
...
+        dev.set_color('led', 'invalid', [])
+
+    with pytest.raises(ValueError):
+        dev.set_color('invalid', 'off', [])
+
+    assert len(dev.device.sent) == 0
+
+def test_h115i_platinum_device_short_enough_storage_path(h115iPlatinumDevice):
+    dev = h115iPlatinumDevice
+    #this test should not be here, the storage backed is tested seperately
```
jonasmalacofilho, line 239: The reason why this was tested here is that it depends on what IDs are used to initialize the RuntimeStorage, which is driver dependent.  Also, this was a regression test, so if possible it should be restored.

liquidctl/driver/hydro_platinum.py
```diff
@@ -136,20 +136,21 @@ def __init__(self, device, description, fan_count, rgb_fans, **kwargs):
             ('led', 'fixed'): 1,
             ('led', 'off'): 0,
         }
-        # the following fields are only initialized in connect()
-        self._data = None
-        self._sequence = None
 
-    def connect(self, **kwargs):
-        """Connect to the device."""
-        super().connect(**kwargs)
```
jonasmalacofilho, line 145: Is there any way around this change?  Otherwise simply probing for devices will cause directories to be created for this device.

Also, why is this only a problem on this driver, and not others that use the RuntimeStorage API?

jonasmalacofilho (commented):
Awesome, thanks!

Can you just check these two comments?

tests/test_hydro_platinum.py
MarshallAsch: okay, Ill add that test back using the real storage driver and not the mocked one.

liquidctl/driver/hydro_platinum.py
MarshallAsch: thats a good question. Im not sure why it didn't come up with any of the other devices, I'll take another look at that.

MarshallAsch: I moved it back to connect, the tests were only failing here before because of the sequence number part, was impacted by other tests. The new check to only set the `RuntimeStorage` if data is None will fix that behaviour for the tests so they continue to use the mock, but will not affect any real runs.

jonasmalacofilho (approved):

PR #266, opened 2021-01-21
Lighting Node Core Support

### LINKED ISSUE #154: Corsair Lighting Node Core Support
Hello, 

I've just started using this and I was wondering if there would be any support for the Corsair Lighting Node? I've attached the system profiler data below. I believe this would be an extension of the implementation from the corsair hydro series. 

``` 
CORSAIR Lighting Node CORE:
              Product ID: 0x0c1a
              Vendor ID: 0x1b1c  (CORSAIR MEMORY INC.)
              Version: 0.09
              Serial Number: 36C0020<snip>C00F5
              Speed: Up to 12 Mb/s
              Manufacturer: Corsair
              Location ID: 0x14a40000 / 8
              Current Available (mA): 500
              Current Required (mA): 50
              Extra Operating Current (mA): 0
```

MarshallAsch:
Closes #154.

This enables support for the Corsair Lighting Node Core.

- [x] update man pages with new device
- [x] update readme with new device
- [x] added device id to the commander pro driver
- [x] added device usage to the commander pro driver docs
- [x] create tests for the new device
- [x] create tests for the new limited number of effect commands
- [x] figure out why led channel 2 does not seem to be working

### DIFF
--- a/liquidctl/driver/commander_pro.py
+++ b/liquidctl/driver/commander_pro.py
@@ -121,6 +121,8 @@ class CommanderPro(UsbHidDriver):
             {'fan_count': 6, 'temp_probs': 4, 'led_channels': 2}),
         (0x1b1c, 0x0c0b, None, 'Corsair Lighting Node Pro (experimental)',
             {'fan_count': 0, 'temp_probs': 0, 'led_channels': 2}),
+        (0x1b1c, 0x0c1a, None, 'Corsair Lighting Node Core (experimental)',
+            {'fan_count': 0, 'temp_probs': 0, 'led_channels': 2}),
     ]
 
     def __init__(self, device, description, fan_count, temp_probs, led_channels, **kwargs):
@@ -201,7 +203,7 @@ class CommanderPro(UsbHidDriver):
         """
 
         if self.device.product_id != 0x0c10:
-            _LOGGER.debug('only the commander pro supports this')
+            _LOGGER.debug('only the Commander Pro supports this')
             return []
 
         connected_temp_sensors = self._data.load('temp_sensors_connected', default=[0]*self._temp_probs)
@@ -444,8 +446,8 @@ class CommanderPro(UsbHidDriver):
 
         direction = _LED_DIRECTION_FORWARD if direction == 'forward' else _LED_DIRECTION_BACKWARD
         speed = _LED_SPEED_SLOW if speed == 'slow' else _LED_SPEED_FAST if speed == 'fast' else _LED_SPEED_MEDIUM
-        start_led = clamp(start_led, 1, 96) - 1
-        num_leds = clamp(maximum_leds, 1, 96-start_led-1)  # there is a current firmware limitation of 96 led's per channel
+        start_led = clamp(start_led, 1, 255) - 1
+        num_leds = clamp(maximum_leds, 1, 255 - start_led - 1)
         random_colors = 0x00 if mode_str == 'off' or len(colors) != 0 else 0x01
         mode = _MODES.get(mode, -1)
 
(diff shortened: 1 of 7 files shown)


liquidctl/driver/commander_pro.py
```diff
@@ -466,6 +468,12 @@ def set_color(self, channel, mode_str, colors, direction='forward', speed='mediu
         saved_effects = [] if mode_str == 'off' else self._data.load('saved_effects', default=[])
         saved_effects += [lighting_effect]
 
+        # check to make sure that too many LED effects are not being sent.
+        # the max seems to be 8 as found here https://github.com/liquidctl/liquidctl/issues/154#issuecomment-762372583
+        if len(saved_effects) > 8:
+            _LOGGER.warning(f'too many lighting effects. Run `liquidctl set {channel} color clear` to reset the effect')
+            return
```
jonasmalacofilho, line 475: Shouldn't we maybe implement something like a ring buffer or some predetermined slots, instead of requiring user intervention?

After all, this is at least in part *our* abstraction, right?

liquidctl/driver/commander_pro.py
```diff
@@ -444,8 +446,8 @@ def set_color(self, channel, mode_str, colors, direction='forward', speed='mediu
 
         direction = _LED_DIRECTION_FORWARD if direction == 'forward' else _LED_DIRECTION_BACKWARD
         speed = _LED_SPEED_SLOW if speed == 'slow' else _LED_SPEED_FAST if speed == 'fast' else _LED_SPEED_MEDIUM
-        start_led = clamp(start_led, 1, 96) - 1
-        num_leds = clamp(maximum_leds, 1, 96-start_led-1)  # there is a current firmware limitation of 96 led's per channel
+        start_led = clamp(start_led, 1, 255) - 1
+        num_leds = clamp(maximum_leds, 1, 255 - start_led - 1)
```
jonasmalacofilho, line 450: I think we can use 136 (4x34) for now.

MarshallAsch: I think it might actually be 136, but I cant seem to find where that was documented again so I figured Id just limit it to one byte. I dont think anything bad will happen if this number is too big.

liquidctl/driver/commander_pro.py
MarshallAsch: I thought about that, but I feel like it might lead to some confusion if previously sent (ie the first) lighting effects  just stop being applied when more are sent. And because the order that the effects are applied matters I felt this ensures that the user knows they have to rethink their effect if they hit this limit.

jonasmalacofilho: Fair enough.

liquidctl/driver/commander_pro.py
jonasmalacofilho: I tend to prefer to be more strict in places like this one (easy fixes if necessary, but and the code more clearly states our [current] beliefs about what works and what doesn't), but I suppose it almost doesn't make a difference.  Feel free to keep it 255 (but I think it's fine to clamp it to 136 and, if necessary, change this again in the future).

MarshallAsch:
Waiting for @kevinlester to confirm that the `led2` issue is fixed and that the LNC does not actually have a channel 2, instead it has one channel.

liquidctl/driver/commander_pro.py
MarshallAsch: So that number was wrong too, based on [this](https://forum.corsair.com/v3/showthread.php?t=173880) link there can be up to 6 QL fans with 34 LED's each (a total of 204 LEDs). So I am updating the value to that.

MarshallAsch:
Correct functionality of the driver was confirmed in https://github.com/liquidctl/liquidctl/issues/154#issuecomment-767218552

Issue #242, opened 2020-12-10
Corsair H100i Platinum SE not setting leds properly


brandon-feder:
Hello. When my computer is booting, both the fans ( and the pump ) light up with there default RGB barf. However, once the liquidcfg service is started, only one fan (and the pump) light up while the other fan does not. Both fans are spinning, but only one lights up. I am not sure whether this is a hardware or software issue, but I was hoping that if it is a software issue, that someone would help. I am running Manjaro Linux if that matters. Any other info you need?

Thanks in advance,
Brandon Feder

jonasmalacofilho:
Hi @brandon-feder,

What specific variant of the H100i do you have?  Can you also post which specific liquidctl commands you are using, as well as their corresponding output with `--debug`?

MarshallAsch:
Not quite related to your immediate problem but there is a fix in progress for that "default RGB barf" that shows up while the computer is booting. (see #193 for more details on that.)

brandon-feder:
Hello. Thanks for your reply. I have a H100i RGB PLATINUM SE. The only commands I run are `sudo liquidctl initialize` and `sudo liquidctl set led color fixed ffffff`. Its output using the `--debug` flag is the following:

```
[DEBUG] liquidctl.cli: running liquidctl v1.4.2
[DEBUG] liquidctl.driver.usb: searching HidapiBus (api=hid, drivers=[CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver])
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c19
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair H100i Platinum SE (experimental)
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[DEBUG] liquidctl.driver.usb: probing drivers for device 046d:408a
[...]

PR #269, opened 2021-01-22
Corsair H100i Platinum SE Number of LEDs fix

### LINKED ISSUE #267: Setting LED colour on Corsair H100i Platinum SE only sets colour for pump and one fan.
Pretty much the title. Fan/pump control works great as expected but setting RGB colour seems to only be half working. 

Running `sudo liquidctl initialize all` shows this:
```
Corsair H100i Platinum SE (experimental)
└── Firmware version    1.1.15
```

Running `sudo liquidctl set led color off` works and turns all of the LEDs off for the pump, fan 1 and fan 2.

But running `sudo liquidctl set led color fixed ffffff` only sets the colour for the pump and fan 1 LED while fan 2 LED remains off.

I'm running a plain Debian 10 Buster.
IMG_2828

MarshallAsch:
Closes #267 and #242.

Changed support to work with a different number of LED's that come with the different stock Hydro Platinum AIO coolers.

This changes one of the options that is passed to the constructor for the `HydroPlatinum` driver. Changed `rgb_fans`  boolean option in the HydroPlatinum driver to a `fan_leds` integer option that is set to the number of LEDs that are on each fan that comes with the cooler.

For the SE cooler that comes with LL fans (16 leds each) this bumps the total number of LEDs to 48, which is too many for to fit in the 2 color commands that are sent. Because of that a 3rd color message is now sent when necessary. 
~~There should be no change for any existing devices.~~

### DIFF
--- a/liquidctl/driver/hydro_platinum.py
+++ b/liquidctl/driver/hydro_platinum.py
@@ -299,7 +299,9 @@ class HydroPlatinum(UsbHidDriver):
         data3 = bytes(itertools.chain(*((b, g, r) for r, g, b in expanded[40:])))
 
         self._send_command(_FEATURE_LIGHTING, _CMD_SET_LIGHTING1, data=data1)
-        self._send_command(_FEATURE_LIGHTING, _CMD_SET_LIGHTING2, data=data2)
+        
+        if self._led_count > 20:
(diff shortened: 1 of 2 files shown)


MarshallAsch:
I think that this can also resolve some of the comments in #138 about the color commands.

PR #273, opened 2021-01-24
Added a new driver github issue template

MarshallAsch:
As the title suggests, this PR will add a new issue template for proposing new device drivers to liquidctl that asks for information that that is often part of the early conversation about adding support. 

This also adds a default pull request template that includes several reminders for things that are frequently forgotten.

Also bumped the pipeline checkout action version.


.github/ISSUE_TEMPLATE/new-driver-request.md
```diff
@@ -0,0 +1,36 @@
...
+**Awesome, it seems like you have a brand new device**
+
+- Would you be willing to test potential implementations once they are complete: *Yes/No*
+- Would you be able to do the protocol analyzing / implementation effort: *Yes/No*
+
+
+----
+Include this for any USB based device:
+```
+<put the output of lsusb -v  here>
```
jonasmalacofilho, line 32: No all users are running Linux.

MarshallAsch: mac would be `system_profiler SPUSBDataType` Do you know what the command would be for windows?

MarshallAsch:
@jonasmalacofilho do you think this is a good idea? I know a lot of the new device requests frequently have the the same things asked and this way it makes it nice and easy to have the questions already there when a user makes the request. 

(Im also working on a default PR template with frequently forgotten steps)

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+Closes # .
+
+## Summary of changes
+
+This added X awesome new feature....
+
+---
+### Tasks that must be compleeted:
+- [ ] Update changelog
```
jonasmalacofilho, line 9: I rather leave updating the changelog to myself (or, at most, to us the maintainers).

jonasmalacofilho:
Yes, these templates will probably help with the first round of communication in issues or PRs.

.github/pull_request_template.md
jonasmalacofilho: We should also mention the style guide.

Which reminds me, I've been also thinking about covering the commit messages in the style guide (or in a separate commit message guide): a summary of the general "good commit message" + the tagging approach I've landed on.  What do you think about this (and about my style of commit messages)?

MarshallAsch: > I rather leave updating the changelog to myself (or, at most, to us the maintainers).

Ill take that out then. 

> Which reminds me, I've been also thinking about covering the commit messages in the style guide...

I haven't actually looked too closely at your commit messages beyond noticing that they are frequently multiple lines. 
I generally just do frequent `git commit -am "my amazing changes here"` and keep them fairly short and I think they are human readable but not all that descriptive .

MarshallAsch:
I was thinking of also adding another issue template for bug reports, not quite sure what to put in that though. (Im also going to move this out of draft state because I am good with it, but might want to add more after some discussion)

Maybe, also add a contributing doc?

MarshallAsch:
Need to add the command to get smbus device info. (vendor, and product ID's). then I think this is good?

jonasmalacofilho:
FYI, I like the idea, but I want to try some slight changes in wording, use of emphasis, etc.  I think there's a subtle balance to be found, where a few typical comment-reply iterations can be skipped but where we still remain quite flexible.

_(For one, we can't require the same from all users, because some tasks can be quite hard to do on Windows (or for Windows-only users); for example, we want contributors to update the man page when applicable, however that's already fairly obscure for most Linux/Mac users, and absolutely unfeasible for Windows-only users)._

I also want to try and see how this looks in the GitHub UI, from the perspective of a user.  Lately[1] I haven't seen many repos using these templates, and I don't even remember how having them impacts the flow from project landing page to <action>.  I'll create a temporary repository to test this.

> Maybe, also add a contributing doc?

We should probably have one, yes.  And an explicit code of conduct (although [I think/hope I'm right that] this hasn't been necessary so far).

---

[1] I have the impression that once there were more repos experimenting with these templates.

PR #283, opened 2021-02-06
Updated instructions for capturing traffic

kevinlester:
Problem: The setup required to enable the capture of network traffic between a source application (usually running in a guest VM) and a device is non-trivial.  The few hoops that people must go through in the setup can lead to frustration and be a deterrent for users providing helpful information to the dev community.  The current instructions for how to set up an environment for capturing traffic are correct, but leave off some the nuances.

Solution: More detailed documentation on how to set up the network capture.  I created these based off my own experiences with a fresh Ubuntu install and capturing traffic between Corsair iCue running in a Windows 10 guest VM and my Corsair Lightning Node Core device.


jonasmalacofilho:
I appreciate the help in improving that document.

However, you made it totally about Windows VM on a LInux host, which can also be a deterrent to our other users on Windows and Mac OS.  The captures they have provided so far (some of them made directly on native Windows installations) have also been very useful to us.

IMO, you should bring back the previous structure, with sections for capturing on Windows, for capturing on Linux (with a Windows VM), and for finding the correct device.  I say IMO because other document structures could also work, as long as they aren't so tightly coupled to the Linux use case.

Additionally, please bring back the section about exporting the data, or at least the mention to JSON exports.

---

> You will need a free Microsoft account to download the ISO,

You probably checked this, but there *used* to be a way to opt-out from that.  Is it really gone?

> but you do not need a license (assuming you are just using the VM for testing).

I think the idea is that testing is ok only in the sense of _evaluating_ Windows or programs on Windows.  Capturing USB traffic from a Windows program is not testing in that sense.

If you're sure that this is clearly permitted under the EULA and/or applicable laws, please keep the sentence but provide a citation and links to the relevant sources.  Otherwise, just remove it.

---

Finally, there are a couple of nits I would also like to ask you to address:

- the readers will assume the document was written by one of the maintainers (and in the end we're indeed responsible for the document); I find the use of the first person odd in something I didn't write, but that will likely be associated with me, so i would like to ask you to make the text more impersonal
- I don't think it's necessary to teach how to install virt-manager on Debian/Ubuntu; if you think it is, then instructions for other common Linux distros should also be provided

kevinlester:
I moved the VM creation steps out to a separate doc so that I could keep the previous document flow generally the same.  Addressed the PR comments.

jonasmalacofilho:
Thank you!

PR #285, opened 2021-02-09
added basic file locking capabilities for unix and windows

### LINKED ISSUE #274: Running two liquidctl processes causes Corsair H100i PRO XT to stop responding to read requests.
I've noticed this twice now; after a while (don't think it's X time but rather X read requests) my H100i PRO XT stops responding. The reason for me thinking it's X reads is that both times this has happend I've had a `while 1; do sleep; liquidctl status; done` loop running. 

I've tried rm-ing usbhid module and loading it again.

I can write, but the process hangs after this:

```
root@pve:~# liquidctl initialize -g                                                                                                                                                                                                                                                                                 
[DEBUG] liquidctl.cli: running liquidctl v1.4.2                                                                                                                                                                                                                                                                     
[DEBUG] liquidctl.driver.usb: searching HidapiBus (api=hidraw, drivers=[CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver])                                                                                                      
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c20                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair H100i PRO XT (experimental)                                                                                                                                                                                                                              
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (drivers=[Hydro690Lc, Legacy690Lc, Modern690Lc, _CommonAsetekDriver])                                                                                                                                                                                              
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c20                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 8087:0024                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 1d6b:0002                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 8087:0024                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 1d6b:0002                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.cli: device: Corsair H100i PRO XT (experimental)                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.keyval: data in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0                                                                                                                                                                                                                                        
[DEBUG] liquidctl.keyval: stored leds_enabled=0 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/leds_enabled)                                                                                                                                                                                                        
[DEBUG] liquidctl.keyval: stored pump_mode=1 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/pump_mode)                                                                                                                                                                                                              
[DEBUG] liquidctl.keyval: no data (file) found for fan1_mode                                                                                                                                                                                                                                                        
[DEBUG] liquidctl.driver.hydro_platinum: falling back to FIXED_DUTY for _FanMode(None)                                                                                                                                                                                                                              
[DEBUG] liquidctl.keyval: no data (file) found for fan1_duty                                                                                                                                                                                                                                                        
[INFO] liquidctl.driver.hydro_platinum: setting fan1 to 100% duty cycle                                                                                                                                                                                                                                             
[DEBUG] liquidctl.keyval: no data (file) found for fan2_mode                                                                                                                                                                                                                                                        
[DEBUG] liquidctl.driver.hydro_platinum: falling back to FIXED_DUTY for _FanMode(None)                                                                                                                                                                                                                              
[DEBUG] liquidctl.keyval: no data (file) found for fan2_duty                                                                                                                                                                                                                                                        
[INFO] liquidctl.driver.hydro_platinum: setting fan2 to 100% duty cycle                                                                                                                                                                                                                                             
[INFO] liquidctl.driver.hydro_platinum: setting pump mode to balanced                                                                                                                                                                                                                                               
[DEBUG] liquidctl.keyval: loaded sequence=25 (from /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/sequence)                                                                                                                                                                                                             
[DEBUG] liquidctl.keyval: stored sequence=26 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/sequence)                                                                                                                                                                                                               
[DEBUG] liquidctl.driver.usb: discarded 0 previously enqueued reports                                                                                                                                                                                                                                               
[DEBUG] liquidctl.driver.usb: writting report 0x00 with 64 bytes: 3f:d0:14:00:ff:05:ff:ff:ff:ff:ff:02:00:00:00:00:ff:02:00:00:00:00:ff:01:00:00:00:00:00:07:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:fe                                                   
^CTraceback (most recent call last):                                                                                                                                                                                                                                                                                
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/cli.py", line 300, in main                                                                                                                                                                                                                                 
    _print_dev_status(dev, dev.initialize(**opts))                                                                                                                                                                                                                                                                  
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 171, in initialize                                                                                                                                                                                                         
    res = self._send_set_cooling()                                                                                                                                                                                                                                                                                  
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 360, in _send_set_cooling                                                                                                                                                                                                  
    return self._send_command(_FEATURE_COOLING, _CMD_SET_COOLING, data=data)                                                                                                                                                                                                                                        
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 331, in _send_command                                                                                                                                                                                                      
    buf = bytes(self.device.read(_REPORT_LENGTH))                                                                                                                                                                                                                                                                   
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/usb.py", line 399, in read                                                                                                                                                                                                                          
    data = self.hiddev.read(length)                                                                                                                                                                                                                                                                                 
  File "hidraw.pyx", line 124, in hidraw.device.read                                                                                                                                                                                                                                                                
OSError: read error                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
During handling of the above exception, another exception occurred:                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
  File "/usr/local/bin/liquidctl", line 10, in <module>                                                                                                                                                                                                                                                             
    sys.exit(main())                                                                                                                                                                                                                                                                                                
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/cli.py", line 300, in main                                                                                                                                                                                                                                 
    _print_dev_status(dev, dev.initialize(**opts))                                                                                                                                                                                                                                                                  
KeyboardInterrupt                                                                                                                                 
```

```
root@pve:~# /usr/local/bin/liquidctl set fan speed 20 10 30  50 50 95 -g                                                                                                                                                                                                                                            
[DEBUG] liquidctl.cli: running liquidctl v1.4.2                                                                                                                                                                                                                                                                     
[DEBUG] liquidctl.driver.usb: searching HidapiBus (api=hidraw, drivers=[CorsairHidPsu, HydroPlatinum, Kraken2, KrakenX3, KrakenZ3, NzxtEPsu, RgbFusion2, SmartDevice, SmartDevice2, _CommonSmartDeviceDriver])                                                                                                      
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c20                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: instanced driver for Corsair H100i PRO XT (experimental)                                                                                                                                                                                                                              
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (drivers=[Hydro690Lc, Legacy690Lc, Modern690Lc, _CommonAsetekDriver])                                                                                                                                                                                              
[DEBUG] liquidctl.driver.usb: probing drivers for device 1b1c:0c20                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 8087:0024                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 1d6b:0002                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 8087:0024                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.driver.usb: probing drivers for device 1d6b:0002                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.cli: device: Corsair H100i PRO XT (experimental)                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.keyval: data in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0                                                                                                                                                                                                                                        
[DEBUG] liquidctl.keyval: stored fan1_mode=0 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/fan1_mode)                                                                                                                                                                                                              
[DEBUG] liquidctl.keyval: stored fan1_profile=[(20, 10), (30, 50), (50, 95)] (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/fan1_profile)                                                                                                                                                                           
[DEBUG] liquidctl.keyval: stored fan2_mode=0 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/fan2_mode)                                                                                                                                                                                                              
[DEBUG] liquidctl.keyval: stored fan2_profile=[(20, 10), (30, 50), (50, 95)] (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/fan2_profile)                                                                                                                                                                           
[INFO] liquidctl.driver.hydro_platinum: setting fan1 to follow profile [(20, 10), (30, 50), (50, 95), (60, 100), (60, 100), (60, 100), (60, 100)]                                                                                                                                                                   
[INFO] liquidctl.driver.hydro_platinum: setting fan2 to follow profile [(20, 10), (30, 50), (50, 95), (60, 100), (60, 100), (60, 100), (60, 100)]                                                                                                                                                                   
[DEBUG] liquidctl.keyval: loaded pump_mode=0 (from /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/pump_mode)                                                                                                                                                                                                            
[INFO] liquidctl.driver.hydro_platinum: setting pump mode to quiet                                                                                                                                                                                                                                                  
[DEBUG] liquidctl.keyval: loaded sequence=28 (from /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/sequence)                                                                                                                                                                                                             
[DEBUG] liquidctl.keyval: stored sequence=29 (in /run/user/0/liquidctl/vid1b1c_pid0c20/loc0/sequence)                                                                                                                                                                                                               
[DEBUG] liquidctl.driver.usb: discarded 0 previously enqueued reports                                                                                                                                                                                                                                               
[DEBUG] liquidctl.driver.usb: writting report 0x00 with 64 bytes: 3f:e8:14:00:ff:05:ff:ff:ff:ff:ff:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:07:14:1a:1e:80:32:f2:3c:ff:3c:ff:3c:ff:3c:ff:14:1a:1e:80:32:f2:3c:ff:3c:ff:3c:ff:3c:ff:00:00:00:00:00:c1                                                   
^CTraceback (most recent call last):                                                                                                                                                                                                                                                                                
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/cli.py", line 304, in main                                                                                                                                                                                                                                 
    _device_set_speed(dev, args, **opts)                                                                                                                                                                                                                                                                            
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/cli.py", line 214, in _device_set_speed                                                                                                                                                                                                                    
    dev.set_speed_profile(args['<channel>'], profile, **opts)                                                                                                                                                                                                                                                       
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 223, in set_speed_profile                                                                                                                                                                                                  
    self._send_set_cooling()                                                                                                                                                                                                                                                                                        
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 360, in _send_set_cooling                                                                                                                                                                                                  
    return self._send_command(_FEATURE_COOLING, _CMD_SET_COOLING, data=data)                                                                                                                                                                                                                                        
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/hydro_platinum.py", line 331, in _send_command                                                                                                                                                                                                      
    buf = bytes(self.device.read(_REPORT_LENGTH))                                                                                                                                                                                                                                                                   
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/driver/usb.py", line 399, in read                                                                                                                                                                                                                          
    data = self.hiddev.read(length)                                                                                                                                                                                                                                                                                 
  File "hidraw.pyx", line 124, in hidraw.device.read                                                                                                                                                                                                                                                                
OSError: read error                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
During handling of the above exception, another exception occurred:                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                    
  File "/usr/local/bin/liquidctl", line 10, in <module>                                                                                                                                                                                                                                                             
    sys.exit(main())                                                                                                                                                                                                                                                                                                
  File "/usr/local/lib/python3.7/dist-packages/liquidctl/cli.py", line 304, in main                                                                                                                                                                                                                                 
    _device_set_speed(dev, args, **opts)                                                                                                                                                                                                                                                                            
KeyboardInterrupt                                                                                                                                                           
```

MarshallAsch:
Closes #274

This should solve issues with multiple instances of liquidctl running at the same time having issues with concurrent access of the data file. 

This is being resolved by using lock files to make the functions for the `RuntimeStorage` class atomic. 

Also adds the function `load_store()` which can get a value, update it and then save it atomically.

- [x] write tests to validate that the calls are atomic
- [x] write tests to validate that multiple `load()`'s can happen at the same time
- [x] update all usages to use the atomic `load_store` function where applicable

There may be an issue on windows that if the program crashes before the lock file can be deleted it will not be able to acquire the lock again, and the program will hang.

### DIFF
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -112,7 +112,7 @@ class RuntimeStorage:
     def load(self, key, of_type=None, default=None):
         """Unstable API."""
 
-        with _shared_lock(key):
+        with self._shared_lock(key):
             value = self._backend.load(key)
 
         if value is None:
@@ -125,7 +125,7 @@ class RuntimeStorage:
     def load_store(self, key, func, of_type=None, default=None):
         """Unstable API."""
 
-        with _exclusive_lock(key):
+        with self._exclusive_lock(key):
             value = self._backend.load(key)
 
             if value is None:
@@ -138,16 +138,16 @@ class RuntimeStorage:
             newValue = func(value)
             self._backend.store(key, newValue)
 
-        return vlue
+        return value
 
     def store(self, key, value):
         """Unstable API."""
-        with _exclusive_lock(key):
+        with self._exclusive_lock(key):
             self._backend.store(key, value)
         return value
 
     @contextmanager
-    def _shared_lock(key):
+    def _shared_lock(self, key):
 
         lockFile = self._backend.lockFile(key)
 
@@ -174,7 +174,7 @@ class RuntimeStorage:
                 f.close()
 
     @contextmanager
-    def _exclusive_lock(key):
+    def _exclusive_lock(self, key):
         lockFile = self._backend.lockFile(key)
 
         if sys.platform == 'win32':
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -138,7 +138,7 @@ class RuntimeStorage:
             newValue = func(value)
             self._backend.store(key, newValue)
 
-        return value
+        return newValue
 
     def store(self, key, value):
         """Unstable API."""
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -6,9 +6,12 @@ SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
 import os
+import fcntl
 import sys
 import tempfile
+
 from ast import literal_eval
+from contextlib import contextmanager
 
 _LOGGER = logging.getLogger(__name__)
 XDG_RUNTIME_DIR = os.getenv('XDG_RUNTIME_DIR')
@@ -87,12 +90,15 @@ class _FilesystemBackend:
         data = repr(value)
         assert literal_eval(data) == value, 'encode/decode roundtrip fails'
         path = os.path.join(self._write_dir, key)
-        fd, tmp = tempfile.mkstemp(dir=self._write_dir, text=True)
         with open(fd, mode='w') as f:
             f.write(data)
             f.flush()
         os.replace(tmp, path)
         _LOGGER.debug('stored %s=%r (in %s)', key, value, path)
+
+    def lockFile(self, key):
+        return os.path.join(self._write_dir, f`${key}.lock`)
+
 
 
 class RuntimeStorage:
@@ -103,7 +109,12 @@ class RuntimeStorage:
 
     def load(self, key, of_type=None, default=None):
         """Unstable API."""
-        value = self._backend.load(key)
+
+
+        with _shared_lock(key):
+            value = self._backend.load(key)
+
+
         if value is None:
             return default
         elif of_type and not isinstance(value, of_type):
@@ -111,7 +122,79 @@ class RuntimeStorage:
         else:
             return value
 
+    def load_store(self, key, func, of_type=None, default=None):
+        """Unstable API."""
+
+        with _exclusive_lock(key):
+            value = self._backend.load(key)
+
+            if value is None:
+                value = default
+            elif of_type and not isinstance(value, of_type):
+                value = default
+            else:
+                value = value
+
+            newValue = func(value)
+            self._backend.store(key, newValue)
+
+        return vlue
+
     def store(self, key, value):
         """Unstable API."""
-        self._backend.store(key, value)
+        with _exclusive_lock(key):
+            self._backend.store(key, value)
         return value
+
+    @contextmanager
+    def _shared_lock(key):
+
+        lockFile = self._backend.lockFile(key)
+
+        if sys.platform == 'win32':
+            again = True
+            while not again:
+                try:
+                    f = open(lockFile, 'x')
+                    again = False
+                except FileExistsError:
+                    again = True
+        else:
+            f = open(lockFile, 'r')
+            fcntl.lockf(f, fcntl.LOCK_SH)
+
+        try:
+            yield
+        finally:
+            if sys.platform == 'win32':
+                f.close()
+                os.remove(f.name)
+            else:
+                fcntl.lockf(f, fcntl.LOCK_UN)
+                f.close()
+
+    @contextmanager
+    def _exclusive_lock(key):
+        lockFile = self._backend.lockFile(key)
+
+        if sys.platform == 'win32':
+            again = True
+            while not again:
+                try:
+                    f = open(lockFile, 'x')
+                    again = False
+                except FileExistsError:
+                    again = True
+        else:
+            f = open(lockFile, 'w')
+            fcntl.lockf(f, fcntl.LOCK_EX)
+
+        try:
+            yield
+        finally:
+            if sys.platform == 'win32':
+                f.close()
+                os.remove(f.name)
+            else:
+                fcntl.lockf(f, fcntl.LOCK_UN)
+                f.close()
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -6,9 +6,10 @@ SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
 import os
-import fcntl
 import sys
 import tempfile
+if sys.platform != 'win32':
+    import fcntl
 
 from ast import literal_eval
 from contextlib import contextmanager
@@ -90,6 +91,7 @@ class _FilesystemBackend:
         data = repr(value)
         assert literal_eval(data) == value, 'encode/decode roundtrip fails'
         path = os.path.join(self._write_dir, key)
+        fd, tmp = tempfile.mkstemp(dir=self._write_dir, text=True)
         with open(fd, mode='w') as f:
             f.write(data)
             f.flush()
@@ -97,23 +99,21 @@ class _FilesystemBackend:
         _LOGGER.debug('stored %s=%r (in %s)', key, value, path)
 
     def lockFile(self, key):
-        return os.path.join(self._write_dir, f`${key}.lock`)
+        return os.path.join(self._write_dir, f'${key}.lock')
 
 
 
 class RuntimeStorage:
     """Unstable API."""
 
-    def __init__(self, key_prefixes):
-        self._backend = _FilesystemBackend(key_prefixes)
+    def __init__(self, key_prefixes, backend=None):
+        self._backend = backend if backend is not None else _FilesystemBackend(key_prefixes)
 
     def load(self, key, of_type=None, default=None):
         """Unstable API."""
 
-
         with _shared_lock(key):
             value = self._backend.load(key)
-
 
         if value is None:
             return default
@@ -161,7 +161,7 @@ class RuntimeStorage:
                     again = True
         else:
             f = open(lockFile, 'r')
-            fcntl.lockf(f, fcntl.LOCK_SH)
+            fcntl.flock(f, fcntl.LOCK_SH)
 
         try:
             yield
@@ -170,7 +170,7 @@ class RuntimeStorage:
                 f.close()
                 os.remove(f.name)
             else:
-                fcntl.lockf(f, fcntl.LOCK_UN)
+                fcntl.flock(f, fcntl.LOCK_UN)
                 f.close()
 
     @contextmanager
@@ -187,7 +187,7 @@ class RuntimeStorage:
                     again = True
         else:
             f = open(lockFile, 'w')
-            fcntl.lockf(f, fcntl.LOCK_EX)
+            fcntl.flock(f, fcntl.LOCK_EX)
 
         try:
             yield
@@ -196,5 +196,5 @@ class RuntimeStorage:
                 f.close()
                 os.remove(f.name)
             else:
-                fcntl.lockf(f, fcntl.LOCK_UN)
+                fcntl.flock(f, fcntl.LOCK_UN)
                 f.close()
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -64,19 +64,22 @@ class _FilesystemBackend:
             os.chmod(self._write_dir, 0o1700)
         _LOGGER.debug('data in %s', self._write_dir)
 
-    def load(self, key):
+    def load(self, key, locked=False):
         for base in self._read_dirs:
             path = os.path.join(base, key)
             if not os.path.isfile(path):
                 continue
             try:
-                with open(path, mode='r') as f:
-                    data = f.read().strip()
-                    if len(data) == 0:
-                        value = None
-                    else:
-                        value = literal_eval(data)
-                    _LOGGER.debug('loaded %s=%r (from %s)', key, value, path)
+
+                with self._shared_lock(key, locked):
+                    with open(path, mode='r') as f:
+                        data = f.read().strip()
+
+                if len(data) == 0:
+                    value = None
+                else:
+                    value = literal_eval(data)
+                _LOGGER.debug('loaded %s=%r (from %s)', key, value, path)
             except OSError as err:
                 _LOGGER.warning('%s exists but could not be read: %s', path, err)
             except ValueError as err:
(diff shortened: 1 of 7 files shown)


tests/_testutils.py
```diff
@@ -9,28 +11,12 @@ def noop(*args, **kwargs):
...
-            return deepcopy(value)
-
-    def store(self, key, value):
-        """Unstable API."""
-        self._cache[key] = value
-        return value
+class MockRuntimeStorage(RuntimeStorage):
+    def __init__(self, key_prefixes, backend=None):
+        if not backend:
+            run_dir = mkdtemp('run_dir')
```
MarshallAsch, line 17: should this use the pytest tempdir utility?

tests/test_keyval.py
```diff
@@ -1,11 +1,75 @@
+import os
 import pytest
-
+import sys
+import time
+from multiprocessing import Process
```
MarshallAsch, line 5: Should we explicitly tell Processes to use `spawn` to start the processes instead of `fork` which according to [this](https://docs.python.org/3/library/multiprocessing.html#process-and-exceptions) might sometimes be the default.

tests/test_keyval.py
```diff
@@ -1,11 +1,75 @@
...
+
+    assert tmpstore.load('key') == '42'
+    assert tmpstore.load('key', of_type=int) is None
+
+
+def test_updates_with_load_store(tmpstore):
+    assert tmpstore.load_store('key', lambda x: x) == (None, None)
+    assert tmpstore.load_store('key', lambda x: x, default=42) == (None, 42)
+    assert tmpstore.load_store('key', lambda x: str(x)) == (42, '42')
+    assert tmpstore.load_store('key', lambda x: x, of_type=int) == ('42', None)
```
MarshallAsch, line 34: These are a good test of tests, I did not think of those extra cases.

MarshallAsch (commented):
Aside from the comments I don't immediately spot any major issues,

liquidctl/keyval.py
```diff
@@ -49,7 +56,7 @@ def _sanitize(self, key):
         return key
 
     def __init__(self, key_prefixes, runtime_dirs=get_runtime_dirs()):
-        key_prefixes = map(self._sanitize, key_prefixes)
+        key_prefixes = [self._sanitize(p) for p in key_prefixes]
```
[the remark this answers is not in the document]
jonasmalacofilho: The issue is in how the map is used bellow, in:

```py
self._read_dirs = [os.path.join(x, *key_prefixes) for x in runtime_dirs]
```

If `runtime_dirs` has more than one item, the map (as written previously) is entirely consumed when constructing the first read_dir, and later read_dirs miss the prefixes.

I caught this bug (introduced in 1.4.0) while testing the `load_store` locks with symlinks and stuff.

liquidctl/keyval.py
```diff
@@ -87,30 +97,112 @@ def store(self, key, value):
...
+                    value = literal_eval(data)
+                    _LOGGER.debug('loaded %s=%r (from %s)', key, value, read_path)
+                except OSError as err:
+                    _LOGGER.warning('%s exists but could not be read: %s', read_path, err)
+                except ValueError as err:
+                    _LOGGER.warning('%s exists but was corrupted: %s', key, err)
+                else:
+                    break
+            else:
+                _LOGGER.debug('no data (file) found for %s', key)
```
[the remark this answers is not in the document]
jonasmalacofilho: It breaks if the `try` exists without an exception.  But this isn't necessary here, so I'll move the `break` to the end of the `try` block.

liquidctl/keyval.py
```diff
@@ -87,30 +97,112 @@ def store(self, key, value):
...
+
+            new_value = func(value)
+
+            data = repr(new_value)
+            assert literal_eval(data) == new_value, 'encode/decode roundtrip fails'
```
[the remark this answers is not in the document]
jonasmalacofilho: Well, if the driver tries to store some arbitrary class instance, there's no prior assurance that its `__repr__` is complete (=contains all necessary information for reconstruction) or that it can be parsed as a Python literal.  So I though it was better to check this when saving the data, instead of later, when reading it.

Another factor was that our tests used to work with a simpler __dict__ implementation, so they wouldn't catch stuff like this because there was no actual serialization.

tests/_testutils.py
jonasmalacofilho: I would have preferred to use it, yes.  But I couldn't figure out how to use the magic `tempdir` fixture within an arbitrary class.  Do you know how it can be done?

tests/test_keyval.py
jonasmalacofilho: I think `fork` is more prone to making bugs appear, so it's probably better that it's the default on Linux.

Also, in the these tests we only open file descriptors after the `fork` (assuming the previous ones have been correctly closed), so it shouldn't impact these tests.

tests/test_keyval.py
jonasmalacofilho: Yes, I wanted to have something that pinned down the behavior around `default` and `of_type`, particularly now that this also interacts with the `load_store` closure.

If we were to document `load_store`, it would say something like: returns a tuple of the old parsed value (before applying default due to None or of_type failure) and the newly stored value (the result of applying `of_type` and `default` and passing that through the supplied closure).

liquidctl/keyval.py
```diff
@@ -87,30 +97,112 @@ def store(self, key, value):
...
+                    fcntl.flock(f, fcntl.LOCK_UN)
+
 
 class RuntimeStorage:
     """Unstable API."""
 
-    def __init__(self, key_prefixes):
-        self._backend = _FilesystemBackend(key_prefixes)
+    def __init__(self, key_prefixes, backend=None):
+        self._backend = backend if backend is not None else _FilesystemBackend(key_prefixes)
```
jonasmalacofilho, line 180: Self nitpick: this could be written in a more clear way.

Also, should be accept `runtime_dirs`  instead of requiring the tests to construct a private `_FilesystemBackend` class?  This would also avoid having the tests pass `key_prefixes` both to the backend and to the `RuntimeStorage` wrapper.

liquidctl/keyval.py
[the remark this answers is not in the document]
MarshallAsch: Ah that makes sense.

tests/_testutils.py
MarshallAsch: I suppose it could be passed in, but that is a but gross.

tests/test_keyval.py
MarshallAsch: `flock` locks are inherited when a process is forked, which might cause some tests to differ, but I suppose its only in the tests so it doesn't matter as much.

liquidctl/keyval.py
```diff
@@ -87,30 +97,112 @@ def store(self, key, value):
...
+            if sys.platform == 'win32':
+                msvcrt.locking(f.fileno(), msvcrt.LK_LOCK, 1)
+            elif shared:
+                fcntl.flock(f, fcntl.LOCK_SH)
+            else:
+                fcntl.flock(f, fcntl.LOCK_EX)
+
+            try:
+                yield f
+            finally:
```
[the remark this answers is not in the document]
MarshallAsch: I dont think our tests make sure that the lock has been released. 
I suppose the test will time out eventually if it is not released tho

liquidctl/keyval.py
MarshallAsch: imo passing a backend to the test could make sense because it would allow us to swap it out with a mock to speed up the driver tests instead of writing to the disk, (which from what my understanding was is best practice anyway?)

tests/test_keyval.py
jonasmalacofilho: With the current code we shouldn't be holding any locks when the `fork` calls are made.

MarshallAsch: good point.

liquidctl/keyval.py
jonasmalacofilho: We used to mock the backend without any actual FS access, but in this branch I actually made all tests revert back to a real FS backend, albeit with temporary and isolated locations.

Using a mocked backend requires keeping it in sync with the real implementation, something we weren't great at.  And that extra work and risk for a small performance improvement in our test suite is just not worth it, IMO.

PR #311, opened 2021-03-20
Add FreeBSD to README

historical-theology:
Add FreeBSD to README

---

Checklist:

- [ ] Verify that all automated tests pass
- [X] Update the README and other applicable documentation pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ Dependencies:
 
 liquidctl requires root privileges (doas or sudo) by default.
 
-NB: As of March 20, 2021, HIDAP (`comms/hidapi`), which is required for `comms/py-hidapi` (and, thus, liquidctl), must be built from the latest port rather than installed as a package, as the package is out of date. Make sure that you have HIDAPI version 0.10.1 or later installed prior to installing `comms/py-hidapi` and liquidctl. Then, liquidctl will work as expected.
+Note: as of March 20, 2021, HIDAPI (`comms/hidapi`), [ ... ]
 
 
 ## Installing on Windows
--- a/README.md
+++ b/README.md
@@ -186,21 +186,9 @@ Optional steps:
 
 liquidctl is maintained in the Ports Collection (thanks to ehaupt@FreeBSD.org), and it is available as a pre-built binary package.
 
-Port:
-
-`sysutils/py-liquidctl`
-
-Binary:
-
-`pkg install py37-liquidctl`
-
-Dependencies:
-
-`devel/py-docopt`
-
-`comms/py-hidapi`
-
-`devel/py-pyusb`
+- port: `sysutils/py-liquidctl`
+- binary: `pkg install py37-liquidctl`
+- dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
 
 liquidctl requires root privileges (doas or sudo) by default.
 
--- a/README.md
+++ b/README.md
@@ -55,20 +55,21 @@ NZXT Kraken X (X42, X52, X62 or X72)
 
 1.  [Supported devices](#supported-devices)
 2.  [Installing on Linux](#installing-on-linux)
-3.  [Installing on Windows](#installing-on-windows)
-4.  [Installing on macOS](#installing-on-macos)
-5.  [The command-line interface](#introducing-the-command-line-interface)
+3.  [Installing on FreeBSD](#installing-on-freebsd)
+4.  [Installing on Windows](#installing-on-windows)
+5.  [Installing on macOS](#installing-on-macos)
+6.  [The command-line interface](#introducing-the-command-line-interface)
      1. [Listing and selecting devices](#listing-and-selecting-devices)
      2. [Initializing and interacting with devices](#initializing-and-interacting-with-devices)
      3. [Supported color specification formats](#supported-color-specification-formats)
-6.  [Automation and running at boot](#automation-and-running-at-boot)
+7.  [Automation and running at boot](#automation-and-running-at-boot)
      1. [Set up Linux using systemd](#set-up-linux-using-systemd)
      2. [Set up Windows using Task Scheduler](#set-up-windows-using-task-scheduler)
      3. [Set up macOS using launchd](#set-up-macos-using-launchd)
-7.  [Troubleshooting](#troubleshooting)
-8.  [Additional documentation](#additional-documentation)
-9.  [License](#license)
-10. [Related projects](#related-projects-2020-edition)
+8.  [Troubleshooting](#troubleshooting)
+9.  [Additional documentation](#additional-documentation)
+10.  [License](#license)
+11. [Related projects](#related-projects-2020-edition)
 
 
 ## Supported devices
@@ -179,6 +180,31 @@ Optional steps:
 
 [udev rules]: extra/linux/71-liquidctl.rules
 [bash completions]: extra/completions/liquidctl.bash
+
+
+## Installing on FreeBSD
+
+liquidctl is maintained in the Ports Collection (thanks to ehaupt@FreeBSD.org), and it is available as a pre-built binary package.
+
+Port:
+
+`sysutils/py-liquidctl`
+
+Binary:
+
+`pkg install py37-liquidctl`
+
+Dependencies:
+
+`devel/py-docopt`
+
+`comms/py-hidapi`
+
+`devel/py-pyusb`
+
+liquidctl requires root privileges (doas or sudo) by default.
+
+NB: As of March 20, 2021, HIDAP (`comms/hidapi`), which is required for `comms/py-hidapi` (and, thus, liquidctl), must be built from the latest port rather than installed as a package, as the package is out of date. Make sure that you have HIDAPI version 0.10.1 or later installed prior to installing `comms/py-hidapi` and liquidctl. Then, liquidctl will work as expected.
 
 
 ## Installing on Windows
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ liquidctl is maintained in the Ports Collection (thanks to ehaupt@FreeBSD.org), 
 - binary: `pkg install py37-liquidctl`
 - dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
 
-liquidctl requires root privileges (doas or sudo) by default.
+By default, root privileges (`doas` or `sudo`) are required to run liquidctl.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,6 +182,31 @@ Optional steps:
 [bash completions]: extra/completions/liquidctl.bash
 
 
+## Installing on FreeBSD
```
jonasmalacofilho, line 185: It seems that DragonFly BSD [has a port too](https://repology.org/project/python:liquidctl/versions), with the same maintainer.

Can you check that your instructions are (at least partially) applicable to it too and, that being true, adjust this section to cover both?

README.md
```diff
@@ -181,6 +182,31 @@ Optional steps:
...
+
+`devel/py-docopt`
+
+`comms/py-hidapi`
+
+`devel/py-pyusb`
+
+liquidctl requires root privileges (doas or sudo) by default.
+
+NB: As of March 20, 2021, HIDAP (`comms/hidapi`), which is required for `comms/py-hidapi` (and, thus, liquidctl), must be built from the latest port rather than installed as a package, as the package is out of date. Make sure that you have HIDAPI version 0.10.1 or later installed prior to installing `comms/py-hidapi` and liquidctl. Then, liquidctl will work as expected.
```
jonasmalacofilho, line 207:
```suggestion
Note: as of March 20, 2021, HIDAPI (`comms/hidapi`), [ ... ]
```

README.md
```diff
@@ -181,6 +182,31 @@ Optional steps:
...
+
+`pkg install py37-liquidctl`
+
+Dependencies:
+
+`devel/py-docopt`
+
+`comms/py-hidapi`
+
+`devel/py-pyusb`
```
jonasmalacofilho, line 203:
```suggestion
- port: `sysutils/py-liquidctl`
- binary: `pkg install py37-liquidctl`
- dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
```

README.md
```diff
@@ -181,6 +182,31 @@ Optional steps:
...
+
+liquidctl requires root privileges (doas or sudo) by default.
```
jonasmalacofilho, line 205:
```suggestion
By default, root privileges (`doas` or `sudo`) are required to run liquidctl.
```

---

Also, is it possible/advised to change this default behavior?

On Linux this is done with udev rules (see `extra/linux/`).

jonasmalacofilho (commented):
Thanks!

I left some comments, can you take a look?

README.md
historical-theology: DragonFlyBSD is a separate operating system with its own rules, and I cannot verify its functionality. With that said, I will add a note that it is included in the DragonFly Ports Collection via the FreeBSD Ports Collection and should have matching dependencies. Check for that shortly.

Explanation: Since DragonFlyBSD is a (distant) fork of FreeBSD, its developers are able to run sophisticated scripts to bring the massive FreeBSD Ports Collection to their own OS. That is (almost certainly) why liquidctl is listed as having the same maintainer, even though that person might not even know that his/her work is also being used on DragonFlyBSD (all ports work is done under open BSD licenses).

README.md
```diff
@@ -181,6 +182,19 @@ Optional steps:
...
+
+liquidctl is maintained in the Ports Collection (thanks to ehaupt@FreeBSD.org), and it is available as a pre-built binary package.
+
+- port: `sysutils/py-liquidctl`
+- binary: `pkg install py37-liquidctl`
+- dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
+
+liquidctl requires root privileges (doas or sudo) by default.
+
+Note: as of March 20, 2021, HIDAPI (`comms/hidapi`), [ ... ]
```
jonasmalacofilho, line 195: Oh, sorry, `[ ... ]` wasn't a literal suggestion, it just meant "no change from this point forwards".

README.md
historical-theology: I _think_ that would best be done with devd(8), of which the GNU userland's udev utility is a (distant) imitation/clone, but I will have to spend a lot of time digging through the man pages to figure out how to get it setup properly for myself before writing a guide -- or else someone who is used to working with such things will need to add that instead.

Until then, I suppose that using `chown` for the file of the USB device is not a downright terrible option, e.g. `sudo chown [user] /dev/ugen[#.#],` as long as it is followed by a careful `chmod 644` (or stricter). But...maybe we should not formally suggest such a bizarre thing. 

I am learning.

README.md
jonasmalacofilho: Oh, i didn't know this process was (possibly) automated, and that the two can vary that substantially.

In that case, "Installing on FreeBSD" is fine.

README.md
jonasmalacofilho: It would be interesting to have this on the docs, but it doesn't need to get in right now.  Take your time...

On Linux we're currently using udev to set a `+uaccess` tag, which is pretty much equivalent to `chmod 0666`.  This is fairly common, even though I don't love it; I've recently pinged some package maintainers to discuss how to improve this.

My idea is something like setting the group to `wheel`, or some other common admin group; the issue is that Linux distros tend to have slightly different ways of handling system admin privileges, and some of them are more complicated than just a special group (and in those distros, these groups may not even exist by default).

README.md
```diff
@@ -182,17 +182,25 @@ Optional steps:
...
 - port: `sysutils/py-liquidctl`
 - binary: `pkg install py37-liquidctl`
 - dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
 
 By default, root privileges (`doas` or `sudo`) are required to run liquidctl.
 
-Note: as of March 20, 2021, HIDAPI (`comms/hidapi`), [ ... ]
+To gain full access as a normal user without `doas` or `sudo`, see devd(8) (FreeBSD-specific guide needed). Also, you might consider manually changing the permission of the file of the USB device for an individual session with `chown`, e.g. `sudo chown [user] /dev/ugen[#.#]`.
+
+Note: as of March 20, 2021, HIDAP (`comms/hidapi`), which is required for `comms/py-hidapi` (and, thus, liquidctl), must be built from the latest port rather than installed as a package, as the latest package is out of date. Make sure that you have HIDAPI version 0.10.1 or later installed prior to installing `comms/py-hidapi` and liquidctl. Then, liquidctl will work as expected.
```
jonasmalacofilho, line 199:
```
HIDAPI
     ^
```

README.md
```diff
@@ -182,17 +182,25 @@ Optional steps:
 [bash completions]: extra/completions/liquidctl.bash
 
 
-## Installing on FreeBSD
+## Installing on FreeBSD or DragonFlyBSD
```
jonasmalacofilho, line 185: There's a space in DragonFly BSD.

README.md
```diff
@@ -182,17 +182,25 @@ Optional steps:
...
+liquidctl is maintained in the FreeBSD Ports Collection (thanks to ehaupt@FreeBSD.org), and it is available as a pre-built binary package.
 
 - port: `sysutils/py-liquidctl`
 - binary: `pkg install py37-liquidctl`
 - dependencies: `devel/py-docopt`, `comms/py-hidapi`, `devel/py-pyusb`
 
 By default, root privileges (`doas` or `sudo`) are required to run liquidctl.
 
-Note: as of March 20, 2021, HIDAPI (`comms/hidapi`), [ ... ]
+To gain full access as a normal user without `doas` or `sudo`, see devd(8) (FreeBSD-specific guide needed). Also, you might consider manually changing the permission of the file of the USB device for an individual session with `chown`, e.g. `sudo chown [user] /dev/ugen[#.#]`.
```
jonasmalacofilho, line 197:
```diff
-[...] see devd(8) (FreeBSD-specific guide needed). Also, you [...]
+[...] see `devd(8)`. Alternatively, you [...]
```

README.md
```diff
@@ -182,17 +182,25 @@ Optional steps:
...
+
+Note: as of March 20, 2021, HIDAP (`comms/hidapi`), which is required for `comms/py-hidapi` (and, thus, liquidctl), must be built from the latest port rather than installed as a package, as the latest package is out of date. Make sure that you have HIDAPI version 0.10.1 or later installed prior to installing `comms/py-hidapi` and liquidctl. Then, liquidctl will work as expected.
+
+### DragonFlyBSD
```
jonasmalacofilho, line 201: There's a space in DragonFly BSD.

historical-theology:
I think that this is as far as I can take this for you folks right now. We do need to get the device rule matter sorted. I know that there is a way. I just need to find it.

Cheers.

README.md
```diff
@@ -182,17 +182,25 @@ Optional steps:
...
+
+liquidctl is available in DragonFly Ports, having been transferred from the FreeBSD Ports Collection. Installation and use should be the same as in FreeBSD (verification needed).
```
jonasmalacofilho, line 203: Do you know _for a fact_ that it was transferred from the FreeBSD collection?  Just because that can happen, doesn't necessarily mean that it happened in this case.

You went to the trouble of thanking ehaupt above... what if ehaupt really did create and test ports for both?

Then, this doesn't really help the user:

```diff
-Installation and use should be the same as in FreeBSD (verification needed).
```

I would simply remove this DragonFly subsubsection (and the heading for the FreeBSD only subsubsection) and just mention DragonFly in the "Installing ..." subsection name.

By the way, you also need to fix the table of contents.

jonasmalacofilho (commented):
A few (final) nitpicks.

Please consider that this is the main piece of documentation that most users will read, so I make an effort to make it fairly nice...

README.md
historical-theology: I am 99.9% certain, as it even says in the DragonFly BSD port file itself "$FreeBSD$", etc., etc. But I think that I found a softer way of simply mentioning the DragonFly availability.

README.md
historical-theology: Good catch. It is not always written that way, probably because everyone (myself included) is used to reading about the 3 oldest established BSDs, none of which use spaces (FreeBSD, OpenBSD, & NetBSD).

README.md
historical-theology: Fixed.

historical-theology:
> A few (final) nitpicks.
> 
> Please consider that this is the main piece of documentation that most users will read, so I make an effort to make it fairly nice...

I appreciate your attention to fine details in documentation. Your documentation is very FreeBSD-like, I dare say -- that is, comprehensive, well written, and succinct.

jonasmalacofilho:
Thanks!  For the complement and for the PR!

Issue #104, opened 2020-04-13
Add support for machine-readable JSON output


jonasmalacofilho:
A machine-readable output could be very useful for third-party software trying to fetch monitoring data from liquidctl; one example would be [Freon](https://github.com/UshakovVasilii/gnome-shell-extension-freon).

Related: [#89 (comment)](https://github.com/jonasmalacofilho/liquidctl/issues/89#issuecomment-590255300)
Related: jonasmalacofilho/liquidctl@98ef4f1bc18b ("Experiment: add support for machine-readable JSON output")
Related: jonasmalacofilho/gnome-shell-extension-freon@7cd683c31af1 ("Experiment: read sensors from liquidctl devices")

Nowaker:
Example output:

```
% liquidctl --json status
[{
	"device": "Asetek 690LC (assuming EVGA CLC)",
	"liquid_temperature": "29.5",
	"fan_speed": "960",
	"pump_speed": "2910",
	"firmware_version": "2.10.0.0"
}]
```

(Indentation added for visibility here in the GitHub issue, the output probably shouldn't be formatted as proper tools exist for that, e.g. `liquidctl --json status | jq`)

jonasmalacofilho:
I see some difficulties in implementing this, but I like the idea.¹  It resembles the hwmon attributes in sysfs.

That said, what about something more similar to the output of `sensors -j`?

---

_¹ Somewhat related to [#89 (bottom part of a later comment)](https://github.com/jonasmalacofilho/liquidctl/issues/89#issuecomment-590757983); apart from that, some reasonably stable and unambiguous device identifier should also be reported, because this is generally needed by the consumers (for example Freon needs this to know this to decide whether the sensor should be in the indicador)._

ghaabor:
I came from the OpenCorsairLink repo cause I've been using that but I see that's on the edge of being deprecated (or at least frozen). This would be great cause I want to somehow monitor my cooler (Corsair Hydro H100v2) stats under Linux and I'd like to do some visualization on them.

The main thing is that I'd give a big :+1:  for this functionality with any kind of machine readable format :)

jonasmalacofilho:
@ghaabor

You may want to take a look at `extra/liquiddump` and the discussion in #124.

I welcome feedback on how liquiddump can be improved (so it eventually makes into the liquidctl CLI), and how the `--json` output should/could look like.

MrEricM:
[...]

Issue #89, opened 2020-02-23
Encoding problem when calling exe from C# console application


flamaniac:
When calling liquidctl from C# console app i get encoding error (tried different encoding, different configuration and 3rd part lib to handle external exe call (like sheller)

CMD output:
```
>> stdio ReadToEnd: NZXT Kraken X (X42, X52, X62 or X72)
 exitCoee= 1
>> err ReadToEnd: ERROR: Unexpected error with NZXT Kraken X (X42, X52, X62 or X72)
Traceback (most recent call last):
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: character maps to <undefined>
 exitCoee= 1
```
As you see first line of output is received. I think there is problem with "├" char

My code:
```
var processStartInfo = new ProcessStartInfo();
            processStartInfo.UseShellExecute = false;
            processStartInfo.CreateNoWindow = true;
            processStartInfo.StandardErrorEncoding = Encoding.UTF8;
            processStartInfo.StandardOutputEncoding = Encoding.UTF8;
            processStartInfo.RedirectStandardOutput = true;
            processStartInfo.RedirectStandardInput = true;
            processStartInfo.RedirectStandardError = true;

            processStartInfo.FileName = "liquidctl.exe";
            processStartInfo.Arguments = "status";

            Process process = new Process();
            process.StartInfo = processStartInfo;
            process.Start();

            process.WaitForExit();

            Console.WriteLine(">> stdio ReadToEnd: " + process.StandardOutput.ReadToEnd()+  " exitCoee= "+ process.ExitCode);
            Console.WriteLine(">> err ReadToEnd: " + process.StandardError.ReadToEnd()+  " exitCoee= "+ process.ExitCode);
```

I have google "UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: " and it seems to by python related.

jonasmalacofilho:
Hi @flamaniac,

Can you add `--debug` to liquidctl's args so we can see the full exception stack?

flamaniac:
1. I have found something like this:
https://stackoverflow.com/questions/46636206/python-3-6-logging-modul-error-unicodeencodeerror-charmap-codec-cant-encod


2. i have added --debug and output looks like that:
```
>> stdio ReadToEnd: NZXT Kraken X (X42, X52, X62 or X72)
 exitCoee= 1
[...]

Issue #84, opened 2020-02-10
Characters used for the list/status "trees" are invalid with LC_ALL=C


jonasmalacofilho:
_Waiting for additional info (can't reproduce yet)._

Reported-by: Jack Zhang ([Twitter](https://twitter.com/Jack_Y_Zhang/status/1226636682033684480?s=20))

jonasmalacofilho:
Additional info:

 - reportedly only happens with `LC_ALL=C` in `/etc/environment` (but I still can't reproduce it)
 - calling liquidctl from a root shell (`sudo -s`) in Konsole

jonasmalacofilho:
```
[DEBUG] liquidctl.cli: running liquidctl v1.3.2
[DEBUG] liquidctl.driver.usb: searching GenericHidBus (api=hidraw, drivers=[CommonSmartDeviceDriver, CorsairHidPsuDriver, KrakenTwoDriver, SeasonicEDriver, SmartDeviceDriver, SmartDeviceV2Driver])
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x46d:0xc07d
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x46d:0xc07d
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1b1c:0x1b15
[DEBUG] liquidctl.driver.usb: searching PyUsbBus (drivers=[AsetekDriver, CommonAsetekDriver, CorsairAsetekDriver, LegacyAsetekDriver])
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x2433:0xb200
[DEBUG] liquidctl.driver.usb: instanced driver for Asetek 690LC (assuming EVGA CLC)
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x8087:0x24
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1d6b:0x2
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1d6b:0x3
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1d6b:0x2
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1b1c:0x1b15
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x46d:0xc07d
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x8087:0x24
[DEBUG] liquidctl.driver.usb: probing drivers for device 0x1d6b:0x2
[...]

PR #314, opened 2021-04-04
Add --json machine readable output for list/initialize/status

jonasmalacofilho:
Output machine readable JSON to stdout for `list`, `initialize` and `status`.

```
$ liquidctl list --match nzxt --json | jq
[
  {
    "description": "NZXT Smart Device (V1)",
    "vendor_id": 7793,
    "product_id": 5908,
    "release_number": 512,
    "serial_number": "497C47533332",
    "bus": "hid",
    "address": "/dev/hidraw4",
    "port": null,
    "driver": "SmartDevice",
    "experimental": false
  },
  {
    "description": "NZXT Kraken X (X42, X52, X62 or X72)",
    "vendor_id": 7793,
    "product_id": 5902,
    "release_number": 512,
    "serial_number": "49874481333",
    "bus": "hid",
    "address": "/dev/hidraw3",
    "port": null,
    "driver": "Kraken2",
    "experimental": false
  }
]
```

```
$ liquidctl --match kraken status --json | jq
[
  {
    "bus": "hid",
    "address": "/dev/hidraw3",
    "description": "NZXT Kraken X (X42, X52, X62 or X72)",
    "status": [
      {
        "key": "Liquid temperature",
        "value": 30.5,
        "unit": "°C"
      },
      {
        "key": "Fan speed",
        "value": 1014,
        "unit": "rpm"
      },
      {
        "key": "Pump speed",
        "value": 2345,
        "unit": "rpm"
      },
      {
        "key": "Firmware version",
        "value": "6.0.2",
        "unit": ""
      }
    ]
  }
]
```

Both examples have been prettified;<sup>1</sup> the actual output has no line breaks or indentation.

The output from `initialize`, as in the API, has the same structure as the one for `status`.

If `LANG=C` is set, non-ASCII characters will be escaped.  Otherwise, Unicode characters are used, allowing for zero-copy parsing of the output in Unicode-aware languages.

Closes: #104
Related: #89
Related: #84

---

Checklist (non-applicable items have been omitted):

- [x] Conform to the style guide in `docs/developer/style-guide.md`
- [x] Verify that the changes work as expected ~~on~~ with real hardware
- [x] [New CLI flag?] Adjust the completion scripts in `extra/completions/`
- [ ] ~~Update (or add) applicable `docs/*guide.md` device guides~~

Extra checks:

- [x] Test output encodings on Windows (with and without `LANG=C`)
- [x] Evaluate performance and usability of generated JSON in a few languages:<sup>2</sup>
    - [ ] ~~Python~~
    - [x] JS<sup>3</sup>: [gnome-shell-extension-freon@84a018469275 `.../liquidctlUtil.js`](https://github.com/jonasmalacofilho/gnome-shell-extension-freon/blob/84a018469275e36874e45724cd8c7a379e5b66cf/freon%40UshakovVasilii_Github.yahoo.com/liquidctlUtil.js)
    - [x] Rust: [rustlab@8c67ce72206d `reading-liquidctl-json`](https://github.com/jonasmalacofilho/rustlab/tree/8c67ce72206da0cd37c8fd4fdeb656b1d79f0bcc/reading-liquidctl-json), [`benches/json_from_liquidctl.rs`](https://github.com/jonasmalacofilho/rustlab/blob/8c67ce72206da0cd37c8fd4fdeb656b1d79f0bcc/benches/benches/json_from_liquidctl.rs)

---

<sup>1</sup> Piping to [`jq`](https://stedolan.github.io/jq/) or [`python -m json.tool`](https://docs.python.org/3/library/json.html#module-json.tool) can be used to prettify the output.
<sup>2</sup> Please leave comments with results in other languages.  
<sup>3</sup> Specifically, SpiderMonkey (used in GNOME JS).

### DIFF
--- a/extra/completions/liquidctl.bash
+++ b/extra/completions/liquidctl.bash
@@ -74,6 +74,7 @@ _liquidctl_main() {
     local boolean_options="
     --verbose -v
     --debug -g
+    --json
     --version
     --help
     --single-12v-ocp
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -214,8 +214,7 @@ def _list_devices_human(devices, *, using_filters, device_id, verbose, debug, **
 
 
 def _dev_status_obj(dev, status):
-    # unlike the humans, machines want to know everything; don't suppress
-    # devices without status data (typically from initialize)
+    # don't suppress devices without status data (typically from initialize)
     if not status:
         status = []
 
@@ -338,6 +337,12 @@ def main():
     logging.basicConfig(level=log_level, handlers=[log_handler])
 
     _LOGGER.debug('running %s', _gen_version())
+
+    # unlike humans, machines want to know everything; imply verbose everywhere
+    # other than when setting default logging level and format (which are
+    # inherently for human consumption)
+    if args['--json']:
+        args['--verbose'] = True
 
     opts = _make_opts(args)
     filter_count = sum(1 for opt in opts if opt in _FILTER_OPTIONS)
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -42,8 +42,7 @@ Other device options:
 Other interface options:
   -v, --verbose                  Output additional information
   -g, --debug                    Show debug information on stderr
-  --json                         Output machine readable JSON
-                                 (list/initialization/status)
+  --json                         JSON output (list/initialization/status)
   --version                      Display the version number
   --help                         Show this message
 
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -42,6 +42,8 @@ Other device options:
 Other interface options:
   -v, --verbose                  Output additional information
   -g, --debug                    Show debug information on stderr
+  --json                         Output machine readable JSON
+                                 (list/initialization/status)
   --version                      Display the version number
   --help                         Show this message
 
@@ -66,9 +68,11 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 import datetime
 import errno
 import inspect
+import json
 import logging
 import os
 import sys
+from numbers import Number
 from traceback import format_exception
 
 import colorlog
@@ -140,7 +144,26 @@ _VALUE_FORMATS = {
 _LOGGER = logging.getLogger(__name__)
 
 
-def _list_devices(devices, using_filters=False, device_id=None, verbose=False, debug=False, **opts):
+def _list_devices_objs(devices):
+    return [
+        {
+            # replace the experimental suffix with a proper field
+            'description': dev.description.replace(' (experimental)', ''),
+            'vendor_id': dev.vendor_id,
+            'product_id': dev.product_id,
+            'release_number': dev.release_number,
+            'serial_number': getattr(dev, 'serial_number', None),
+            'bus': dev.bus,
+            'address': dev.address,
+            'port': dev.port,
+            'driver': type(dev).__name__,
+            'experimental': dev.description.endswith('(experimental)'),
+        }
+        for dev in devices
+    ]
+
+
+def _list_devices_human(devices, *, using_filters, device_id, verbose, debug, **opts):
     for i, dev in enumerate(devices):
         warnings = []
 
@@ -189,6 +212,30 @@ def _list_devices(devices, using_filters=False, device_id=None, verbose=False, d
         print('')
 
     assert 'device' not in opts or len(devices) <= 1, 'too many results listed with --device'
+
+
+def _dev_status_obj(dev, status):
+    # unlike the humans, machines want to know everything; don't suppress
+    # devices without status data (typically from initialize)
+    if not status:
+        status = []
+
+    return {
+        'bus': dev.bus,
+        'address': dev.address,
+        # suppress the experimental suffix (check .experimental from list)
+        'description': dev.description.replace(' (experimental)', ''),
+        # JSON values can only be objects, arrays, numbers, strings, or
+        # false/null/true; convert all non-numeric status values to strings
+        'status': [
+            [
+                key,
+                val if isinstance(val, Number) else str(val),
+                unit
+            ]
+            for key, val, unit in status
+        ],
+    }
 
 
 def _print_dev_status(dev, status):
@@ -316,7 +363,12 @@ def main():
         selected = [compat[device_id]]
 
     if args['list']:
-        _list_devices(selected, using_filters=bool(filter_count), device_id=device_id, **opts)
+        if args['--json']:
+            objs = _list_devices_objs(selected)
+            json.dump(objs, sys.stdout, ensure_ascii=(os.getenv('LANG', None) == 'C'))
+        else:
+            _list_devices_human(selected, using_filters=bool(filter_count),
+                                device_id=device_id, json=json, **opts)
         return
 
     if len(selected) > 1 and not (args['status'] or args['all']):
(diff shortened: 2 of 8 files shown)


liquidctl/cli.py
```diff
@@ -191,6 +213,30 @@ def _list_devices(devices, using_filters=False, device_id=None, verbose=False, d
     assert 'device' not in opts or len(devices) <= 1, 'too many results listed with --device'
 
 
+def _dev_status_obj(dev, status):
+    # unlike the humans, machines want to know everything; don't suppress
+    # devices without status data (typically from initialize)
```
MarshallAsch, line 218: Does this comment also mean that the json output should be equivalent to the verbose flag?

liquidctl/cli.py
```diff
@@ -191,6 +213,30 @@ def _list_devices(devices, using_filters=False, device_id=None, verbose=False, d
...
+        # suppress the experimental suffix (check .experimental from list)
+        'description': dev.description.replace(' (experimental)', ''),
+        # JSON values can only be objects, arrays, numbers, strings, or
+        # false/null/true; convert all non-numeric status values to strings
+        'status': [
+            [
+                key,
+                val if isinstance(val, Number) else str(val),
+                unit
+            ]
```
MarshallAsch, line 234: I for-see some issues with this, specificly using a list of (key, value, units) because some statuses do not have units. 
ie. my graphics card:
`{"bus": "i2c-4", "address": "0x2a", "description": "ASUS Strix RTX 2080 Ti OC", "status": [["Mode", "Off", ""]]}` 

And the temperature units are written in `°C` by leaving it like this it could lead to some issues if units change or parsing the non-ascii character `°`.

Imo this will lead to the encouragement of status parsing based off of array indexes rather than by status message type (for lack of a better phrase). 

I know that changing this would also require changing all of the device drivers and it may not be work it for an initial version of this. It might be worth changing unit to a type and then using that type to map to a string for the unit. 

type = 'temp' => `°C` or `°F` (configurable later?)
type = 'duty' => `%` 
type = 'speed' => `rpm`
type = 'color' => `''`
type = 'light_mode' => `''`

and then the json list would be a list of:
```json
{
    "label" : key,
    "value": val,
    "type": type,
    "unit": unit
}
```

liquidctl/cli.py
jonasmalacofilho: I considered that, and opted to *not* imply `--verbose` in `status`/`initialize` since that enables output from more _devices._  But now I'm thinking that was probably a mistake.

liquidctl/cli.py
jonasmalacofilho: > And the temperature units are written in °C by leaving it like this it could lead to some issues if units change [...]

The units are as stable as the values or keys themselves.  That is, considerably stable, although it's something we can occasionally change if there's a compelling reason to do so.

> [...] or parsing the non-ascii character °.

That's a very fringe case, but nonetheless, it's supported: `LANG=C` will escape any non-ASCII characters.

Additionally, supporting JSON implies supporting Unicode:

> All Unicode characters may be placed within the quotation marks, except for the characters that must be escaped [...]
> —_[RFC 7159](https://tools.ietf.org/html/rfc7159.html#section-7)

> Imo this will lead to the encouragement of status parsing based off of array indexes rather than by status message type (for lack of a better phrase).

Unfortunately this already happens with the API, but it's caused by the external list, not by the tuple items themselves.

I think we can/should offer better documentation, with examples of recommended use.

Otherwise, the only way to have this not happen is to return something that cannot be indexed, like an object (of status items).  But that's silly, inefficient, and also raises another issue: it could be interpreted to mean that fields that exist today will always be there.  Yes, we could document that's not the case, but at that point we can just document good ways to use the list.

> It might be worth changing unit to a type and then using that type to map to a string for the unit.

This is related to some statues not having units: we simply don't know what type they are right now, and even with a new `get_status_v2` API, it would still be pretty hard to enumerate all types of statuses in a way that's useful to everyone.

Additionally, the units are, IMO, always necessary, and they already convey the type of the data items that are more important and/or more likely to be used.  For generic processing, I think that temperatures and speeds (in rpm) are the main sought-after data, and those are trivial to filter using the units.

Other use cases would already need to assume a very specific device (e.g. your graphics card, that reports the lighting mode), and so could simply match the key/label.

> and then the json list would be a list of:

Using objects for each status item would be terribly inefficient, both in space, as well as during encoding and decoding.  

Given the uniform structure of every status item (even if, admittedly, it's not very detailed), I really think that a tuple, or the closest thing to it in JSON, are more suitable.

But I'll see if I can come up with something better, as well as do some proper benchmarks.

liquidctl/cli.py
MarshallAsch: The current implementation does enable output from all devices. If i were to run `liquidctl status --json` I get output that includes my GPU, but the gpu is not listed from a normal `liquidctl status`. Based on your code that _seems_ intentional. imo, `--json` should imply verbose if the current implementation has all devices included in the list with no status output.

liquidctl/cli.py
MarshallAsch: Your response addresses all of the concerns that I had, and I did not think of the encoding / decoding offences. Im going change my stance to this is all good and mark this as resolved.

liquidctl/cli.py
jonasmalacofilho: Hm, it does, but the actual status are still suppressed at the driver level, which I (now) think is wrong.

Of course, in the case of I²C/SMBus devices `--unsafe <...>` will also be necessary to avoid the values from being suppressed.  But that's reasonable, I think.

MarshallAsch: > Of course, in the case of I²C/SMBus devices --unsafe <...> will also be necessary to avoid the values from being suppressed. But that's reasonable, I think.

I agree with that part

liquidctl/cli.py
```diff
@@ -191,6 +213,29 @@ def _list_devices(devices, using_filters=False, device_id=None, verbose=False, d
...
+    if not status:
+        status = []
+
+    return {
+        'bus': dev.bus,
+        'address': dev.address,
+        # suppress the experimental suffix (check .experimental from list)
+        'description': dev.description.replace(' (experimental)', ''),
+        # JSON values can only be objects, arrays, numbers, strings, or
+        # false/null/true; convert all non-numeric status values to strings
```
jonasmalacofilho, line 227: This is too extreme, or at least shouldn't be relied on by consumers of this JSON.

In the future, arrays, tuples and other types that can be represented in JSON should probably be kept.  Although, IIRC, we don't currently have any of those; the closest are time spans, which could perhaps be serialized as spans of [milli]seconds, instead of a ISO-like time string.

I need to think about this a bit more... @MarshallAsch, what do  you think?

sgehrman:
merge it.  :)

extra/completions/liquidctl.bash
```diff
@@ -74,6 +74,7 @@ _liquidctl_main() {
     local boolean_options="
     --verbose -v
     --debug -g
+    --json
```
MarshallAsch, line 77: yay this was remembered :)

liquidctl/cli.py
MarshallAsch: Personally I think having the timespans as milliseconds would be a lot better than  using [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) time intervals because those look nasty imo. And if interval strings were used then they would have be parsed and generated and thats no fun.

liquidctl/cli.py
```diff
@@ -352,6 +419,11 @@ def log_error(err, msg, append_err=False, *args):
     if errors:
         sys.exit(errors)
 
+    if args['--json']:
+        print(json.dumps(obj_buf, ensure_ascii=(os.getenv('LANG', None) == 'C')))
```
MarshallAsch, line 423: Can we add a `--beautify` flag to for the json output and if it is set  then include `indent=4` as one of the parameters? For the sake of crazy people like me who sometimes want to look at the json output?

MarshallAsch (approved):
Looks good, it works, output is now verbose which is nice. and when `LANG="C"` the temperature unit looks like `"\u00b0C"` so that is good

liquidctl/cli.py
jonasmalacofilho: Do you think piping to `python -m json.tool` (included in Python) or `jq` would be sufficient?  As usual, I would prefer to *not* add more options.

extra/completions/liquidctl.bash
jonasmalacofilho: Only thanks to the checklist : )

liquidctl/cli.py
MarshallAsch: oh ya that is definitely sufficient, I did not know `python -m json.tool` was a thing (it might be work mentioning that in the docs)

PR #318, opened 2021-04-04
317 log colors

### LINKED ISSUE #317: colorize log output
Pretty colors are nice, who does not like them? answer: nobody, everyone loves colors. 

Having colorized log output for warnings, errors, and debug messages will make it easier to quickly spot important warnings and other messages. 

I have found the module [colorlog](https://pypi.org/project/colorlog/) that seems to work for this task. 
I am not sure if it should be used as an optional dependancy, but one that is installed by default as a dependancy, or if it should be required.

MarshallAsch:
Closes #317 

Logging now prints with pretty colors yay. This adds an optional module dependancy of  [colorlog](https://pypi.org/project/colorlog/).

The log format is the same, except the log level is now printed in a color.
---

Checklist:

### DIFF
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -71,12 +71,8 @@ import os
 import sys
 from traceback import format_exception
 
+import colorlog
 from docopt import docopt
-try:
-    import colorlog
-    use_color_log=True
-except ImportError:
-    use_color_log=False
 
 from liquidctl.driver import *
 from liquidctl.error import NotSupportedByDevice, NotSupportedByDriver, UnsafeFeaturesNotEnabled
@@ -269,26 +265,20 @@ def main():
 
     if args['--debug']:
         args['--verbose'] = True
-        format_color='%(log_color)s[%(levelname)s]%(name)s%(reset)s: %(message)s'
-        format_basic='[%(levelname)s] %(name)s: %(message)s'
-        level=logging.DEBUG
+        format_color = '%(log_color)s[%(levelname)s] %(name)s%(reset)s: %(message)s'
+        level = logging.DEBUG
     elif args['--verbose']:
-        format_color='%(log_color)s%(levelname)s%(reset)s: %(message)s'
-        format_basic='%(levelname)s: %(message)s'  
-        level=logging.INFO
+        format_color = '%(log_color)s%(levelname)s%(reset)s: %(message)s'
+        level = logging.INFO
     else:
-        format_color='%(log_color)s%(levelname)s%(reset)s: %(message)s'
-        format_basic='%(levelname)s: %(message)s'
-        level=logging.WARNING
+        format_color = '%(log_color)s%(levelname)s%(reset)s: %(message)s'
+        level = logging.WARNING
         sys.tracebacklimit = 0
-   
-    if use_color_log:
-        formatter = colorlog.ColoredFormatter(format_color)
-        handler = logging.StreamHandler()
-        handler.setFormatter(formatter)
-        logging.basicConfig(level=level, handlers=[handler])
-    else:
-        logging.basicConfig(level=level, format=format_basic)  
+
+    formatter = colorlog.TTYColoredFormatter(fmt=format_color, stream=sys.stdout)
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+    logging.basicConfig(level=level, handlers=[handler])
 
     _LOGGER.debug('running %s', _gen_version())
 
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -72,6 +72,11 @@ import sys
 from traceback import format_exception
 
 from docopt import docopt
+try:
+    import colorlog
+    use_color_log=True
+except ImportError:
+    use_color_log=False
 
 from liquidctl.driver import *
 from liquidctl.error import NotSupportedByDevice, NotSupportedByDriver, UnsafeFeaturesNotEnabled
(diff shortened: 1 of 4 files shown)


jonasmalacofilho:
The new dependency (`colorlog`) is ok:

- [x] Limited/reasonable transitive dependencies? (only `colorama`, exclusively on Windows)
- [x] Supports Linux?
    - [x] Available on ArchLinux? (`python-colorlog`)
    - [x] Available on Fedora? (`python3-colorlog`)
    - [x] Available on Debian? (`python3-colorlog`)
    - [x] Can be expected to be packaged in other distros?
- [x] Support Windows? (with `colorama`)
    - [x] Easily installed (pure Python or has binary wheels)? (yes, including `colorama`)
- [x] Supports Mac OS? (I don't see why not)
- [x] Handles redirection/strips ANSI escape codes when the output is not a terminal? (supported, opt-in and not currently enabled)

.gitignore
```diff
@@ -5,3 +5,4 @@ build
 dist
 liquidctl/extraversion.py
 .coverage
+env/**
```
jonasmalacofilho, line 8: This PR is probably not the right place for this.

Additionally, since the venv directory name can be different (for example, I'm more used to seeing `venv`), depending on the developer, it's something more suitable for each developer's individual global gitignore file ([`core.excludesFile`](https://git-scm.com/docs/gitignore)).

If you don't mind, can you drop the commit that added this?

liquidctl/cli.py
```diff
@@ -72,6 +72,11 @@
 from traceback import format_exception
 
 from docopt import docopt
+try:
+    import colorlog
+    use_color_log=True
+except ImportError:
+    use_color_log=False
```
jonasmalacofilho, line 79: I think it can be a normal/required dependency.

liquidctl/cli.py
```diff
@@ -264,13 +269,28 @@ def main():
...
+        level=logging.INFO
     else:
-        logging.basicConfig(level=logging.WARNING, format='%(levelname)s: %(message)s')
+        format_color='%(log_color)s%(levelname)s%(reset)s: %(message)s'
+        format_basic='%(levelname)s: %(message)s'
+        level=logging.WARNING
         sys.tracebacklimit = 0
+   
+    if use_color_log:
+        formatter = colorlog.ColoredFormatter(format_color)
```
jonasmalacofilho, line 286: I think this needs to use [`TTYColoredFormatter`](https://github.com/borntyping/python-colorlog/blob/1297254a29a4d6a68cc5e0a477a075f4421dd285/colorlog/colorlog.py#L224) instead:

> Blanks all color codes if not running under a TTY.
> This is useful when you want to be able to pipe colorlog output to a file.

This is probably enough, although I don't honestly remember exactly how this stuff works.  But, essentially, we don't want ANSI escape codes if the output is redirected to a file or piped to another program.

liquidctl/cli.py
```diff
@@ -264,13 +269,28 @@ def main():
...
+        handler = logging.StreamHandler()
+        handler.setFormatter(formatter)
+        logging.basicConfig(level=level, handlers=[handler])
+    else:
+        logging.basicConfig(level=level, format=format_basic)
```
jonasmalacofilho, line 291: As this branch isn't necessary once we make colorlog required, the code can be simplified a bit.

liquidctl/cli.py
```diff
@@ -264,13 +269,28 @@ def main():
 
     if args['--debug']:
         args['--verbose'] = True
-        logging.basicConfig(level=logging.DEBUG, format='[%(levelname)s] %(name)s: %(message)s')
-        _LOGGER.debug('running %s', _gen_version())
+        format_color='%(log_color)s[%(levelname)s]%(name)s%(reset)s: %(message)s'
```
jonasmalacofilho, line 272: I think you may have accidentally consumed a space:
```diff
-[%(levelname)s] %(name)
+[%(levelname)s]%(name)
```

I think the former looks better, but let me know if you disagree and this was intentional.

jonasmalacofilho (commented):
I agree, this is useful.

We should probably also revisit the layout of the main output, and add some color to it too.  But let's leave that for later... although, preferably, before these colored logs reach a release, because I would prefer to not add a dependency in one version just to replace it on the next.

Besides the comments left inline, I also noticed that there's at least one error path in `liquidctl.cli` that doesn't currently use `_LOGGER.error`:

```
$ python -m liquidctl.cli -m sadasd status
Error: no devices matches available drivers and selection criteria
^^^^^
  not colored
```

jonasmalacofilho:
Two more comments:

- the README also needs to be updated to list the new dependency;
- nit: I would really prefer to have spaces around the assignment operator.<sup>1, 2</sup>

_<sup>1</sup> Do you know in which languages, other than sh/bash, are those spaces normally omitted?  
_<sup>2</sup> Well, PEP-8 would also agree... but, regardless, it's a particular pet-peeve of mine.  Hopefully, you can indulge me on this one?

.gitignore
MarshallAsch: yup, my bad. I just dumped it there not really thinking about it. I can fix that in the morning

liquidctl/cli.py
MarshallAsch: okay, I wasn't sure how you wanted to handle it I remember seeing (or I thought I saw) something elsewhere that made use of optional dependancies for non core libraries.

jonasmalacofilho: Hm, now that you mention it, do we currently have an optional dependency?

I only though about `smbus`: it was optional for a while, but was made into a conditional (Linux-only) dependency for 1.5.0.  But that was a core dependency, so you were probably thinking of something else... that I could be missing.

liquidctl/cli.py
MarshallAsch: that did not occur to me before when I implemented this. (I also did not look through the code, I assumed that they would have documented the things in the readme). Ill check this and update it. 

I will also test it when not running under a TTY, Im not quite sure how to test that but Ill see what I can come up with.

liquidctl/cli.py
MarshallAsch: yup it looks like I did, I like the space more too, and I messed that up a few times during my initial testing with copy paste errors. Ill fix that though

MarshallAsch:
> nit: I would really prefer to have spaces around the assignment operator

Yup me too, I forgot about the linter to check that :(

On a related note, why didnt the linter in the pipeline catch that?

> 3 Named function arguments are not assignments,...

I did know know that the space can not be there, it did not occur to me _why_ there are never spaces in the function arguments. 


> Do you know in which languages....

I cant think of any that _require_ there to be spaces or require there to not be any, bash is the only one I can think of and I always mess that one up the first time I test my scripts. I had thought I remembered either ADA or Fortran requiring a single spaces but I double checked that and it isn't important.

liquidctl/cli.py
MarshallAsch: I think I may just have been thinking of of the fallback usb dependancies here:
https://github.com/liquidctl/liquidctl/blob/8ef2e79fd5f385487ff4a1a8e4ee92f0670a6d7f/liquidctl/driver/usb.py#L62-L68

jonasmalacofilho:
> On a related note, why didnt the linter in the pipeline catch that?

It's configured to only issue warnings on style checks, and even then, they aren't reported as part of the overall commit/PR status.
We could fix that, but it would require filtering only the warnings that are sensible.

Another, probably better, alternative would be to find an automatic formatter that produces reasonable output _for our code._  I'm not that attached to my own style of formatting Python, as long as the alternative is not unreadable.  And then we could have the same tool run in "check" mode in the pipeline.

PR #319, opened 2021-04-04
195 config file

MarshallAsch:
Closes: #195 

This is my work in progress to add a config file to liquidctl so that some of the cli flags dont need to be given to do the "normal" setup of the device. 

TODO tasks:
- [x] define fallback config file locations
- [x] validate 'basic' structure of the config files (required and mutual excludes sections are present, some general types are correct)
- [ ] add logging to print which files are being checked / loaded
- [ ] implement config value overrides using cli flags
- [ ] pass config file values to the device drivers
- [ ] implement 'validation only' driver calls that do no setting changes on the devices
- [x] create tests for config file locations
- [ ] create tests for 'basic' file validation checks, (`/dev/null`, missing required fields, etc.)
- [ ] create checks that the validation mode does not send commands that change device settings
- [ ] create documentation about how to create and use config files and how they work
- [ ] update man pages with the config file information
---

Checklist:

- [ ] Verify that all automated tests pass


liquidctl/config/load.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # treat all other platforms as *nix and conform to XDG basedir spec
+        if XDG_CONFIG_HOME:
+            files.append(os.path.join(XDG_CONFIG_HOME, appname, 'config.toml'))
+
+        files.append(os.path.expanduser(os.path.join('~', '.config', appname, 'config.toml')))
+        files.append(os.path.expanduser(os.path.join('~', f'.{appname}.toml')))
+
+        if XDG_CONFIG_DIRS:
+            files.append(os.path.join(XDG_CONFIG_DIRS, appname, 'config.toml'))
```
jonasmalacofilho, line 33: XDG_CONFIG_DIRS should come first, and if it's not present, it should default to `/etc`.

liquidctl/config/load.py
```diff
@@ -0,0 +1,53 @@
...
+    elif sys.platform == 'linux':
+        XDG_CONFIG_HOME = os.getenv('XDG_CONFIG_HOME')
+        XDG_CONFIG_DIRS = os.getenv('XDG_CONFIG_DIRS')
+
+        # treat all other platforms as *nix and conform to XDG basedir spec
+        if XDG_CONFIG_HOME:
+            files.append(os.path.join(XDG_CONFIG_HOME, appname, 'config.toml'))
+
+        files.append(os.path.expanduser(os.path.join('~', '.config', appname, 'config.toml')))
+        files.append(os.path.expanduser(os.path.join('~', f'.{appname}.toml')))
```
jonasmalacofilho, line 30: These should not override a set XDG_CONFIG_HOME.  I think the latter is also use to _change_ the effective config dir for a particular program.

liquidctl/config/load.py
jonasmalacofilho, line 33: XDG_CONFIG_DIRS should come first, and if it's not present, it should default to `/etc`.

liquidctl/config/load.py
```diff
@@ -0,0 +1,53 @@
...
+    can be
+
+    This will also handle the -f overrides
+    """
+    files = []
+    if file:
+        files.append(file)
+
+    if sys.platform == 'win32':
+        files.append(os.path.join(os.getenv('APPDATA'), appname, 'config.toml'))
```
jonasmalacofilho, line 17: May need to consider system-wide settings as well.

liquidctl/config/load.py
```diff
@@ -0,0 +1,53 @@
...
+    elif sys.platform == 'darwin':
+        files.append(os.path.expanduser(os.path.join('~/Library/Application Support', appname, 'config.toml')))
```
jonasmalacofilho, line 19: Needs to consider system-wide settings as well.

liquidctl/config/validation.py
```diff
@@ -0,0 +1,157 @@
+import logging
```
jonasmalacofilho, line 1: General comment: this looks hard to maintain in the long run.

Have you consider a generic solutions, perhaps relying on each driver exposing some additional information?

setup.py
```diff
@@ -70,7 +70,7 @@ def run(self):
 
 make_extraversion()
 
-install_requires = ['docopt', 'pyusb', 'hidapi']
+install_requires = ['docopt', 'pyusb', 'hidapi', 'tomlkit']
```
jonasmalacofilho, line 73: Can you see if this dependency passes all checks I mentioned for `colorlog` on #318?

Also, there seem to be three common library choices for Python (`tomlkit`, `pytomlpp`, and `toml`), so I'm curious of why/how we picked this one.

liquidctl/cli.py
```diff
@@ -45,6 +46,10 @@
   --version                      Display the version number
   --help                         Show this message
 
+Config file options:
+  -f, --file <file>              Select a specific config file to use
+  --ignore-config                Do not use any of the config files
```
jonasmalacofilho, line 51: `--ignore-config` is probably unnecessary, `-f /dev/null` or `-f <some empty file>` are already sufficient for this very niche use case.

jonasmalacofilho, line 51: `--ignore-config` is probably unnecessary, `--file /dev/null` or `--file <some empty file>` are already sufficient for this very niche use case.

liquidctl/cli.py
```diff
@@ -9,6 +9,7 @@
   liquidctl [options] set <channel> color <mode> [<color>] ...
   liquidctl --help
   liquidctl --version
+  liquidctl [-f] --validate
```
jonasmalacofilho, line 12: Does this work?  I think it would need to be `liquidctl [-f <file>] --validate` for docopt to parse it correctly.  Additionally, `--validate` needs to be listed as an option, or be a command.

Finally, at this point, it may be unclear to the user what `--validate` validates... maybe `--validate-config` or `--check-config`?  (This part may no longer apply if `--file` is renamed to `--config`).

liquidctl/cli.py
```diff
@@ -45,6 +46,10 @@
   --version                      Display the version number
   --help                         Show this message
 
+Config file options:
+  -f, --file <file>              Select a specific config file to use
```
jonasmalacofilho, line 50: Let's save `-f` for something else, there's no need to spend a one-line option here.

I also don't think `--file` is super clear (anywhere else but this specific place)... what do you think about `--config`?

liquidctl/config/load.py
MarshallAsch: did I mix up the orders of these? isnt `XDG_CONFIG_DIRS=/etc` and `XDG_CONFIG_HOME=~/.config`? if so that is the order that we want, the specific user folder is first and then it should fall back to the system specific ones. 

Ill re read the XDG spec again over the weekend, maybe I read it wrong.

liquidctl/config/load.py
MarshallAsch: ya these are both fallbacks from the `XDG_CONFIG_HOME` folder, I think I may be misunderstanding what you are saying though

liquidctl/config/load.py
MarshallAsch: This is the application configure file location that apple suggests using. I can add an `/etc/liquidctl` dir here as well though

liquidctl/config/load.py
MarshallAsch: I'm not familiar enough with windows to know what those folders would be, this is what cameup when I was googling where to put application configs on windows

liquidctl/cli.py
MarshallAsch: > Does this work? I think it would need to be liquidctl [-f <file>] --validate for docopt to parse it correctly. Additionally, --validate needs to be listed as an option, or be a command.

I did test this and I thought it worked as expected, but Ill check again. 


> Finally, at this point, it may be unclear to the user what --validate validates..

I know theat nginx using a `-t` flag to do that validation and they call it test. perhaps that would be better. And I do like the `--config` flag instead of `--file` but I also like short flags more than long names and I dont think that `-c` would make sense.

liquidctl/cli.py
MarshallAsch: I like that and will update in the next commit

liquidctl/cli.py
MarshallAsch: ohhh I like the `/dev/null` idea, but Ill need to make sure that the parser actually handles that

setup.py
MarshallAsch: Yup Ill go through that checklist that you made. 

I picked this one because it supported a newer version of the toml specification and I got it from the list of libraries that are listed on the toml site. I also looked at `toml` but not `pytomlpp` Ill check that out to see which might be better

liquidctl/config/load.py
jonasmalacofilho: I think that, if XDG_CONFIG_HOME is set and non-empty, then `~/liquidctl/config.toml` and `~/config.toml` should *not* be checked.  XDG_CONFIG_HOME should completely override the _location_ of the user's settings, even if no config is actually found in the XDG location.

liquidctl/config/load.py
jonasmalacofilho: You're right, my bad.

liquidctl/config/validation.py
MarshallAsch: I too am not super happy with how I am approaching this, it feels horrendous. But I also think this general structure would not be that hard to maintain because each function is responsible for checking a single block and then it delegates the sub block to a different function. 

But Ill look  into other config file parsers to see how they handle similar cases

liquidctl/config/load.py
jonasmalacofilho: There's `/Library` as well:

> `/Library`(macOS only)—Apps that share resources between users store those resources in this version of the Library directory
> —_[The Library Directory Stores App-Specific Files](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1)

But perhaps it's better to store this Unix-style config file in a Unix-style path... we may need to consult some Mac OS people on this one.

@CaseySJ, do you have any idea on what we should do here?

liquidctl/cli.py
jonasmalacofilho: The parser would only need to handle an empty configuration file... the kernel should handle the rest for us : )

liquidctl/config/load.py
```diff
@@ -0,0 +1,53 @@
...
+    else:
+        files.append(os.path.expanduser(os.path.join('~', f'.{appname}.toml')))
+
+    return files
+
+def filter_config_files(files=[], ignore_config=False, **kwargs):
+    """
+    This will take a list of file paths and return the first one that exists from the list.
+    If the ignore config option is set then it will return None even if files do exist.
+    """
```
jonasmalacofilho, line 43: Alternatively, we might want to merge the sections from the various files that we found, allowing the user config file to override _sections_ of the global/system config.

_(I think this is what I was thinking of when I messed up above)._

liquidctl/config/validation.py
jonasmalacofilho: I don't think the parsing is the issue, but rather how we map this to the actual devices and their capabilities.  That's the part that I'm worried might be hard to maintain.

setup.py
jonasmalacofilho: Yes, I noticed that `toml` is not listed as supporting TOML 1, but I wasn't able to verify that in their repository... maybe they've upgraded since?

By the way `pytomlpp` should (already or soon) support TOML 1.0.0, since it claims to be a wrapper over the C++ `tomlplusplus`, which nowadays supports 1.0.0.

liquidctl/config/load.py
CaseySJ: > @CaseySJ, do you have any idea on what we should do here?

You're right. On macOS there are two places where applications can store any app-specific files:

- For a particular user: `~/Library/Application Support/<app name>/file(s)`
- For all users: `/Library/Application Support/<app name>/file(s)`

jonasmalacofilho: Thanks @CaseySJ !

MarshallAsch: > or all users:.... 

I will add that to the mac location checking then.

MarshallAsch: > Alternatively, we might want to merge the sections from the various files that we found....

Im not sure how I feel about that, it makes sense but at the same time I have recently been working with spring boot and their config files work like that where they can override specific sections. Thats has been suuuuuuper confusing to work with to figure out what config file is causing things to go wrong because you might forget about what is in which file or where all the different locations are.

liquidctl/config/load.py
MarshallAsch: > `~/liquidctl/config.toml` and `~/config.toml`

Neither of those are checked it is. `~/.config/liquidctl/config.toml` and `~/.liquidctl.toml`


> I think that, if `XDG_CONFIG_HOME` is set and non-empty, then .... should not be checked 

I disagree with that, `XDG_CONFIG_HOME` would be a variable set that would apply to all applications and I may have some files there for other applications but maybe not this one. I think that the fallback locations should still be checked.

jonasmalacofilho: I messed up the names... I meant the correct ones.

---

I would like to make two arguments against always checking `~/.config` and ~/.liquidctl.toml`:

> There is *a single* base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable `$XDG_CONFIG_HOME`. 
> [ ... ]
> `$XDG_CONFIG_HOME` *defines the* base directory relative to which user specific configuration files should be stored. *If* `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.
> —_XDG Base Directory Specification,_ emphasis mine.

First, my interpretation is that the spec clearly states that:

- `$XDG_CONFIG_HOME` is the only place where user specific files should be stored; this is also in-line with what, IMO, is the spirit of that spec: stop the madness of having config files everywhere in the user's home directory.
- `$HOME/.config/` should only be used as a default for `$XDG_CONFIG_HOME`, if the latter is not set or empty.

Second, `$XDG_CONFIG_HOME` can also be used to override the location of the user configuration files for a specific program, and this seems to like a perfectly valid (if not very common) use case.

If a user chooses to run `XDG_CONFIG_HOME=/some_other_path liquidctl`, they could very well mean that they only want that location to be searched (for the user specific files).  For example, to test how liquidctl behaves with a different configuration than what they normally use, or with no configuration at all.  Sure, there are other ways of doing those things, but my point here is simply that it's _valid_ for a user to achieve them using `$XDG_CONFIG_HOME`.

liquidctl/config/load.py
jonasmalacofilho: I agree that it can be confusing.  But it's fairly common on Unix programs, so it may be expected.

Also, I think it's the only sensible way to have both system wide and user specific settings.  If the presence of a user specific config file makes the system wide one be ignored, this may also cause some confusion.

One thing that may help here (regardless of which approach is taken) is that we provide some information on `--debug`:

```
INFO: ~/.config/liquidctl/config.toml overrides /etc/liquidctl/config.toml
INFO: section dev1.<foo>.<bar> in ~/.config/liquidctl/config.toml overrides previous values
```

liquidctl/config/load.py
MarshallAsch: That makes sense and I don't know why, but I interpreted it differently initially, however I now agree with you. 

For some reason I interpreted as `XDG_CONFIG_HOME` should be checked first and if no file is there check `$HOME/.config/` which now that I think about it makes no sense. 

Ill update this later today (and make code code a little better)

jonasmalacofilho: The counter argument is that a user that just recently set `$XDG_CONFIG_HOME` may still expect `liquidctl` to use his old configuration file, especially if it's in `~/.liquidctl.toml`.

Perhaps we should stop treating `~/.liquidctl.toml` and `~/.config/liquidctl/config.toml` like they are the same thing: the former could always be an override, while the latter could only be used as a default for missing of empty `$XDG_CONFIG_HOME`.

MarshallAsch:
I have a few other things that I need to do today, but I am planning on updating the code from all of these comments this afternoon / evening.

MarshallAsch:
I have a though @jonasmalacofilho I know that the XDG base directory system is only really used on linux machines but why not include checking that folder on all *nix like machines (everything but windows)?

jonasmalacofilho:
> I have a though @jonasmalacofilho I know that the XDG base directory system is only really used on linux machines but why not include checking that folder on all *nix like machines (everything but windows)?

It could potentially conflict with other specs or common ways of doing things in those systems, and I don't know enough about them to gauge whether that will actually happen.

But do you think it would be useful?

Also, do we know of other programs that do this on the free BSDs and/or OSX?

jonasmalacofilho:
@ehaupt Since you maintain the liquidctl ports for FreeBSD and DrangonFly BSD, can I ask your opinion on whether it would be ok, or maybe even preferable, to follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) when looking for our config file on the(se) BSDs?

In particular, this would mean our config file would _generally_ live at `$HOME/.config/liquidctl/config.toml`.

MarshallAsch:
> Also, do we know of other programs that do this on the free BSDs and/or OSX?

git seems to follow this according to their docs, and the behaviour is expressed when I move the `~/.gitconfig` file to `~/.config/git/config` on my mac

ehaupt:
> @ehaupt Since you maintain the liquidctl ports for FreeBSD and DrangonFly BSD, can I ask your opinion on whether it would be ok, or maybe even preferable, to follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) when looking for our config file on the(se) BSDs?
> 
> In particular, this would mean our config file would _generally_ live at `$HOME/.config/liquidctl/config.toml`.

From a packaging standpoint I see no reason against switching to XDG base directory locations. On my FreeBSD Xfce4 desktop I have many programs using XDG locations.

jonasmalacofilho:
@ehaupt 

Thank you!

---

@MarshallAsch 

That makes sense.  I'm fine with it then, thanks!

liquidctl/config/load.py
MarshallAsch: I am no longer checking these paths on macos, only the XDG and home dir

PR #326, opened 2021-04-11
Improve Smart Device fan status output for 1.7

jonasmalacofilho:
Improves the order of the items, how the control mode is returned/displayed, and the behavior when the device has not detected the channel as in use, but it could be.

Changes the output from

```
NZXT Smart Device (V1)
├── Fan 1                        PWM
├── Fan 1 current               0.00  A
├── Fan 1 speed                 1461  rpm
├── Fan 1 voltage              11.91  V
├── Fan 2                        PWM
├── Fan 2 current               0.02  A
├── Fan 2 speed                 1331  rpm
├── Fan 2 voltage              11.91  V
├── Fan 3                          —
├── Firmware version           1.0.7
├── LED accessories                2
├── LED accessory type    HUE+ Strip
├── LED count (total)             20
└── Noise level                   64  dB
```

to

```
NZXT Smart Device (V1)
├── Fan 1 speed                 1473  rpm
├── Fan 1 voltage              11.91  V
├── Fan 1 current               0.01  A
├── Fan 1 control mode           PWM
├── Fan 2 speed                 1341  rpm
├── Fan 2 voltage              11.91  V
├── Fan 2 current               0.02  A
├── Fan 2 control mode           PWM
├── Fan 3 speed                 1352  rpm
├── Fan 3 voltage              11.91  V
├── Fan 3 current               0.02  A
├── Fan 3 control mode           N/A
├── Firmware version           1.0.7
├── LED accessories                2
├── LED accessory type    HUE+ Strip
├── LED count (total)             20
└── Noise level                   65  dB
```

Specifically, this PR will:

- output fan speed, voltage and current even if the channel is reported as not in use (at initialization time);
- explicitly name the control mode values as that;
- order items in a more sensible way, prioritizing fan speed over control mode;
- if a channel is not in use, return control mode None at the API level;
- let the CLI convert None to "N/A" (previous the output was an em-dash).

Related: #314 (mostly around control mode == None)

## Important

This replaces the `"Fan {num}"` key, for which the possible values were `["—", "PWM", "DC"]`, with `"Fan {num} control mode"`, with possible values `[None, "PWM", "DC"]`.  The `None` gets converted to `N/A` in the CLI, in `_print_dev_status()`.

I think this is ok: the semantics of `get_status()` allow for elements to be removed/added/replaced, this is a small change to a value that's probably not used very much.  I think the improvement this brings, while not major, is still worth it (as a user of this particular device).

Still, @MarshallAsch, i would like a second opinion on this change.

---

Checklist:

- [x] Conform to the style guide in `docs/developer/style-guide.md`
- [x] Update (or add) applicable `docs/*guide.md` device guides

### DIFF
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -200,6 +200,8 @@ def _print_dev_status(dev, status):
     for k, v, u in status:
         if isinstance(v, datetime.timedelta):
             v = str(v)
+        elif isinstance(v, bool):
+            v = 'Yes' if v else 'No'
         elif v is None:
             v = 'N/A'
         else:
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -201,6 +201,8 @@ def _print_dev_status(dev, status):
         if isinstance(v, datetime.timedelta):
             v = str(v)
             u = ''
+        elif v is None:
+            v = 'N/A'
         else:
             valfmt = _VALUE_FORMATS.get(u, '')
             v = f'{v:{valfmt}}'
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -200,7 +200,6 @@ def _print_dev_status(dev, status):
     for k, v, u in status:
         if isinstance(v, datetime.timedelta):
             v = str(v)
-            u = ''
         elif v is None:
             v = 'N/A'
         else:
--- a/liquidctl/driver/commander_pro.py
+++ b/liquidctl/driver/commander_pro.py
@@ -188,7 +188,7 @@ class CommanderPro(UsbHidDriver):
             fanModes = res[1:self._fan_count+1]
             self._data.store('fan_modes', fanModes)
             status += [
-                ('Fan {i + 1} control mode', _fan_mode_desc(fanModes[i]), '')
+                (f'Fan {i + 1} control mode', _fan_mode_desc(fanModes[i]), '')
                 for i in range(6)
             ]
 
--- a/liquidctl/driver/commander_pro.py
+++ b/liquidctl/driver/commander_pro.py
@@ -99,18 +99,18 @@ def _quoted(*names):
     return ', '.join(map(repr, names))
 
 
-def _get_fan_mode_description(mode):
+def _fan_mode_desc(mode):
     """This will convert the fan mode value to a descriptive name.
     """
 
-    if mode == _FAN_MODE_DISCONNECTED:
-        return 'Auto/Disconnected'
-    elif mode == _FAN_MODE_DC:
+    if mode == _FAN_MODE_DC:
         return 'DC'
     elif mode == _FAN_MODE_PWM:
         return 'PWM'
     else:
-        return 'UNKNOWN'
+        if mode != _FAN_MODE_DISCONNECTED:
+            _LOGGER.warning('unknown fan mode: {mode:#04x}')
+        return None
 
 
 class CommanderPro(UsbHidDriver):
@@ -178,10 +178,8 @@ class CommanderPro(UsbHidDriver):
             temp_connected = res[1:5]
             self._data.store('temp_sensors_connected', temp_connected)
             status += [
-                ('Temp sensor 1', 'Connected' if temp_connected[0] else 'Not Connected', ''),
-                ('Temp sensor 2', 'Connected' if temp_connected[1] else 'Not Connected', ''),
-                ('Temp sensor 3', 'Connected' if temp_connected[2] else 'Not Connected', ''),
-                ('Temp sensor 4', 'Connected' if temp_connected[3] else 'Not Connected', ''),
+                (f'Temperature probe {i + 1}', bool(temp_connected[i]), '')
+                for i in range(4)
             ]
 
         if self._fan_count > 0:
@@ -190,12 +188,8 @@ class CommanderPro(UsbHidDriver):
             fanModes = res[1:self._fan_count+1]
             self._data.store('fan_modes', fanModes)
             status += [
-                ('Fan 1 Mode', _get_fan_mode_description(fanModes[0]), ''),
-                ('Fan 2 Mode', _get_fan_mode_description(fanModes[1]), ''),
-                ('Fan 3 Mode', _get_fan_mode_description(fanModes[2]), ''),
-                ('Fan 4 Mode', _get_fan_mode_description(fanModes[3]), ''),
-                ('Fan 5 Mode', _get_fan_mode_description(fanModes[4]), ''),
-                ('Fan 6 Mode', _get_fan_mode_description(fanModes[5]), ''),
+                ('Fan {i + 1} control mode', _fan_mode_desc(fanModes[i]), '')
+                for i in range(6)
             ]
 
         return status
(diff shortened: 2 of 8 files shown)


liquidctl/cli.py
```diff
@@ -201,6 +201,8 @@ def _print_dev_status(dev, status):
         if isinstance(v, datetime.timedelta):
             v = str(v)
             u = ''
+        elif v is None:
+            v = 'N/A'
```
jonasmalacofilho, line 205: I wasn't sure if I should set `u = ''` or not (in the general case).

MarshallAsch:
I will check this tomorrow evening. 

Is the intention of this set of changes to only be for the smart devices, or also other devices that can report the fan control mode?

jonasmalacofilho:
I thought this was the only device to report it, but now I see there's the Commander Pro as well.

In that case, I would like to make the output from both a bit more similar...  I'll take a better a look, but I'm thinking something in the lines of (comments only for this discussion):

```diff
diff --git a/liquidctl/driver/commander_pro.py b/liquidctl/driver/commander_pro.py
index 3c5f78291728..c0c937c8ab39 100644
--- a/liquidctl/driver/commander_pro.py
+++ b/liquidctl/driver/commander_pro.py
@@ -104,13 +104,13 @@ def _get_fan_mode_description(mode):
     """
 
     if mode == _FAN_MODE_DISCONNECTED:
-        return 'Auto/Disconnected'
+        return 'N/A'  # the fan cannot be controlled (at least in the current implementation)
     elif mode == _FAN_MODE_DC:
         return 'DC'
     elif mode == _FAN_MODE_PWM:
         return 'PWM'
     else:
-        return 'UNKNOWN'
+        return 'Unknown'  # DC and PWM are acronyms, this isn't
 
 
 class CommanderPro(UsbHidDriver):
@@ -178,10 +178,11 @@ class CommanderPro(UsbHidDriver):
             temp_connected = res[1:5]
             self._data.store('temp_sensors_connected', temp_connected)
             status += [
-                ('Temp sensor 1', 'Connected' if temp_connected[0] else 'Not Connected', ''),
-                ('Temp sensor 2', 'Connected' if temp_connected[1] else 'Not Connected', ''),
-                ('Temp sensor 3', 'Connected' if temp_connected[2] else 'Not Connected', ''),
-                ('Temp sensor 4', 'Connected' if temp_connected[3] else 'Not Connected', ''),
+                # don't use title case for consistency
+                ('Temp sensor 1', 'Connected' if temp_connected[0] else 'Not connected', ''),
+                ('Temp sensor 2', 'Connected' if temp_connected[1] else 'Not connected', ''),
+                ('Temp sensor 3', 'Connected' if temp_connected[2] else 'Not connected', ''),
+                ('Temp sensor 4', 'Connected' if temp_connected[3] else 'Not connected', ''),
             ]
 
         if self._fan_count > 0:
@@ -190,12 +191,13 @@ class CommanderPro(UsbHidDriver):
             fanModes = res[1:self._fan_count+1]
             self._data.store('fan_modes', fanModes)
             status += [
-                ('Fan 1 Mode', _get_fan_mode_description(fanModes[0]), ''),
-                ('Fan 2 Mode', _get_fan_mode_description(fanModes[1]), ''),
-                ('Fan 3 Mode', _get_fan_mode_description(fanModes[2]), ''),
-                ('Fan 4 Mode', _get_fan_mode_description(fanModes[3]), ''),
-                ('Fan 5 Mode', _get_fan_mode_description(fanModes[4]), ''),
-                ('Fan 6 Mode', _get_fan_mode_description(fanModes[5]), ''),
+                # add explicit "control"
+                ('Fan 1 control mode', _get_fan_mode_description(fanModes[0]), ''),
+                ('Fan 2 control mode', _get_fan_mode_description(fanModes[1]), ''),
+                ('Fan 3 control mode', _get_fan_mode_description(fanModes[2]), ''),
+                ('Fan 4 control mode', _get_fan_mode_description(fanModes[3]), ''),
+                ('Fan 5 control mode', _get_fan_mode_description(fanModes[4]), ''),
+                ('Fan 6 control mode', _get_fan_mode_description(fanModes[5]), ''),
             ]
 
         return status
```

Additionally, I'm also starting to wonder if we can omit "control" xor "mode" from the labels.

liquidctl/cli.py
MarshallAsch: Do you mean also set the units to `''` when the value is `None`?

MarshallAsch (changes requested):
Im marking this as request changes only for the commander pro stuff that there is a comment for. But other than that I think it looks good. I do like the idea of having consistent output quite a bit.

liquidctl/cli.py
jonasmalacofilho: Exactly, in case we ever get something like a temperature that's N/A.

In this hypothetical scenario `get_status()` would probably return `("That temperature", None, "°C")`, as that's the natural thing to do inside a driver.  But what should we output for `--json`?

jonasmalacofilho (commented):
Thanks for the review!

Just to double check: have you considered and do you agree with my argument in the **Important** section of the PR description?

MarshallAsch:
> Just to double check: have you considered and do you agree with my argument in the Important section of the PR description?

Yes I did read that and I do agree. 
Related to #314 would the value still be `N/A` in the JSON payload?

liquidctl/cli.py
MarshallAsch: That situation happens for the commander pro if the status is requested before the device is initialized, well currently it returns 0 for everything.  (it was an implementation decision that I decided to make, although that could have been a bad one to make).

for JSON we could make the value `null` which is valid in the spec or coerce it to a `0`? although I don't like the 0 idea as much

jonasmalacofilho: Sorry, I don't know why I mentioned `--json`, it has nothing to do with this PR.

Instead, what should we output here (for humans, not `--json`)?

```
That temperature        N/A  °C
```

or

```
That temperature        N/A  
```

For consistency with other (presumably available) values, I prefer the first one, but I'm not 100% sure.

jonasmalacofilho: Re. the Commander Pro, if we're 100% sure that the device hasn't been initialized, then a warning + `None` (N/A for humans) could be an improvement, yes.  But only if we're certain that's not what the device is actually reporting... if the device is initialized and reports 0 than we should keep that zero.

jonasmalacofilho:
> Related to #314 would the value still be N/A in the JSON payload?

No, it will be `null`.  I was going to change that there after this PR, but thanks for reminding me anyway.

liquidctl/cli.py
MarshallAsch: > Instead, what should we output here ...

I kind of like the second one, but the first is more consistent so I think that is probably a better choice to go with

jonasmalacofilho:
I'll submit the changes to the Commander Pro as a separate PR.

jonasmalacofilho:
@MarshallAsch

Besides a quick review of the Commander Pro changes, can you also runtime test the new code (& CLI) with your device?

MarshallAsch:
The comander pro does not work:

```bash
(venv) $  liquidctl initialize all
Corsair Hydro H100i Platinum (experimental)
└── Firmware version    1.1.15  

Corsair Commander Pro (experimental)
├── Firmware version            0.9.212  
├── Bootloader version              0.5  
├── Temperature probe 1             Yes  
├── Temperature probe 2             Yes  
├── Temperature probe 3             Yes  
├── Temperature probe 4             Yes  
├── Fan {i + 1} control mode         DC  
├── Fan {i + 1} control mode         DC  
├── Fan {i + 1} control mode         DC  
├── Fan {i + 1} control mode        N/A  
├── Fan {i + 1} control mode        N/A  
└── Fan {i + 1} control mode        N/A  
```

liquidctl/driver/commander_pro.py
```diff
@@ -190,12 +188,8 @@ def initialize(self, **kwargs):
             fanModes = res[1:self._fan_count+1]
             self._data.store('fan_modes', fanModes)
             status += [
-                ('Fan 1 Mode', _get_fan_mode_description(fanModes[0]), ''),
-                ('Fan 2 Mode', _get_fan_mode_description(fanModes[1]), ''),
-                ('Fan 3 Mode', _get_fan_mode_description(fanModes[2]), ''),
-                ('Fan 4 Mode', _get_fan_mode_description(fanModes[3]), ''),
-                ('Fan 5 Mode', _get_fan_mode_description(fanModes[4]), ''),
-                ('Fan 6 Mode', _get_fan_mode_description(fanModes[5]), ''),
+                ('Fan {i + 1} control mode', _fan_mode_desc(fanModes[i]), '')
```
MarshallAsch, line 191: missing the `f` here for the format string

jonasmalacofilho: Thanks, fixed.

MarshallAsch:
this seems all good now, it worked as expected. on my device

Issue #218, opened 2020-11-12
Corsair Commander Core


jbrgilbrech:
Can you add support for this AIO?

---

Original issue title: Corsair H150 Elite Cappalex

billygilbrech:
If you need anything also please add me on Discord billyg#2475

billygilbrech:
This devices connected to the Corsair Commander Core.

MarshallAsch:
Are you able to give the output from `lsusb -v` for that device?

And would you be able to capture some USB traffic between iCue and the device?

jonasmalacofilho:
Hopefully the Commander CORE, like the Lighting Node CORE, is also based on the Commander Pro.

billygilbrech:
I have no idea how to capture traffic.

billygilbrech:
Also on Windows, what's a softtware to grab USB Ids

jonasmalacofilho:
@billygilbrech, please try start with [this short guide](https://github.com/jonasmalacofilho/liquidctl/blob/master/docs/developer/capturing-usb-traffic.md#capturing-usb-traffic-on-a-native-windows-host).

We can probably infer which is the correct device by ourselves, and the traffic capture should also give us the USB IDs and descriptors.

billygilbrech:
[capture.zip](https://github.com/jonasmalacofilho/liquidctl/files/5539989/capture.zip)

billygilbrech:
I changed all settings from the AIO in iCue. Here are the provided captures I got.

billygilbrech:
1b1c:0c1c

MarshallAsch:
This seems to use the `usb.capdata` field for almost all of the communications. and there are an awful lot of really long messages that are almost entirely zeros. You didn't make any notes about what you changed when do you? @jonasmalacofilho I'm not quite sure how to figure out what the captured data actually means without a description of what happened when.


@billygilbrech can you please also give the output of `lsusb -d 1b1c:0c1c -v` (that gives us some information such as the request and responce lengths to help filter out some of the other usb traffic that is on that usb bus)

In that capture there are what appears to be traffic with 13 different devices. on that bus.

jonasmalacofilho:
@MarshallAsch 

The best way I found to filter these captures is by searching for the right `GET DESCRIPTOR Response DEVICE`: in this case this leads to filter with the expression `usb.device_address == 9`.
[...]

PR #348, opened 2021-06-08
Commander Core: Cleanup and add read error checking

ParkerMc:
Clean up Commander Core code and documentation.

Ensure we received the right data on read.  
The first two bytes of data in response to the read command appear to be a data type so now those are being checked.

Related: #218 

---

Checklist:

- [x] Submit relevant data, scripts or dissectors to https://github.com/liquidctl/collected-device-data

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/commander_core.py
+++ b/liquidctl/driver/commander_core.py
@@ -12,7 +12,7 @@ SPDX-License-Identifier: GPL-3.0-or-later
 import logging
 
 from liquidctl.driver.usb import UsbHidDriver
-from liquidctl.error import NotSupportedByDriver
+from liquidctl.error import ExpectationNotMet, NotSupportedByDriver
 from liquidctl.util import u16le_from
 
 _LOGGER = logging.getLogger(__name__)
@@ -22,15 +22,20 @@ _RESPONSE_LENGTH = 1024
 
 _INTERFACE_NUMBER = 0
 
+_CMD_WAKE = (0x01, 0x03, 0x00, 0x02)
+_CMD_SLEEP = (0x01, 0x03, 0x00, 0x01)
 _CMD_GET_FIRMWARE = (0x02, 0x13)
-_CMD_INIT = (0x01, 0x03, 0x00, 0x02)
 _CMD_RESET = (0x05, 0x01, 0x00)
 _CMD_SET_MODE = (0x0d, 0x00)
 _CMD_GET = (0x08, 0x00)
 
-_MODE_DETECT_RGB = 0x20
-_MODE_GET_SPEEDS = 0x17
-_MODE_GET_TEMPS = 0x21
+_MODE_LED_COUNT = (0x20,)
+_MODE_GET_SPEEDS = (0x17,)
+_MODE_GET_TEMPS = (0x21,)
+
+_DATA_TYPE_SPEEDS = (0x06, 0x00)
+_DATA_TYPE_LED_COUNT = (0x0f, 0x00)
+_DATA_TYPE_TEMPS = (0x10, 0x00)
 
 
 class CommanderCore(UsbHidDriver):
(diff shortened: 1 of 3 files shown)


jonasmalacofilho:
_(just copying what we already talked about in the Discord server, so Marshall can find everything in a single place)_

One suggestion I have is to maybe use a context manager (pseudo code):

```py
def get_status(self, **kwargs):
    with self._hardware_fallback():
        for i, speed in enumerate(self._get_speeds()):
            # ...
        # ...
        return status

@contextmanager
def _hardware_fallback(self):
    try:
        self._send_command(_CMD_WAKE)
        yield self
    finally:
        self._send_command(_CMD_SLEEP)
```

Well, the example above is not great: it's weird that something called `_hardware_fallback()` sends the `_CMD_WAKE`...

The other thing is that it may be sensible to add a couple of tests to ensure that both `initialize()` and `get_status()` will indeed fallback to hardware mode (and to make sure we don't break this in the future).

liquidctl/driver/commander_core.py
```diff
@@ -9,10 +9,11 @@
 SPDX-License-Identifier: GPL-3.0-or-later
 """
 
+from contextlib import contextmanager
```
jonasmalacofilho, line 12: Nit: `import ...` before `from ... import ...`.

liquidctl/driver/commander_core.py
```diff
@@ -140,6 +138,17 @@ def _get_temps(self):
 
         return temps
 
+    def _read_data(self, mode, data_type):
+        self._send_command(_CMD_RESET)
+        self._send_command(_CMD_SET_MODE, mode)
+        res = self._send_command(_CMD_GET)
+
+        input_type = tuple(x for x in res[3:5])
+        if input_type != data_type:
```
jonasmalacofilho, line 147:
```suggestion
        if tuple(res[3:5]) != data_type:
```

liquidctl/driver/commander_core.py
```diff
@@ -140,6 +138,17 @@ def _get_temps(self):
...
+            raise ExpectationNotMet('Device returned incorrect data type')
```
jonasmalacofilho, line 148: Nit: error messages should start with a lowercase letter.

liquidctl/driver/commander_core.py
```diff
@@ -43,50 +49,46 @@ class CommanderCore(UsbHidDriver):
...
-            num_leds = u16le_from(res, offset=8+i*4)
-            label = 'AIO LED count' if i == 0 else f'RGB port {i} LED count'
-            status += [(label, num_leds if connected else None, '')]
-
-        # Get what temp sensors are connected
-        for i, temp in enumerate(self._get_temps()):
-            connected = temp is not None
-            label = 'Water temperature sensor' if i == 0 else f'Temperature sensor {i}'
-            status += [(label, connected, '')]
+            num_devices = raw_data[0]
```
jonasmalacofilho, line 62: I find `raw_data` slightly confusing for someone trying to quickly match the code to a capture, given that it isn't the raw response (which is what I would expect based on the name).

That said, I think you're right to strip out the boilerplate in `_read_data()`, so maybe just invert the names: use `res` in the callers (e.g. here) and `raw_data` inside `_read_data()`.

jonasmalacofilho (commented):
Thanks!  Looks good to me, and I only spotted a few minor code style issues.

P.S. I had previously misunderstood the purpose of the `_CMD_SLEEP` (but I don't think it changed anything).

ParkerMc:
> P.S. I had previously misunderstood the purpose of the `_CMD_SLEEP` (but I don't think it changed anything).

Would it be better to name it something else? Wake and sleep are just the two best names I could come up with.

jonasmalacofilho:
> Would it be better to name it something else? Wake and sleep are just the two best names I could come up with.

No need, I think they are fine.

jonasmalacofilho (approved):

jonasmalacofilho:
Thanks again!

ParkerMc:
Thank you for all the guidance and the base to build off of!

Issue #367, opened 2021-08-01
Corsair Lighting Node Core


Collbrothers:
Running `liquidctl --match Core set sync color fixed 0080ff` will only affect one pixel of one fan (see img below)
https://imgur.com/a/Ro9jO8i
Additional information about the system:

- Affected device: Corsair QL 120 RGB & Node Core
- Operating system: Arch Linux
- Liquidctl version: AUR 1.7.0

jonasmalacofilho:
@MarshallAsch,

I think `--maximum-leds` should default to all the LEDs, so that the lighting modes also default to all LEDs/fans when neither this nor the `--start-led` option has been specified.

What do you think?  Is it at least feasible to implement this change?

(The simplest way to implement it would be to default to the maximum number of LEDs that a channel can support.  But does the controller accept that without locking up, even when there aren't as many connected accessories/LED, or when `--start-led` is greater than one?)

MarshallAsch:
It does not mess up the device if too many leds are specified.

I think that that would be a good idea and I can get that implemented tomorrow morning.

jonasmalacofilho:
Thanks!

MarshallAsch:
@Collbrothers do you have 6 of the QL fans connected? And if you do can you confirm that the command `liquidctl --match Core set sync color fixed 0080ff --maximum-leds 204` does indeed turn on all of the LEDs, @jonasmalacofilho identified a potential off by one error that might result in the last LED to not light up.

jonasmalacofilho:
I think that test can also be adapted to 4 or any other number of fans.  Which brings another question: @MarshallAsch , do you want @Collbrothers to test with the version he already has installed (1.7.0, apparently[1]) or with the latest commit in #368?

---

@Collbrothers, you said you have liquidctl AUR 1.7.0; that's odd, on Arch liquidctl is on the official repositories (community) at 1.7.0, or on AUR as a `-git` package.  There's no 1.7.0 on the AUR.  Can you clarify what version you have installed (maybe show us the output of `liquidctl --version`)?

MarshallAsch:
I actually think the new code is fine I tested it by scaling the Max leds to 3 and the edge behaviour worked as expected, so it may have been wrong before (it would have only been an issue for the edge case)

PR #368, opened 2021-08-02
changed the default number of the maximum LEDs to use for the commander node deviced

MarshallAsch:

This will change the default value for the `maximum-leds` flag to be all LED's and not just `1` LED. 

Fixes: #367 

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 _Summary for the 1.7.1 release: fix a bug when colorizing the log output._
 
 Changelog since 1.7.0:
+### Changed
+ - Changed the default behavior for the default value for `maximum-leds` to be all the LED's instead of one for the Corsair Commander family of devices 
 ### Fixed
  - Fix `KeyError` when logging due to colorlog<6
  - Swap DEBUG and INFO level colors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
 _Summary for the 1.7.1 release: fix a bug when colorizing the log output._
 
 Changelog since 1.7.0:
-### Changed
- - Changed the default behavior for the default value for `maximum-leds` to be all the LED's instead of one for the Corsair Commander family of devices 
 ### Fixed
  - Fix `KeyError` when logging due to colorlog<6
  - Swap DEBUG and INFO level colors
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 _Summary for the 1.7.1 release: fix a bug when colorizing the log output._
 
 Changelog since 1.7.0:
+### Changed
+ - Changed the default behavior for the default value for `maximum-leds` to be all the LED's instead of one for the Corsair Commander family of devices
```
jonasmalacofilho, line 8: This entry is the the wrong place: 1.7.1 has already been release.

For this reason and a few others I generally prefer to update the changelog independently of the actual changes, usually once a week.

So you can just drop this and I'll update the changelog later.

P.S. I also considering marking this as a "fix", which I think is reasonable and would also allow us to ship it in a 1.7.2.  This is a bit more relevant than usual because it seems that this release cycle is going slow, so it's possible that the October release may turn out to just be 1.7.2.

tests/test_commander_pro.py
```diff
@@ -864,7 +864,7 @@ def test_set_color_hardware_default_start_end(commanderProDevice):
 
     assert sent[3].data[0] == 0x35
     assert sent[3].data[2] == 0x00  # start led
-    assert sent[3].data[3] == 0x01  # num leds
+    assert sent[3].data[3] == 203  # num leds
```
jonasmalacofilho, line 867: Aren't there 204 LEDs between 1 and 204?  So if the comment is right ("num leds"), should this assert if the value is equal to 204?

CHANGELOG.md
MarshallAsch: oops my bad, I just read the since part and put it in, but Ill remove it

tests/test_commander_pro.py
MarshallAsch: I may have been off by one in the initial code that set the value. I believe there was one user that had the maximum number. I asked the original reporter if they are able to confirm the off by one error. But ill make the change to fix that as well and add a test for it if there is not one already

jonasmalacofilho:
LGTM, but waiting to see what's the answer to the question you made on #367.

PR #371, opened 2021-08-19
Add status support for MSI MPG Coreliquid K360

Tatsh:
This PR adds minimal support to display status information from an MSI MPG Series CORELIQUID K360 AIO device.

Other devices probably work, but I do not yet know their device IDs.

---

Checklist:

New CLI flag?

New device?

New driver?


README.md
```diff
@@ -120,14 +121,14 @@ The following devices are supported by this version of liquidctl.  See each guid
...
 | Power supply | [NZXT E500, E650, E850](docs/nzxt-e-series-psu-guide.md) | USB HID | <sup>_p_</sup> |
 
-<sup>_L_</sup> _Requires the `--legacy-690lc` flag._  
-<sup>_U_</sup> _Requires `--unsafe` features._  
-<sup>_Z_</sup> _Requires replacing the device driver [on Windows](#installing-on-windows)._  
-<sup>_a_</sup> _Architecture-specific limitations._  
-<sup>_e_</sup> _Experimental support._  
-<sup>_n_</sup> _New driver, only available on git._  
-<sup>_p_</sup> _Only partially supported._  
-<sup>_x_</sup> _Only supported on Linux._
```
jonasmalacofilho, line 130: The trailing whitespace is semantically relevant here, and required by Markdown to force a line break.

README.md
```diff
@@ -596,7 +597,7 @@ You are also encouraged to contribute to the documentation and to these examples
 
 ## License
 
-liquidctl – monitor and control liquid coolers and other devices.
```
jonasmalacofilho, line 599: Relevant trailing whitespace again.

docs/msi-mpg-series-coreliquid-guide.md
```diff
@@ -0,0 +1,3 @@
+# MSI MPG Series CORELIQUID
+
+This driver only supports reporting status. Only the K360 model is supported.
```
jonasmalacofilho, line 3: This file is important and needs to follow the same general template of the other `*guide.md` files.

tests/test_msi_7zw0led.py
```diff
@@ -0,0 +1,49 @@
...
+        buf += pack('h', 509)
+        buf += pack('h', 1045)
+        buf += pack('h', 1754)
+        buf += bytearray([0, 0, 0, 0, 0x7d, 0, 0x7d, 0, 0, 0])
+        buf += pack('h', 20)
+        buf += pack('h', 20)
+        buf += pack('h', 20)
+        buf += pack('h', 20)
+        buf += pack('h', 50)
+        buf += bytearray([0] * 32)
```
jonasmalacofilho, line 31: Aren't these `pack()` calls fragile regarding different endianess?

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
...
+            ('Fan 1 speed', array[2] + (array[3] << 8), 'rpm'),
+            ('Fan 1 duty', array[0x16] + (array[0x17] << 8), '%'),
+            ('Fan 2 speed', array[4] + (array[5] << 8), 'rpm'),
+            ('Fan 2 duty', array[0x18] + (array[0x19] << 8), '%'),
+            ('Fan 3 speed', array[6] + (array[7] << 8), 'rpm'),
+            ('Fan 3 duty', array[0x1A] + (array[0x1B] << 8), '%'),
+            ('Water block fan speed', array[8] + (array[9] << 8), 'rpm'),
+            ('Water block fan duty', array[0x1C] + (array[0x1D] << 8), '%'),
+            ('Pump fan speed', array[0xA] + (array[0xB] << 8), 'rpm'),
+            ('Pump fan duty', array[0x1E] + (array[0x1F] << 8), '%'),
```
jonasmalacofilho, line 42: Shouldn't these be "Pump speed" and "Pump duty" (no "fan")?

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_status(self, **kwargs):
+        buf = bytearray(_REPORT_LENGTH)
+        buf[1] = 0xd0
+        buf[2] = 0x31
+        self.device.clear_enqueued_reports()
+        self.device.write(buf)
+        array = bytes(self.device.read(_REPORT_LENGTH))
+        if array[1] != 0x31:
+            raise ValueError('Unexpected value in response buffer')
```
jonasmalacofilho, line 31: Assuming this is a rare event, maybe an assertion would be more appropriate?

(If on the other hand this is a common occurrence, then we need to handle it some other way, perhaps with retries).

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
...
+Copyright (C) 2020–2021  Andrew Udvare and contributors
+SPDX-License-Identifier: GPL-3.0-or-later
+"""
+
+from liquidctl.driver.usb import UsbHidDriver
+
+_REPORT_LENGTH = 64
+
+
+class MSI7ZW0LED(UsbHidDriver):
```
jonasmalacofilho, line 16: `MSI7ZW0LED` is quite hard to read.  I think `MpgCooler` is a better one.

("MPG" is important because the "MAG" line appears to use a different OEM, meaning it will probably also use a different protocol; on the other hand, "MSI" can be inferred from "MPG").

README.md
```diff
@@ -95,6 +95,7 @@ The following devices are supported by this version of liquidctl.  See each guid
 | AIO liquid cooler | [Corsair Hydro H100i Pro XT, H115i Pro XT, H150i Pro XT](docs/corsair-platinum-pro-xt-guide.md) | USB HID | <sup>_e_</sup> |
 | AIO liquid cooler | [Corsair iCUE H100i, H115i, H150i Elite Capellix](docs/corsair-commander-core-guide.md) | USB HID | <sup>_ep_</sup> |
 | AIO liquid cooler | [EVGA CLC 120 (CL12), 240, 280, 360](docs/asetek-690lc-guide.md) | USB | <sup>_Z_</sup> |
+| AIO liquid cooler | [MSI MPG Series CORELIQUID K360](docs/msi-mpg-series-coreliquid-guide.md) | USB HID | <sup>_penx_</sup> |
```
jonasmalacofilho, line 98: MSI, as most other vendors, abuses all caps and long product names.

When that happens, we intentionally deviate from the offical name and adopt something a bit more sensible; in this case, I like "MSI MPG Coreliquid K360".

extra/linux/71-liquidctl.rules
```diff
@@ -153,6 +153,9 @@ SUBSYSTEMS=="usb", ATTRS{idVendor}=="048d", ATTRS{idProduct}=="5702", TAG+="uacc
 # Gigabyte RGB Fusion 2.0 8297 Controller
 SUBSYSTEMS=="usb", ATTRS{idVendor}=="048d", ATTRS{idProduct}=="8297", TAG+="uaccess"
 
+# MSI MPG Series CORELIQUID K360
```
jonasmalacofilho, line 156: MSI MPG Coreliquid K360.

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
...
+    """7ZW0LED is the internal name in the MSI Center app."""
+
+    SUPPORTED_DEVICES = [
+        (0x0db0, 0xb130, None, 'MSI MPG Series CORELIQUID K360', {}),
```
jonasmalacofilho, line 20: MSI MPG Coreliquid K360.

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
+"""liquidctl drivers for MSI liquid coolers.
+
+Supported devices:
+
+- MPG CORELIQUID K360
```
jonasmalacofilho, line 5: MPG Coreliquid K360.

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
+"""liquidctl drivers for MSI liquid coolers.
```
jonasmalacofilho, line 1: If this module is meant to be the home of the drivers for many different MSI liquid coolers families (which makes sense), then I think it should be called `msi.py` or, at most, `msi_cooler.py`.

liquidctl/driver/msi_7zw0led.py
```diff
@@ -0,0 +1,43 @@
...
+
+Supported devices:
+
+- MPG CORELIQUID K360
+
+Copyright (C) 2020–2021  Andrew Udvare and contributors
```
jonasmalacofilho, line 7: Did you really start this in 2020?

README.md
```diff
@@ -41,14 +41,14 @@ NZXT Kraken X (X42, X52, X62 or X72)
 ├── Liquid temperature     31.7  °C
 ├── Fan speed               801  rpm
 ├── Pump speed             2239  rpm
-└── Firmware version      6.0.2
```
jonasmalacofilho, line 44: This change is fine, but FYI the actual output has those two trailing spaces when the status item has no unit.

jonasmalacofilho (commented):
Hi,

Thanks!

I only have a couple of questions (like confirming which are the pump sensors and the endianess impact on `pack()`) and a few other minor notes and/or nitpicks.

Regarding the man page, I don't think a device that only supports `status` needs an explicit section there.  I imagine that was your reasoning as well?

tests/test_msi_7zw0led.py
Tatsh: I think so. They follow USB's endianness so little-endian. Will add the endianness symbols.

liquidctl/driver/msi_7zw0led.py
Tatsh: IIRC as I cannot see it right now, the MSI UI says 'Pump Fan' and 'Water Block Fan'. Maybe this is a mistake on their part and I'm happy to correct it. Of course there is no physical 'pump fan' or 'water block fan'. As for the speed reported, I don't know what it means.

Tatsh:
I might get more time to contribute the rest of the implementation. It's kind of easy since we can just open MSI's .NET DLL in ILSpy or similar, but there's a fair bit of functionality.

Tatsh:
Random question: how come I am getting the output twice for a single device? It seems that hidraw's `enumerate()` is outputting duplicates for me.

```
 $ liquidctl list
Device #0: MSI MPG Coreliquid K360
Device #1: MSI MPG Coreliquid K360
```

```
(ins)>>> pp([x for x in hidraw.enumerate() if x['manufacturer_string'] == 'MSI'])
[{'interface_number': 0,
  'manufacturer_string': 'MSI',
  'path': b'/dev/hidraw6',
  'product_id': 45360,
  'product_string': 'MPG CORELIQUID',
  'release_number': 1,
  'usage': 0,
  'usage_page': 1,
  'vendor_id': 3504},
 {'interface_number': 0,
  'manufacturer_string': 'MSI',
  'path': b'/dev/hidraw6',
  'product_id': 45360,
  'product_string': 'MPG CORELIQUID',
  'release_number': 1,
  'usage': 1,
  'usage_page': 65280,
  'vendor_id': 3504}]
```

Tatsh:
I think I found the reason. https://github.com/libusb/hidapi/pull/139

Do you plan to deduplicate [this call](https://github.com/liquidctl/liquidctl/blob/main/liquidctl/driver/usb.py#L479)? I think it will not hurt anything.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,40 @@
...
+    ]
+
+    def get_status(self, **kwargs):
+        buf = bytearray(_REPORT_LENGTH)
+        buf[1] = 0xd0
+        buf[2] = 0x31
+        self.device.clear_enqueued_reports()
+        self.device.write(buf)
+        array = bytes(self.device.read(_REPORT_LENGTH))
+        assert array[1] == 0x31, 'Unexpected value in response buffer'
```
MarshallAsch, line 28: Is there a reason for this assertion? Are there other possible values that could be here that represent a known error case?

tests/test_msi.py
```diff
@@ -0,0 +1,49 @@
...
+        buf += pack('<h', 20)
+        buf += pack('<h', 20)
+        buf += pack('<h', 50)
+        buf += bytearray([0] * 32)
+        return buf[:length]
+
+
+def test_mpg_core_liquid_k360_get_status(mpgCoreLiquidK360Device):
+    dev = mpgCoreLiquidK360Device
+    fan1, fan1d, fan2, fan2d, fan3, fan3d, wbfan, wbfand, pump, pumpd = dev.get_status()
```
MarshallAsch, line 37: Can you add a test for an invalid payload? (related to previous comment)

MarshallAsch (changes requested):
Are you able to also addd a protocol description to the `docs/developer/protocol` folder and open a related PR to https://github.com/liquidctl/collected-device-data that adds some packet traces.

liquidctl/driver/msi.py
Tatsh: This check is in the original code that MSI wrote so I thought I should replicate it.

Tatsh: > Are you able to also add a protocol description to the `docs/developer/protocol` folder and open a related PR to https://github.com/liquidctl/collected-device-data that adds some packet traces.

Yes. I can provide [a lot](https://github.com/Tatsh/liquidctl/commit/390baa396cbb2b610cc7090ef80b98b734bd2571).

Unfortunately the device I got was defective and I already sent it back to Newegg for a refund. The branch linked has every single function converted from the C# decompilation to Python. I have tested everything with the real hardware except uploading image files to the OLED.

I went with a Corsair device that has a Commander Core, which is not fully implemented or bug free. In particular, every time `status` is called it is resetting the device's power state causing it to blink. The iCUE software does not do this.

MarshallAsch: Is there a reason why you are not making the PR using [this branch](https://github.com/Tatsh/liquidctl/commit/390baa396cbb2b610cc7090ef80b98b734bd2571)? if everything was tested there to see if it worked?

Also the protocol description would be something like [this](https://github.com/liquidctl/liquidctl/blob/main/docs/developer/protocol/lighting_node_rgb.md) rather than an implementation, it allows for easier understanding of the protocol than reading through the code, and the protocol _should_ remain consistent even if the actual implementation changes or is only implemented in parts (as is the case for this PR)

jonasmalacofilho:
Given that you no longer have a device to test the driver with, maybe we should keep this PR open with the full code (i.e. including what is currently only on your `msik360-2` branch) and wait for someone else to come in.

While I'm fairly confident that the current code in this PR mostly works, having just `get_status()` available and no real development effort/interest for the remaining features is probably not very useful.

There's also the issue of multiple usage/usage pages on platforms other than Linux (per comments on #372).  It's quite possible that this driver wont always work on Mac and/or Windows.

Tatsh:
How about closing for now and reopen if someone files an issue about this device or similar? I can keep my fork with the fully converted code available.

jonasmalacofilho:
I think a draft PR would have the benefit of making your work easier to find.

jonasmalacofilho:
I haven't checked the license/EULA that applies to the MSI tool, but I'm guessing that you can't legally redistribute `MSI_7ZW0Led.cs`.  If that's indeed the case, we can't have it here either.

The argument against `extra/unpack-api-coreliquid-dll` may be more nuanced (depending on what the EULA states, and how applicable the EULA is), but I still don't think it's worth the potential pain.  Plus it's not something that should live in `extra/` anyway.

Tatsh:
> I haven't checked the license/EULA that applies to the MSI tool, but I'm guessing that you can't legally redistribute `MSI_7ZW0Led.cs`. If that's indeed the case, we can't have it here either.

Usually EULAs do not allow disassembly or decompiling. I have not read MSI's but it's standard text. The question is whether or not it applies especially if you only download the file and don't install the software like what my script does. I think in that case it doesn't but IANAL.

If you want me to remove it that's fine, but I wouldn't see too much issue with code that is generated by a decompiler. The decompilation is an interpretation and is not MSI's code.  [Some companies fail to understand this](https://github.com/GTAmodding/re3/issues/1336).

> The argument against `extra/unpack-api-coreliquid-dll` may be more nuanced (depending on what the EULA states, and how applicable the EULA is), but I still don't think it's worth the potential pain. Plus it's not something that should live in `extra/` anyway.

Where should it go?

jonasmalacofilho:
Let's leave the [decompiled] MSI code out entirely.  As for your extraction script, it's not relevant to our driver in the long term, so I think a better place for it would be (pasting it in) this thread.

---

Both of your arguments are interesting, I'll try to read up of more things along those lines.  But, for now, I'm personally not fully convinced that they would hold, mostly on the copyright side: a test I've seen cited many times is whether the argument would hold (or at least not fail completely) when applied to a book.

In the first case, skipping the page that contains the copyright notice and taking out later pages from a book does not bypass copyright.  (Also: I don't want a world where every book has the copyright notice on the cover and all pages just to prevent a bypass like this).

In the other one, compilation and subsequent decompilation might both be comparable to translations (I've seen at least one mention of this, IIRC in EFF's  reverse engineering FAQ), which also don't bypass copyright.  It may give the translator copyright over the translation, but a license from the copyright holder of the original work is still required.

Anyway, AINAL either.  And, more importantly, _I absolutely think we should be able to use that code_ both for compatibility with third-party applications like liquidctl, as well as to repair/fix/extend MSI's software (since software is required by these physical products that are _sold)_.

Tatsh:
Script to fetch the `API_Coreliquid.dll` file:

```shell
#!/usr/bin/env bash
# This script downloads and unpacks API_Coreliquid.dll from MSI.
set -e
url=$(curl 'https://www.msi.com/product_ajax/get_support_item' \
        --data-raw 'support_type=download&type=utility&product_id=234374' \
        --compressed |
    grep -E 'href="https://download.msi.com/.*MSI_Center.*\.zip' |
    sed -r -e 's/.*href="([^"]+)".*/\1/')
original_dir=$(pwd -P)
tmp_dir=$(mktemp -d)
version=$(basename "$url" | sed -r -e 's/\.zip$//' | cut -d_ -f3)

pushd "$tmp_dir" &>/dev/null
if ! [ -f "${HOME}/.cache/msicenter-${version}.zip" ]; then
    wget "${url}" \
        -O "${HOME}/.cache/msicenter-${version}.zip.__download__" || true
    mv "${HOME}/.cache/msicenter-${version}.zip.__download__" \
        "${HOME}/.cache/msicenter-${version}.zip"
fi
unzip -o "${HOME}/.cache/msicenter-${version}.zip"
innoextract 'MSI Center'*.exe
unzip -o app/*.*bundle 'MSI%20Center_*_x64.*x'
unzip -oj MSI%20Center_*_x64.*x DCv2/Package/MSI%20Center%20SDK.exe
innoextract MSI%20Center%20SDK.exe
innoextract tmp/PrePackage/Coreliquid_*.exe
cp -f app/API_Coreliquid.dll "${original_dir}/"
popd &>/dev/null
rm -fR "$tmp_dir"
```

drpars:
Hello;
Do you continue to work on msi coreliquid k360? Will liquidctl have support for k360?

aapo-kossi:
I have this device and am interested in getting this merge through! Do we need more tests or to run current ones? Already cloned Tatsh's fork and am in the process of setting it up for myself.

Tatsh:
> I have this device and am interested in getting this merge through! Do we need more tests or to run current ones? Already cloned Tatsh's fork and am in the process of setting it up for myself.

Please take over so I can close this and delete my copy of the repo.

PR #372, opened 2021-08-20
driver/usb: on Linux, dedupe HID list

Tatsh:
Since hidapi 0.10.1, hidapi's enumerate function returns multiple instances of the same device based on usage and usage page.

See https://github.com/libusb/hidapi/pull/139

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/usb.py
+++ b/liquidctl/driver/usb.py
@@ -474,9 +474,27 @@ class HidapiDevice:
             _LOGGER.debug('sent %d total bytes, expected %d', res, len(data))
         return res
 
+    @staticmethod
+    def dedupe_path(infos):
+        """This is due to a change to hidapi since version 0.10.1 that reports
+        multiple instances of a device with differing usage and usage page
+        keys.
+
+        See https://github.com/libusb/hidapi/pull/139
+        """
+        if sys.platform != 'linux':
+            return infos
+        seen_paths = {}
+        new_infos = []
+        for info in infos:
+            if seen_paths.get(info['path']) is None:
+                seen_paths[info['path']] = True
+                new_infos.append(info)
+        return new_infos
+
     @classmethod
     def enumerate(cls, api, vid=None, pid=None):
-        infos = api.enumerate(vid or 0, pid or 0)
+        infos = HidapiDevice.dedupe_path(api.enumerate(vid or 0, pid or 0))
         if sys.platform == 'darwin':
             infos = sorted(infos, key=lambda info: info['path'])
         for info in infos:


liquidctl/driver/usb.py
```diff
@@ -474,9 +474,28 @@ def send_feature_report(self, data):
...
+        keys.
+
+        See https://github.com/libusb/hidapi/pull/139
+        """
+        if sys.platform != 'linux':
+            return infos
+        seen_paths = {}
+        new_infos = []
+        for info in infos:
+            if info['path'] in seen_paths:
```
MarshallAsch, line 490: perhaps this may be more efficient
```
if seen_paths.get(info['path']) is None:
   seen_paths[info['path']] = True
   new_infos.append(info)
```

MarshallAsch (commented):
Looks good, I just have one comment

liquidctl/driver/usb.py
Tatsh: Fixed.

MarshallAsch:
This looks good thanks for making the change. I have thought of another question, then all is good with me, (we will just wait for @jonasmalacofilho to take a look when he gets a chance. 

This seems like it should be backwards compatible have you been able to check if it works with older versions of hidapi, is this something that should have a test?

jonasmalacofilho:
I'm not sure  this is the correct way to handle multiple usage/usage pages on all platforms.

On Mac (and, IIRC, Windows) each usage/usage page will be associated with a _different_ path, and the correct one must be picked for the usage/usage page of interest.

So far we have handled that in the driver code, specifically [with a custom `probe()` implementation](https://github.com/liquidctl/liquidctl/blob/6295354afb8dcc25fb812ebbf1437a8d51b644ab/liquidctl/driver/rgb_fusion2.py#L107-L126) that only yields a driver instance for the correct usage/usage page pair (or if no such information is available, for compatibility with older versions of hidapi).

---

P.S. I'm sorry for the delay, I had a loss in the family.

Tatsh:
> I'm not sure this is the correct way to handle multiple usage/usage pages on all platforms.

This only affects Linux and only if the paths are the same. Sometimes when there's multiple usages there will be two or more `/dev/hidraw*` devices for a single device, though we don't know of any devices supported by liquidctl that do this. The MSI K360 I no longer have had two usages.

jonasmalacofilho:
> This only affects Linux and only if the paths are the same.  [...]

I could be missing something, but I think you were too short to dismiss my concerns.

Other platforms are affected as well.  And since on those the paths are not the same (because there are separate/different kernel devices for each usage/usage page pair), the fix proposed here is not sufficient.

Tatsh:
Okay. This will not work. I don't get this issue now with the Corsair. Revisit if it comes up again.

Closing.

PR #381, opened 2021-09-14
Added text to liquidcfg.service

Pollomoolokki:
liquidcfg.service needs to have workdir and user specified if installed from pip. Otherwise it fails to start - not finding the module liquidctl.cli

### DIFF
--- a/README.md
+++ b/README.md
@@ -499,7 +499,7 @@ For even more flexibility, you can also write a Python program that calls the dr
 
 ### Set up Linux using systemd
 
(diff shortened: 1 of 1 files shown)


jonasmalacofilho:
Since `pip` is not the ideal installation method on Linux (that's a distro package), `WorkingDirectory` and `User` directives should probably be mentioned (only) in a note.  Additionally, `WorkingDirectory` doesn't cover the virtual environment use case, right?

---

Good catch on the typo.

PR #402, opened 2021-12-16
Adds support for the Corsair H60i Pro XT AIO device

jarrettgilliam:
Adds support for the Corsair H60i Pro XT AIO device.

Related: #387 

---

Checklist:

New CLI flag?

New device?

New driver?


jarrettgilliam:
I was able to confirm everything documented [here](https://github.com/liquidctl/liquidctl/blob/main/docs/corsair-platinum-pro-xt-guide.md) works correctly. There are 16 addressable LEDs on the water block, and none on the fan. The device ships with only 1 fan but 2 fan headers so I left the `fan_count` at 2.

The only thing I wasn't able to get working are the udev rules, but I'm probably doing something wrong there.

Please let me know if I missed anything!

jonasmalacofilho (approved):
Looks good to me, merging. Thanks!

jonasmalacofilho:
> The only thing I wasn't able to get working are the udev rules, but I'm probably doing something wrong there.

Because the udev rules change the permissions of the device in `/dev`, it can sometimes be tricky to get it effectively happen if the device already exists, even if the rule is forcibly triggered.

If you unbind the device from the HID and USB subsystems, reload the rules, and bring the device back, it usually works.  But if you don't mind it, rebooting is usually a simpler (and faster) way to achieve the same thing.

PR #424, opened 2022-02-09
Hacky implementation of color fading for RGB Fusion 2.0

pde:
This is a rather hackish (but fairly efficient) prototype implementation of color fading for the RGB Fusion 2.0, which appears not to inherently support it. Too architecturally ugly to be merged as is, but perhaps interesting anyway. Probably this should be re-implemented within `liquidctl.cli` rather than as a separate front end (but since this runs forever it's rather different, so an argument could be made either way).


jonasmalacofilho:
Since this is a loop, it should indeed be a separate tool, at least with the current CLI design and architecture.

That said, you don't need the hackish hook; instead, use our APIs.  You can take a look at the example/prototypes in `extra/*`.

Additionally, this script should also live in `extra/*`, or maybe `extra/contrib/*` to start since I wouldn't be maintaining it myself.

pde:
Sounds good, I am probably done hacking for today but will come back at some point to switch this to the API.

jonasmalacofilho:
If you have any questions, let me know.

You can also take a look at the docs for the specific driver APIs, if you need information on driver-specific arguments:

```
$ pydoc liquidctl.driver.rgb_fusion2.RgbFusion2
Help on class RgbFusion2 in liquidctl.driver.rgb_fusion2:

liquidctl.driver.rgb_fusion2.RgbFusion2 = class RgbFusion2(liquidctl.driver.usb.UsbHidDriver)
 |  liquidctl.driver.rgb_fusion2.RgbFusion2(device, description, **kwargs)
 |  
 |  liquidctl driver for Gigabyte RGB Fusion 2.0 USB controllers.
 |  
 |  [ ... ]
 |  
 |  set_color(self, channel, mode, colors, speed='normal', **kwargs)
 |      Set the color mode for a specific channel.
 |      
 |      Up to eight individual channels are available, named 'led1' through
 |      'led8'.  In addition to these, the 'sync' channel can be used to apply
 |      the same settings to all channels.
 |      
 |      The table bellow summarizes the available channels.
 |      
 |      | Mode         | Colors required | Speed is customizable |
 |      | ------------ | --------------- | --------------------- |
 |      | off          |            zero |                    no |
 |      | fixed        |             one |                    no |
 |      | pulse        |             one |                   yes |
 |      | flash        |             one |                   yes |
 |      | double-flash |             one |                   yes |
 |      | color-cycle  |            zero |                   yes |
 |      
 |      `colors` should be an iterable of zero or one `[red, blue, green]`
 |      triples, where each red/blue/green component is a value in the range
 |      0–255.
 |      
 |      `speed`, when supported by the `mode`, can be one of: `slowest`,
 |      `slower`, `normal` (default), `faster`, `fastest` or `ludicrous`.
 |  
 |  [ ... ]
```

(Although this particular driver is pretty close to the `BaseDriver`).

pde:
Okay, rewritten against the API and moved to `extra/contrib`.

extra/contrib/fusion_rgb_cycle.py
```diff
@@ -0,0 +1,80 @@
...
+    lookup += cols
+
+
+lookup += reversed(lookup)
+if args.debug:
+    print("Cycling through", lookup)
+while True:
+    for x in range(2 * STEPS):
+        c = lookup[x]
+        with dev.connect():
```
jonasmalacofilho, line 78: You can certainly keep `connect()` out of the inner loop, and maybe even out of the outer `while` loop.

The `connect()` method should only be necessary if you want to allow other programs to talk to the device while you're sleeping (and even then if's only really needed on some platforms, since this particular device is a HID), or if the device disconnects for some external reason (like resuming for suspend-to-memory).

In fact, if the device disconnects for an external reason, it might also be necessary to enumerate it again, with a new call to `find()`.

extra/contrib/fusion_rgb_cycle.py
```diff
@@ -0,0 +1,80 @@
...
+colors = [c.Color("#" + x) for x in args.colors]
+if len(colors) < 2:
+    colors.append(teal)
+if len(colors) < 2:
+    colors.append(bluer)
+
+devs = liquidctl.find_liquidctl_devices()
+for dev in devs:
+    if "Gigabyte RGB Fusion 2.0" in dev.description:
+        break
```
jonasmalacofilho, line 49: If you're looking for a specific device class, you can use `RgbFusion2.find_supported_devices()`, which will only enumerate the required bus and only return devices of the specified class.

If on top of that you still need to do the equivalent of `--match` on the command line, you can pass `match="<prefix>"` to that method (`find_liquidctl_devices()` also supports it).

extra/contrib/fusion_rgb_cycle.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
```
jonasmalacofilho, line 1: Please add a suitable copyright notice and SPDX license header in a comment or module docstring.

It should look something like:

```
Copyright (C) 2022–2022  Peter Eckersley
```

jonasmalacofilho (commented):
I left a few comments, two of them on things that may make your script a bit simpler and/or more efficient.

Also, since this script depends on a library not already in use by liquidctl, can you please have pytest explicitly and unconditionally ignore it in [`conftest.py`](https://github.com/liquidctl/liquidctl/blob/main/conftest.py)?

Otherwise anyone running `python -m pytest` would also need to install the same library.  This should also fix the CI failures.

extra/contrib/fusion_rgb_cycle.py
pde: Okay, for now I'm just going to use the first of these RGBFusion devices but if you have a recommended way to select one if the system has several, we could use that.

extra/contrib/fusion_rgb_cycle.py
pde: I went for the middle option, in case it's more robust. The script was already not producing noticeable load, and this performs 60x fewer `connect()` calls...

jonasmalacofilho (approved):

PR #426, opened 2022-02-14
Documentation & bugfix

pde:
Add documentation for the `fusion_rgb_cycle` script, and fix a bug that would make sequences with more than two colors non-smooth.


docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -29,8 +29,8 @@ Gigabyte RGB Fusion 2.0 5702 Controller
 
 ## Lighting
 
-The controllers support six color modes: `off`, `fixed`, `pulse`, `flash`,
-`double-flash` and `color-cycle`.
+The controllers have built-in support for six color modes: `off`, `fixed`, `pulse`, `flash`, `double-flash` and `color-cycle`. The `color-cycle` mode fades between all
+color hues, but the extra `fusion_rgb_cycle` script can be used to fade between specific colors.
```
jonasmalacofilho, line 33: For a new user, this now contradicts a later sentence:

> The more elaborate color/animation schemes supported by the motherboard on the
addressable headers are not currently supported.

Also, I think the complete path should be used, with a link:

```diff
- `fusion_rgb_cycle`
+ [`extra/contrib/fusion_rgb_cycle.py`]
   [ ... ]
+ [`extra/contrib/fusion_rgb_cycle.py`]: ../extra/contrib/fusion_rgb_cycle.py
```



---

(Nitpick:) in the same paragraph, either hard wrap at ~80 columns (strongly preferred), or don't hard wrap at all.

docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -44,7 +44,7 @@ apply a setting to all lighting channels.
 ```
 # liquidctl set sync color off
 # liquidctl set led1 color fixed 350017
-# liquidctl set led2 color pulse ff2608
+# liquidctl set led2 color pulse
```
jonasmalacofilho, line 47: What's the reason for this change?

docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -86,3 +86,13 @@ Homebrew, along with a script to run on wake that issues the necessary
 liquidctl commands and restores desired lighting effects.
 
 [automate the configuration at boot time]: ../README.md#automation-and-running-at-boot
+
+## Fading between colors
+
+The `extra/contrib/fusion_rgb_cycle.py` script can be used to have one of these controllers cycle through a list of colors of the user's choice. It uses the `fixed` mode but updates the controller continuously. It can be invoked with:
```
jonasmalacofilho, line 92: Link to the script.

Suggestion: s/invoked/started/

docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -86,3 +86,13 @@ Homebrew, along with a script to run on wake that issues the necessary
...
+
+```
+# extra/contrib/fusion_rgb_cycle.py 350017 ff2608
+```
+
+If more two colors are specified, they will be cycled in turn. The [color space](https://facelessuser.github.io/coloraide/colors/) for cycling can be specified, eg `--space hsv`, channels can be specified with eg, `--channel led6` and the time in seconds for each color transition can be set with, eg `--speed 1.2`.
```
jonasmalacofilho, line 98:
```suggestion
If more than two colors are specified, they will be cycled in turn. The [color space] for cycling can be specified with `--space <space>`, channels can be specified with `--channel <channel>` and the time in seconds for each color transition can be set with `--speed <seconds>`.

[color space]: https://facelessuser.github.io/coloraide/colors/
```

extra/contrib/fusion_rgb_cycle.py
```diff
@@ -62,29 +62,28 @@
...
 if args.debug:
-    print("Cycling through", lookup)
+    print("Cycling through", len(lookup), "colors:\n", lookup)
 
 
 while True:
     # reconnect occasionally, just in case these connections die and we can bring them back
     with dev.connect():
-        for x in range(2 * args.steps):
+        for x in range(len(lookup)):
```
jonasmalacofilho, line 86: Possible simplification:

```diff
-        for x in range(len(lookup)):
-            c = lookup[x]
+        for x in lookup:
```

docs/gigabyte-rgb-fusion2-guide.md
pde: Did you mean to remove the hyperlink to the color space docs with this change? There are quite a lot of them, and they produce quite different visual effects, so I think those docs will help some people...

docs/gigabyte-rgb-fusion2-guide.md
pde: Mistake, thanks for catching!

docs/gigabyte-rgb-fusion2-guide.md
pde: My reason for putting it where I did was to give the user a summary of their options for patterns in one place. I think that sentence is still broadly true (for instance the motherboard supports an audio-driven color scheme I think), but I've tweaked it to be clear there :)

pde: (Take a look at the current diff and see if that seems good?)

jonasmalacofilho: That makes sense.  When I first read it really didn't raise any flags, it's just later that I noticed what a new user could find a bit confusing.  But it's good now, thanks for the change.

docs/gigabyte-rgb-fusion2-guide.md
jonasmalacofilho: No, I just used the split link `[link]` and link definition `[link]: url` form.  Mostly because I was already suggesting other adjustments to the paragraph, this document is doing hard line wraps, and a long URL can be a pain for those.

But elsewhere you've consistently used the normal `[link](url)` form, so you can simply ignore it.

---

That said, there's still a `than` missing at the start of the sentence.

docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -29,8 +29,10 @@ Gigabyte RGB Fusion 2.0 5702 Controller
 
 ## Lighting
 
-The controllers have built-in support for six color modes: `off`, `fixed`, `pulse`, `flash`, `double-flash` and `color-cycle`. The `color-cycle` mode fades between all
-color hues, but the extra `fusion_rgb_cycle` script can be used to fade between specific colors.
+The controllers have built-in support for six color modes: `off`, `fixed`, `pulse`,
+`flash`, `double-flash` and `color-cycle`. The `color-cycle` mode fades between all
+color hues, but the extra [fusion_rgb_cycle.py](extra/contrib/fusion_rgb_cycle.py)
```
jonasmalacofilho, line 34: I think the relative address needs to start with `../extra`, since we're in `docs/`, not the root.

docs/gigabyte-rgb-fusion2-guide.md
```diff
@@ -89,10 +91,16 @@ liquidctl commands and restores desired lighting effects.
 
 ## Fading between colors
 
-The `extra/contrib/fusion_rgb_cycle.py` script can be used to have one of these controllers cycle through a list of colors of the user's choice. It uses the `fixed` mode but updates the controller continuously. It can be invoked with:
+The [fusion_rgb_cycle.py](extra/contrib/fusion_rgb_cycle.py) script can be used to have
```
jonasmalacofilho, line 94: Also `../extra`.

extra/contrib/fusion_rgb_cycle.py
```diff
@@ -2,14 +2,21 @@
...
 
-import time
-import sys
+
+try:
+    import coloraide as c
+except ImportError:
+    print("The python coloraide package is not installed...")
+    print("Try installing it with: pip install coloraide")
+    sys.exit(1)
```
jonasmalacofilho, line 18: I really like this is a quality-of-life tweak!

Issue #439, opened 2022-03-07
Aqua Computer D5 Next Pump


TheSkyHeart:
Device type: Pump with and Integrated Controller, one Extra PWM Header and one RGB Header
Product page: 
[D5 NEXT](https://shop.aquacomputer.de/product_info.php?products_id=3785)
First-party software: [AquaSuite](https://aquacomputer.de/software.html)
Other commonly used tools: 

The device appears to support the following features:

 - monitoring of temperatures or other environment sensors: Temp and Flow Sensor, with LCD Display
 - monitoring of fan or pump speeds or duty cycles: Yes
 - monitoring of voltages, currents or power:  Yes
 - configurable fan or pump speeds: Yes
 - configurable voltages or current/power limits: 
 - configurable lighting of embedded LEDs: Yes 
 - configurable lighting of accessories like RGB fans or LED strips: Yes

The device is physically connected to the host system using:

- USB:  it can be connected with normal USB-A using a specific Cable
- PCI-E: 
- onboard the motherboard:  Yes Internal USB header
- other: 

The device uses which protocols to communicate with the host system:

- HID: 
- USB: 
- I²C: 
- SMBus: 
- other: 

Additional information about the device:
as mentioned by @jonasmalacofilho in issue https://github.com/liquidctl/liquidctl/issues/437:
those devices already have have mainline, out-of-tree, or WIP hwmon drivers:
[aquacomputer_d5next](https://www.kernel.org/doc/html/latest/hwmon/aquacomputer_d5next.html)
[(aquacomputer_d5next) Add support for Aquacomputer Farbwerk 360](https://patchwork.kernel.org/project/linux-hwmon/patch/20220227205625.8775-1-savicaleksa83@gmail.com/)

Finally, I can help with:

testing changes on Linux: Yes
testing changes on Windows: Yes
testing changes on Mac OS:
attempting to capture USB traffic: Yes
attempting to capture I²C/SMBus traffic: Yes if it's guided
analyzing traffic data:
documenting the protocol:
implementing the driver:
assuming documentation and/or assistance will be provided.

jonasmalacofilho:
There's now initial monitoring support for the D5 Next in the latest `main` branch.  Please give it a try.

Issue #438, opened 2022-03-07
Aqua Computer OCTO, Quadro, Farbwerk 360


TheSkyHeart:
Device type:  Fan Controller, RGB Controller, Temp Sensor, Flow Sensor
Product page: 
[OCTO](https://shop.aquacomputer.de/product_info.php?products_id=3832)
[Quadro](https://shop.aquacomputer.de/product_info.php?products_id=3773)
[farbwerk 360](https://shop.aquacomputer.de/product_info.php?products_id=3798)
First-party software: AquaSuite
Other commonly used tools:  not tested

The device appears to support the following features:

 - monitoring of temperatures or other environment sensors: Yes, also Flow Sensors
 - monitoring of fan or pump speeds or duty cycles:  Yes
 - monitoring of voltages, currents or power:  Yes
 - configurable fan or pump speeds:  Yes
 - configurable voltages or current/power limits: 
 - configurable lighting of embedded LEDs: Yes
 - configurable lighting of accessories like RGB fans or LED strips: Yes

The device is physically connected to the host system using:

- USB:  it can be connected with normal USB-A using a specific Cable
- PCI-E: 
- onboard the motherboard: Yes Internal USB header
- other: 

The device uses which protocols to communicate with the host system:

- HID: 
- USB: 
- I²C: 
- SMBus: 
- other: 

Additional information about the device:

as mentioned by @jonasmalacofilho in issue #437:
those devices already have have mainline, out-of-tree, or WIP hwmon drivers:
[aquacomputer_octo-hwmon](https://github.com/detrophy/aquacomputer_octo-hwmon)
[aquacomputer_quadro-hwmon](https://github.com/leoratte/aquacomputer_quadro-hwmon)

Finally, I can help with:

- testing changes on Linux:  Yes
- testing changes on Windows:  Yes
- testing changes on Mac OS: 
- attempting to capture USB traffic:  Yes
- attempting to capture I²C/SMBus traffic: Yes if it's guided 
- analyzing traffic data: 
- documenting the protocol: 
- implementing the driver: 

assuming documentation and/or assistance will be provided.

aleksamagicka:
Hi! Author of the aquacomputer_d5next driver here. Octo support is currently in process of being reviewed/in iteration for mainline, and a Quadro is on its way to me. I'll look at the code in this repo (not necessarily immediately, because college) and see what I can do. 😄 Same goes for the Farbwerk 360, Octo, Quadro, since I know most of their usable register offsets.

[...]

PR #482, opened 2022-07-23
Aquacomputer D5 Next sensor reading support

aleksamagicka:
As promised in #438, here is the first PR for adding support for Aquacomputer devices to liquidctl. This PR adds a new driver called `aquacomputer`, and it currently supports reading sensor values of the D5 Next watercooling pump.

Aquacomputer devices share same layouts of sensor substructures/groups in their HID reports, so adding sensor reading support for other devices, such as Octo, Quadro, Farbwerk, Farbwerk 360 should be trivial. (Just need to specify the offsets in device specific data and write some cases for non-general sensors.) They are all supported by the [partially mainlined driver](https://github.com/aleksamagicka/aquacomputer_d5next-hwmon) that I maintain.

Some advanced features of the devices are still being investigated and reverse engineered, and additionally as far as I know nobody has looked at the RGB controls yet. The kernel driver supports sensor reading and directly controlling the PWM inputs, where available.

Marking this PR as draft for now, as I have to check more boxes down below. I've formatted the code using black (haven't used it before), so hopefully it looks as it should.

And a question for the future: the kernel driver support for devices varies between kernel versions. Is there a clean way to get the kernel version from liquidctl already?

Related: #439 

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The notes are sorted alphabetically, major (upper case) notes before minor
 | AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_p_</sup> |
 | DDR4 DRAM          | [Corsair Vengeance RGB](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
 | DDR4 DRAM          | [Generic DDR4 temperature sensor](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
-| Fan/LED controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_ehn_</sup> |
+| Fan/pump controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_ehn_</sup> |
 | Fan/LED controller | [Corsair Commander Pro](docs/corsair-commander-guide.md) | USB HID | <sup>_h_</sup> |
 | Fan/LED controller | [Corsair Commander Core](docs/corsair-commander-core-guide.md) | USB HID | <sup>_ep_</sup> |
 | Fan/LED controller | [Corsair Commander Core XT](docs/corsair-commander-core-guide.md) | USB HID | <sup>_enp_</sup> |
@@ -136,7 +136,7 @@ The notes are sorted alphabetically, major (upper case) notes before minor
 | Fan/LED controller | [NZXT RGB & Fan Controller](docs/nzxt-hue2-guide.md) | USB HID | <sup>_h_</sup> |
 | Fan/LED controller | [NZXT Smart Device](docs/nzxt-smart-device-v1-guide.md) | USB HID | <sup>_h_</sup> |
 | Fan/LED controller | [NZXT Smart Device V2](docs/nzxt-hue2-guide.md) | USB HID | <sup>_h_</sup> |
-| Fan/AIO controller | [NZXT H1 V2](docs/nzxt-hue2-guide.md) | USB HID | <sup>_e_</sup> |
+| Fan/pump controller | [NZXT H1 V2](docs/nzxt-hue2-guide.md) | USB HID | <sup>_e_</sup> |
 | Graphics card RGB  | [Select ASUS GTX and RTX cards](docs/nvidia-guide.md) | I²C | <sup>_Ux_</sup> |
 | Graphics card RGB  | [Select EVGA GTX 1070, 1070 Ti and 1080 cards](docs/nvidia-guide.md) | I²C | <sup>_Ux_</sup> |
 | Motherboard        | [ASUS Aura LED motherboards](docs/asus-aura-led-guide.md) | USB HID | <sup>_e_</sup> |
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ The notes are sorted alphabetically, major (upper case) notes before minor
 | AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_p_</sup> |
 | DDR4 DRAM          | [Corsair Vengeance RGB](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
 | DDR4 DRAM          | [Generic DDR4 temperature sensor](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
+| Fan/LED controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_hn_</sup> |
 | Fan/LED controller | [Corsair Commander Pro](docs/corsair-commander-guide.md) | USB HID | <sup>_h_</sup> |
 | Fan/LED controller | [Corsair Commander Core](docs/corsair-commander-core-guide.md) | USB HID | <sup>_ep_</sup> |
 | Fan/LED controller | [Corsair Commander Core XT](docs/corsair-commander-core-guide.md) | USB HID | <sup>_enp_</sup> |
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The notes are sorted alphabetically, major (upper case) notes before minor
 | AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_p_</sup> |
 | DDR4 DRAM          | [Corsair Vengeance RGB](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
 | DDR4 DRAM          | [Generic DDR4 temperature sensor](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
-| Fan/LED controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_hn_</sup> |
+| Fan/LED controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_ehn_</sup> |
 | Fan/LED controller | [Corsair Commander Pro](docs/corsair-commander-guide.md) | USB HID | <sup>_h_</sup> |
 | Fan/LED controller | [Corsair Commander Core](docs/corsair-commander-core-guide.md) | USB HID | <sup>_ep_</sup> |
 | Fan/LED controller | [Corsair Commander Core XT](docs/corsair-commander-core-guide.md) | USB HID | <sup>_enp_</sup> |
--- a/docs/aquacomputer-d5next-guide.md
+++ b/docs/aquacomputer-d5next-guide.md
@@ -3,12 +3,18 @@ _Driver API and source code available in [`liquidctl.driver.aquacomputer`](../li
 
 ## Initialization
 
-Initialization is not required. The pump sends a status HID report every second as soon as it's connected.
+Initialization is _currently_ not required, but is recommended. It outputs the firmware version:
+
+```
+Aquacomputer D5 Next
+└── Firmware version    1023
+```
+
+The pump automatically sends a status HID report every second as soon as it's connected.
 
 ## Monitoring
 
-The D5 Next exposes sensor values such as liquid temperature and two groups of fan sensors, for the pump and the optionally connected fan. These groups provide RPM speed, voltage, current and power readings. The
-pump additionally exposes +5V and +12V voltage rail readings.
+The D5 Next exposes sensor values such as liquid temperature and two groups of fan sensors, for the pump and the optionally connected fan. These groups provide RPM speed, voltage, current and power readings. The pump additionally exposes +5V and +12V voltage rail readings:
 
 ```
 # liquidctl status
--- a/docs/aquacomputer-d5next-guide.md
+++ b/docs/aquacomputer-d5next-guide.md
@@ -7,7 +7,8 @@ Initialization is _currently_ not required, but is recommended. It outputs the f
 
 ```
 Aquacomputer D5 Next
-└── Firmware version    1023
+├── Firmware version           1023
+└── Serial number       03500-24905
 ```
 
 The pump automatically sends a status HID report every second as soon as it's connected.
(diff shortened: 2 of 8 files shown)


jonasmalacofilho:
Awesome, thanks for working on this!

> Some advanced features of the devices are still being investigated and reverse engineered, and additionally as far as I know nobody has looked at the RGB controls yet.

Small note: in this case, I think it'd be more appropriate to raise `NotSupportedByDriver`.

> And a question for the future: the kernel driver support for devices varies between kernel versions. Is there a clean way to get the kernel version from liquidctl already?

Not yet.

Also, while getting the kernel version should be relatively straightforward, that wouldn't tell us the full picture, as the user could be running an out-of-tree driver.

Do you know how we might get information about the module?

---

I haven't done a proper review yet, but at first glance the patch looks very looks good.

aleksamagicka:
> Small note: in this case, I think it'd be more appropriate to raise NotSupportedByDriver.

Thanks. `NotSupportedByDriver` sound better, I wasn't sure exactly what exception to raise while I was writing it. I'll explore the code more.

> Also, while getting the kernel version should be relatively straightforward, that wouldn't tell us the full picture, as the user could be running an out-of-tree driver.

Yes, that's a concern as well. I'm not yet that intimate with liquidctl, but if I'm not mistaken I saw some drivers accepting custom CLI flags. Maybe that could be used to override assumptions the code would make about the kernel version it detects. I'll have to explore and think about it more, because there's features piling up at the kernel driver repo that I have yet to send in. But purely for the sensor stuff for now, it should be a clear cut of what kernel supports what. I have a table in the [readme](https://github.com/aleksamagicka/aquacomputer_d5next-hwmon#overview) to track that, aside from WIP branches for some other devices that I'm not going to send here just yet.

jonasmalacofilho:
> I'm not yet that intimate with liquidctl, but if I'm not mistaken I saw some drivers accepting custom CLI flags. Maybe that could be used to override assumptions the code would make about the kernel version it detects.

I think that could get quite messy quite fast.

A think a combination of us implementing graceful degradation and the user having the option to use `--direct-access` might be enough, and better than a ton of CLI options.

(On a related note, for super hack level or only-enable-if-you-known-what-you're-doing options we also have `--unsafe` features and environment variables).

liquidctl/driver/aquacomputer.py
```diff
@@ -0,0 +1,258 @@
...
+            )
+            sensor_readings.append(fan_current)
+
+        # Special-case sensor readings
+        if self._device_info["type"] == self.DEVICE_D5NEXT:
+            # Read +5V voltage rail value
+            plus_5v_voltage = ("+5V voltage", self._hwmon.get_int("in2_input") * 1e-3, "V")
+            sensor_readings.append(plus_5v_voltage)
+
+            if self._hwmon.has_attribute("in3_input"):
```
aleksamagicka, line 211: Should a warning be produced if this is not available?

README.md
```diff
@@ -125,6 +125,7 @@ The notes are sorted alphabetically, major (upper case) notes before minor
 | AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_p_</sup> |
 | DDR4 DRAM          | [Corsair Vengeance RGB](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
 | DDR4 DRAM          | [Generic DDR4 temperature sensor](docs/ddr4-guide.md) | SMBus | <sup>_Uax_</sup> |
+| Fan/LED controller | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_ehn_</sup> |
```
aleksamagicka, line 128: It's category is more of a "watercooling pump/fan controller", since it only controls it's own LEDs... Is the existing text OK or should I write it differently?

liquidctl/driver/aquacomputer.py
jonasmalacofilho: Yeah, I think so.  That attribute is useful, and our other drivers emit warnings in similar situations:

```py
_LOGGER.warning('some attributes cannot be read from %s kernel driver', self._hwmon.module)
```

I don't think the warning needs to be more specific than that, even though here it's just one attribute.  In the future there might be more (or at least different) attributes missing, and I don't think most users need to care about the specifics (while this is an useful attribute, everything else is still quite functional without it).

That said, one downside of the non-specific warning is that users will need to check the documentation and/or try `--direct-access` just to see what they might be missing.  But that's a general problem with all of our (hwmon-aware) drivers, so this probably isn't the best place to consider it.

README.md
jonasmalacofilho: I think "Fan/pump controller" might be a reasonably appropriate and short name, and it can also replace the "Fan/AIO controller" category used for the NZXT H1 V2.

jonasmalacofilho:
Just saw that you marked the PR as ready for review, thanks!  I'll try to do it this weekend.

README.md
aleksamagicka: Thanks, marked both as that.

liquidctl/driver/aquacomputer.py
aleksamagicka: Added that.

liquidctl/driver/aquacomputer.py
```diff
@@ -0,0 +1,262 @@
...
+_AQC_FAN_VOLTAGE_OFFSET = 0x02
+_AQC_FAN_CURRENT_OFFSET = 0x04
+_AQC_FAN_POWER_OFFSET = 0x06
+_AQC_FAN_SPEED_OFFSET = 0x08
+
+_AQC_STATUS_READ_ENDPOINT = 0x01
+
+
+def get_unaligned_be16(msg, offset):
+    return msg[offset] << 8 | msg[offset + 1]
```
jonasmalacofilho, line 43: While this obviously works, I think the Python way is to use `int.from_bytes`.  It's a lot more verbose, but for that we already have a `liquidctl.util.u16be_from` wrapper.

liquidctl/driver/aquacomputer.py
```diff
@@ -0,0 +1,262 @@
...
+            "plus_5v_voltage": 0x39,
+            "plus_12v_voltage": 0x37,
+            "temp_sensors_label": ["Liquid temperature"],
+            "fan_speed_label": ["Pump speed", "Fan speed"],
+            "fan_power_label": ["Pump power", "Fan power"],
+            "fan_voltage_label": ["Pump voltage", "Fan voltage"],
+            "fan_current_label": ["Pump current", "Fan current"],
+            "status_report_length": 0x9E,
+        }
+    }
```
jonasmalacofilho, line 65: While `SUPPORTED_DEVICES` is public, that's a mistake from liquidctl 1.0.[^1]   So I think it's better to keep these new constants private, even if they're observable through `SUPPORTED_DEVICES`.

By the way, I really liked how clean these made the driver setup!

[^1]: Unfortunately there used to be third-party code that depended on `SUPPORTED_DEVICES`.  It seems like some of it now only uses our _intended_ public interface, so maybe I can finally look into making it private.

liquidctl/driver/aquacomputer.py
```diff
@@ -0,0 +1,262 @@
...
+        a suspended state, or if the device has just been (re)connected.  In
+        those scenarios, no other method, except `connect()` or `disconnect()`,
+        should be called until the device and driver has been (re-)initialized.
+
+        Returns None or a list of `(property, value, unit)` tuples, similarly
+        to `get_status()`.
+        """
+
+        fw = self.firmware_version
+        _LOGGER.debug("raw firmware version: %r", fw)
```
jonasmalacofilho, line 92: If the raw firmware version is identical to our human-readable output, it isn't necessary to log it.

docs/aquacomputer-d5next-guide.md
```diff
@@ -0,0 +1,27 @@
+# Aquacomputer D5 Next watercooling pump
+_Driver API and source code available in [`liquidctl.driver.aquacomputer`](../liquidctl/driver/aquacomputer.py)._
+
+## Initialization
+
+Initialization is not required. The pump sends a status HID report every second as soon as it's connected.
```
jonasmalacofilho, line 6: If there's a chance initialization will be necessary for future devices (and there probably is), it's better to continue to recommend that the device be initialized, and only note that _currently_ it _might_ be a no-op.

Additionally, since `initialize` outputs the firmware version, there already is value in calling/executing it (especially when reporting bugs to us).  For the same reason, it also makes sense to show an example of it here as well.

liquidctl/driver/aquacomputer.py
```diff
@@ -0,0 +1,262 @@
...
+
+        return [("Firmware version", fw, "")]
+
+    def _get_status_directly(self):
+        msg = self._read()
+
+        sensor_readings = []
+
+        # Read temp sensor values
+        for idx, temp_sensor_offset in enumerate(self._device_info["temp_sensors"]):
```
jonasmalacofilho, line 102: This is probably just as easy to read as is, if not more, but a reminder that you could also do something like:

```suggestion
        for label, offset in zip(self._device_info["temp_sensors_label"], self._device_info["temp_sensors"]):
```

...here and in the similar loops bellow.

tests/test_aquacomputer.py
```diff
@@ -0,0 +1,135 @@
...
+    with pytest.raises(NotSupportedByDriver):
+        mockD5NextDevice.set_fixed_speed("pump", 84)
+
+
+def test_kraken_speed_profiles_not_supported(mockD5NextDevice):
+    with pytest.raises(NotSupportedByDriver):
+        mockD5NextDevice.set_speed_profile("fan", 1)
+
+    with pytest.raises(NotSupportedByDriver):
+        mockD5NextDevice.set_speed_profile("pump", 1)
```
jonasmalacofilho, line 135: A `None` or empty list/tuple are a bit less unexpected than an int as placeholders for the profile parameter.  It doesn't matter a lot, but personally I did notice the weird int and stopped for a second to check what was going on.

jonasmalacofilho (commented):
Hi,

I really appreciate the attention to detail, the PR is super clean.  I left a few notes, but they are very minor.

Regarding the remaining applicable items in the checklist, you can leave updating the man page to me, as for now that'll just be a placeholder, and I already need to add those to some other devices.

But can you also try to document the protocol (as much as currently implemented by the driver) in `docs/developer/protocol/*`?

liquidctl/driver/aquacomputer.py
aleksamagicka: Agreed, `u16be_from` is better.

liquidctl/driver/aquacomputer.py
aleksamagicka: Thanks, I've been aware of zip() before, but I never had an opportunity to use it (in Python or elsewhere). I converted that loop to use it.

aleksamagicka:
Stuff you commented on should now be resolved, please take a look.

I'll add a protocol description later during the day, hopefully. I've written about it [before](https://github.com/aleksamagicka/aquacomputer_d5next-hwmon/blob/main/re-docs/d5next.md), so I'll use it to speed up the process as it's good enough for the sensor report aspect. (The control/configuration report has been reverse engineered a bit more throughly, and that needs to be updated.)

liquidctl/driver/aquacomputer.py
jonasmalacofilho: Let me start by saying sorry for the noise!

I think it's (more) important to be consistent throughout the file.  And while it's possible to use zip in the other loops too, it'll probably look a bit unwieldy when more than one label is needed per iteration.  Honestly, even in this loop, I think I prefer the look of your original version, thanks to the shorter and more uniform lines.

I also thought that there could be a performance penalty in doing the extra indexed (and checked) access, but I can't actually see that in a quick microbenchmark, at least not with Python 3.10.

So I might revert this back before merging the PR, unless you want to try using zip on the other loops.

Again, sorry!

jonasmalacofilho:
> Stuff you commented on should now be resolved, please take a look.

Thank you!

> I'll add a protocol description later during the day, hopefully. I've written about it [before](https://github.com/aleksamagicka/aquacomputer_d5next-hwmon/blob/main/re-docs/d5next.md), so I'll use it to speed up the process as it's good enough for the sensor report aspect. (The control/configuration report has been reverse engineered a bit more throughly, and that needs to be updated.)

Yes, that's good enough, thanks!

liquidctl/driver/aquacomputer.py
aleksamagicka: No problem, I'll revert it now to make it easier.

aleksamagicka: Just saw the edit - off the top of my head there will be 8-10 max register offsets (for Octo), so we're good.

aleksamagicka:
I've pushed the docs and added serial number retrieval to initialization. I've made the protocol docs general, because the devices are similar and I'll be extending it in future PRs once this one is merged.

jonasmalacofilho:
Merged, thank you!

> I've made the protocol docs general, because the devices are similar and I'll be extending it in future PRs once this one is merged.

Yeah, that makes sense.

jonasmalacofilho:
Err, I just noticed an issue.

You're overriding the `serial_number` property, and that will cause a HID read during probe if a `--serial` filter is used (and you have one of these devices).

Probing is not supposed to result in any reads or writes, and in the case of USB HIDs (and other USB devices), that properly should only return the USB `iSerialNumber` descriptor.

---

Related to it, there something else I noticed but had so far left for later:

Unless you end up (in the next devices) needing the firmware version and serial number outside of `initialize`, it's probably better to not parse them on every `_read`.

We also currently don't expose a general firmware version property for all devices.  While it's fine to do have it (when, like here, it' seems stable and cheap to compute), it's also not necessary.

aleksamagicka:
> You're overriding the serial_number property, and that will cause a HID read during probe if a --serial filter is used (and you have one of these devices).

Sorry, that was accidental, I didn't realize there was an existing function with that same name. Is it OK to rename it so the true one is not overriden?

> Unless you end up (in the next devices) needing the firmware version and serial number outside of initialize, it's probably better to not parse them on every _read.

Makes sense, I'll move parsing to their respective functions.

jonasmalacofilho:
> Sorry, that was accidental, I didn't realize there was an existing function with that same name.

Yeah, I missed it too.

>  Is it OK to rename it so the true one is not overriden?

In the case of the serial number I think it's probably better to not have it as public property.

IIRC we don't have it on any other drivers, and for those specifically interested in it, it would be a way to bypass calling `initialize()`, which might not be ideal in the long term.

(Besides `initialize` potentially becoming necessary at some point, we also explicitly document that the specific keys and values returned from it are _not_ stable, and that's important because we can't guarantee that the firmware wont change these things).

PR #489, opened 2022-08-04
Aquacomputer driver enhancements

aleksamagicka:
For `aquacomputer.py`, make serial number property private and read that and firmware version only when they are requested, but not already cached.

Related: #482

---
Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/aquacomputer.py
+++ b/liquidctl/driver/aquacomputer.py
@@ -76,7 +76,7 @@ class Aquacomputer(UsbHidDriver):
 
         # Read when necessary
         self._firmware_version = None
-        self._serial_number = None
+        self._serial = None
 
         self._device_info = device_info
 
@@ -93,7 +93,7 @@ class Aquacomputer(UsbHidDriver):
         """
 
         fw = self.firmware_version
-        serial_number = self.serial_number
+        serial_number = self._serial_number
 
         return [("Firmware version", fw, ""), ("Serial number", serial_number, "")]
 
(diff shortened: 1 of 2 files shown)


aleksamagicka:
As far as I can see I didn't change anything that would upset the keyval test...

jonasmalacofilho:
> As far as I can see I didn't change anything that would upset the keyval test...

I think we just experienced a corruption of the Python interpreter.

The assertion failure, if I read it correctly, was caused by in that specific run a method call
```py
assert store.load(...) == 2
```
somehow became a method reference, turning the assertion in memory to something like
```py
assert store.load == 2
```
which obviously is false.

---

Anyway, I restarted the job and then it completed fine.

liquidctl/driver/aquacomputer.py
```diff
@@ -93,7 +93,7 @@ def initialize(self, **kwargs):
         """
 
         fw = self.firmware_version
-        serial_number = self.serial_number
+        serial_number = self._serial_number
```
jonasmalacofilho, line 96: Shouldn't this be `self._serial`?  And how isn't this tripping the tests?

---

Also, needing two reads in the normal path (which is to call `initialize` first) isn't ideal, even if the interval between reports is fairly short.

I think it would be better if the properties used a shared `_read_device_statics` (or something like it) method to parses both fields from a single `_read`.

aleksamagicka: `self._serial` is where the serial number gets stored from within the `self._serial_number` property (currently), it would be empty on its own.

jonasmalacofilho: Oh, yeah, of course!

jonasmalacofilho:
Thanks!

PR #499, opened 2022-08-28
Add support for setting fan speed for D5 Next

aleksamagicka:
Add support for setting pump and fan speed for the D5 Next pump in the `aquacomputer.py` driver.

I had to use the `crcmod` library to calculate CRC-16/USB checksums, because the control HID report (that contains all the settings for a device) has one at the end and it needs to be updated when the settings are changed. Please let me know if I missed adding it somewhere.

Also I wanted to ask if it's OK to utilize the hwmon driver for writing stuff to the device. I don't really see than being done in other drivers, so maybe `--use-hwmon` can be passed in here to let the user go through the kernel driver?

Related: #439 

---

Checklist:

New CLI flag?

New device?

New driver?


jonasmalacofilho:
Sorry again for the delay.

> Also I wanted to ask if it's OK to utilize the hwmon driver for writing stuff to the device.

Yes, and it should actually be preferred (when there's a hwmon driver).

> I don't really see than being done in other drivers,

That's mostly just because hwmon support in liquidctl is new and not yet complete, but also because not all current hwmon drivers expose PWM control (e.g. Kraken X2, because of device/protocol issues, or Corsair RMi/HXi, due to safety concerns, IIRC).

> so maybe `--use-hwmon` can be passed in here to let the user go through the kernel driver?

No, I think it should be the other way around, just as with `get_status()`: the hwmon driver is used by default, unless `--direct-access` is passed.

Besides being more consistent, this minimizes the chances of putting either the device or the kernel driver in a inconsistent/unexpected state.

aleksamagicka:
Great, thanks. I'll rework it to use hwmon first, if possible.

aleksamagicka:
Writing to hwmon is now implemented in the aquacomputer driver, please take a look.

docs/aquacomputer-d5next-guide.md
```diff
@@ -36,6 +38,20 @@ Aquacomputer D5 Next
...
+
+```
+liquidctl set pump speed 56
+              ^^^^       ^^
+             channel    duty
+```
+
+Valid channel values on the D5 Next are `pump` and `fan`. The hwmon driver is always used first, if it's available and
+exposes `pwmX` and `pwmX_enable` attributes. You can pass in `--direct-access` to bypass it and communicate with the
+device directly.
```
jonasmalacofilho, line 53: I don't think it's necessary to talk about hwmon here, especially since the very next section is about it (and how to bypass it).

liquidctl.8
```diff
@@ -236,7 +236,7 @@ Internal data used by some drivers.
 .SS Aquacomputer Farbwerk 360
 .SS Aquacomputer Octo
 .SS Aquacomputer Quadro
-Cooling channels: (D5 Next, Octo, Quadro:) not yet supported; (Farbwerk 360:) not applicable.
+Cooling channels: (D5 Next): supported, (Octo, Quadro:) not yet supported; (Farbwerk 360:) not applicable.
```
jonasmalacofilho, line 239:
```suggestion
Cooling channels: (D5 Next): \fIfan\fR, \fIpump\fR; (Octo, Quadro:) not yet supported; (Farbwerk 360:) not applicable.
```

liquidctl/driver/aquacomputer.py
```diff
@@ -340,17 +354,79 @@ def set_speed_profile(self, channel, profile, **kwargs):
...
             raise NotSupportedByDevice()
 
+        # Clamp duty between 0 and 100
+        duty = clamp(duty, 0, 100)
+
+        if self._hwmon and not direct_access:
+            _LOGGER.info(
+                "bound to %s kernel driver, writing fixed speed to hwmon", self._hwmon.driver
+            )
+            return self._set_fixed_speed_hwmon(channel, duty)
```
jonasmalacofilho, line 421: This will fail if the hwmon driver is present, but doesn't expose the necessary PWM attributes.

I think we should instead automatically fallback to direct access in that case.

liquidctl/driver/aquacomputer.py
```diff
@@ -340,17 +354,79 @@ def set_speed_profile(self, channel, profile, **kwargs):
...
+
+        if self._hwmon:
+            _LOGGER.warning(
+                "directly writing fixed speed despite %s kernel driver", self._hwmon.driver
+            )
```
jonasmalacofilho, line 426: Related to the previous comment, in my opinion a warning is only necessary if we're re-implementing (and possibly conflicting with) something the kernel driver actually supports/exposes.

liquidctl/driver/hwmon.py
```diff
@@ -35,6 +35,9 @@ def get_string(self, name):
     def get_int(self, name):
         return int(self.get_string(name))
 
+    def write_str(self, name, value):
+        (self.path / name).write_text(value)
```
jonasmalacofilho, line 39: For symmetry, I think a `write_int` API would be better, and move the `str()` conversion within it.

On a related note, it would also probably make sense to rename `get_int` to `read_int`, also for symmetry.

tests/test_aquacomputer.py
```diff
@@ -154,12 +184,32 @@ def test_d5next_get_status_from_hwmon(mockD5NextDevice, tmp_path):
     assert sorted(got) == sorted(expected)
 
 
-def test_d5next_set_fixed_speeds_not_supported(mockD5NextDevice):
-    with pytest.raises(NotSupportedByDriver):
-        mockD5NextDevice.set_fixed_speed("fan", 42)
+def test_d5next_set_fixed_speeds_directly(mockD5NextDevice):
```
jonasmalacofilho, line 187: Similarly to `test_d5next_get_status_directly`, this should also test whether `--direct-access` is honored, even if a (complete) hwmon driver is present.

But I do see how this might be getting a bit repetitive/annoying.

tests/test_aquacomputer.py
```diff
@@ -154,12 +184,32 @@ def test_d5next_get_status_from_hwmon(mockD5NextDevice, tmp_path):
...
-        mockD5NextDevice.set_fixed_speed("pump", 84)
+    fan_report, pump_report = mockD5NextDevice.device.sent
+
+    assert fan_report.number == 3
+    assert fan_report.data[0x40:0x43] == [0, 19, 136]  # 0, <5000>
+    assert pump_report.number == 3
+    assert pump_report.data[0x95:0x98] == [0, 32, 208]  # 0, <8400>
+
+
+def test_d5next_set_fixed_speeds_hwmon(mockD5NextDevice, tmp_path):
```
jonasmalacofilho, line 199: It might be wise to also test whether the fallback to direct mode when the PWM attributes are missing.

jonasmalacofilho (commented):
Thanks!

jonasmalacofilho:
Since crcmod is stable but not actively maintained, let's exceptionally pin its required version to 1.7.

```
# on setup.cfg, .github/workflows/*.yml, README.md, etc:
crcmod == 1.7
```

I usually don't do this type of thing in order to avoid issues with distro-supplied packages.  But in this case we really don't expect to see a new version, and I'm not super comfortable leaving a highly popular but _potentially[^1]_ neglected dependency unpinned.

[^1]: I don't know the crcmod developers.  It's perfectly possible that they are active and super careful, and that crcmod simply hasn't needed any change in the last decade.  However, I couldn't immediately assert that's indeed the case, and pinning this one dependency should be nearly painless... after all, it hasn't received any _intentional_ updates in more than a decade.

liquidctl/driver/hwmon.py
aleksamagicka: Since renaming `get_int` to `read_int` touches more drivers than `aquacomputer.py`, do you want me to create a separate PR right after this one or do it here?

aleksamagicka:
Thanks for the review, things you mentioned should be fixed now, except the `get_int` to `read_int` conversion.

liquidctl/driver/hwmon.py
jonasmalacofilho: You're right, that would be better.

aleksamagicka: Will do!

jonasmalacofilho:
Thanks!

Issue #464, opened 2022-06-04
Corsair Commander Core giving error: 'response does not match command'


Grooty12:
I have just gotten the Corsair H100i Capellix liquid cooler, and when i try to use any command with it, this output is given: `ERROR: unexpected error with Corsair Commander Core (experimental): AssertionError('response does not match command')`

Basically all commands used gives the error (initialize, status, set). The expected behavior is that they work, but for all of them this happens:
```
$ liquidctl --match core status
ERROR: unexpected error with Corsair Commander Core (experimental): AssertionError('response does not match command')
```

Here is the output of the --debug flag:
```
$ liquidctl --match core status --debug                                                                                                     1 ✘ 
[DEBUG] (cli) (main): liquidctl: 1.9.1
[DEBUG] (cli) (main): platform: Linux-5.18.1-arch1-1-x86_64-with-glibc2.35
[DEBUG] (cli) (_log_requirements): python: 3.10.4 (main, May 14 2022, 05:21:19) [GCC 12.1.0]
[DEBUG] (cli) (_log_requirements): colorlog: 6.5.0
[DEBUG] (cli) (_log_requirements): docopt: 0.6.2
[DEBUG] (cli) (_log_requirements): hidapi: 0.11.2
[DEBUG] (cli) (_log_requirements): pyusb: 1.2.1
[DEBUG] (cli) (_log_requirements): smbus: 1.1
[DEBUG] (cli) (_log_requirements): libusb-package: version n/a (No package metadata was found for libusb-package)
[DEBUG] (smbus) (find_devices): searching LinuxI2c
[DEBUG] (smbus) (find_devices): LinuxI2c drivers: Ddr4Temperature, EvgaPascal, RogTuring, VengeanceRgb
[DEBUG] (smbus) (find_devices): I²C adapter: i2c-3
[DEBUG] (smbus) (find_devices): I²C adapter: i2c-1
[DEBUG] (smbus) (find_devices): I²C adapter: 0-0037 (skipped, not a bus or unsupported adapter)
[DEBUG] (smbus) (find_devices): I²C adapter: i2c-6
[DEBUG] (smbus) (find_devices): I²C adapter: 0-0053 (skipped, not a bus or unsupported adapter)
[DEBUG] (smbus) (find_devices): I²C adapter: i2c-4
[DEBUG] (smbus) (find_devices): I²C adapter: 0-0051 (skipped, not a bus or unsupported adapter)
[...]

PR #501, opened 2022-08-30
Commander core update

Meriemi-git:
Add support for the last Corsair Commander Core firmware v2.10.219.
I follow the explanation of ParkerMC to fix the issue, I changed the flow inside the _read_data and _write_data functions to follow the flow described by ParkerMC : set_mode > read / write > reset

I changed the read operation inside _send_command function to read until response starts with 0x00.

After changes to the command flow the tests failed with an "Key Error" at line 117 of file test_commander_core.py.
This is because when i delayed de reset command in read_data and when we start with a set_mode command the _modes array of the Mock device is empty. To prevent this i set in _init_ a None value inside the _modes array.

Fixes: [#464](https://github.com/liquidctl/liquidctl/issues/464)

---

Checklist:

- [x] Submit relevant data, scripts or dissectors to https://github.com/liquidctl/collected-device-data

New CLI flag?

- [] Adjust the completion scripts in `extra/completions/`

New device?

New driver?

### DIFF
--- a/tests/test_commander_core.py
+++ b/tests/test_commander_core.py
@@ -31,7 +31,7 @@ class MockCommanderCoreDevice:
 
         self._last_write = bytes()
         self._modes = {}
-
+        self._is_reset = False
         self._awake = False
 
         self.response_prefix = ()
@@ -52,7 +52,7 @@ class MockCommanderCoreDevice:
         if self._awake:
             if self._last_write[2] == 0x08:  # Get data
                 channel = self._last_write[3]
-                mode = self._modes[channel]
+                mode = self._modes.get(channel)
                 if mode[1] == 0x00:
                     if mode[0] == 0x17:  # Get speeds
                         data.extend([0x06, 0x00])
@@ -114,16 +114,20 @@ class MockCommanderCoreDevice:
 
         if data[2] == 0x0d:
             channel = data[3]
-            if self._modes[channel] is None:
+            if  (self._is_reset or not self._awake) and self._modes.get(channel) is None:
                 self._modes[channel] = data[4:6]
+                self._is_reset = False
+            else:
+                raise ExpectationNotMet('Previous channel was not reset')
         elif data[2] == 0x05 and data[3] == 0x01:
             self._modes[data[4]] = None
+            self._is_reset = True
         elif data[2] == 0x01 and data[3] == 0x03 and data[4] == 0x00:
             self._awake = data[5] == 0x02
         elif self._awake:
             if data[2] == 0x06:  # Write command
                 channel = data[3]
-                mode = self._modes[channel]
+                mode = self._modes.get(channel)
                 length = u16le_from(data[4:6])
                 data_type = data[8:10]
                 written_data = data[10:8+length]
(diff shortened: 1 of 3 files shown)


Meriemi-git:
I have test this on my Corsair H100i with firmware v.2.10.219, it works but i'm not very confident about the few modifications of the unit tests.

tests/test_commander_core.py
```diff
@@ -30,7 +30,7 @@ def __init__(self):
         self.sent = list()
 
         self._last_write = bytes()
-        self._modes = {}
+        self._modes = [None]
```
jonasmalacofilho, line 33: To me it _seems_ that the issue is that 

```py
# In MockCommanderCoreDevice.write():
if self._modes[channel] is None:
```

should have always been

```py
if self._modes.get(channel) is None:
```

What do you think, @ParkerMc?

---

Either way, initializing a dict with a list doesn't feel right.

.gitignore
```diff
@@ -5,4 +5,5 @@
 /dist/
 __pycache__/
 liquidctl/_version.py
-.DS_Store
\ No newline at end of file
+.DS_Store
+.vscode
```
jonasmalacofilho, line 9: Please use a personal global gitignore file for editor files, given that they are a personal and therefore vary varied choice.  See [`gitignore`](https://git-scm.com/docs/gitignore) for more information, especially the `core.excludesFile` part.

jonasmalacofilho (commented):
Thanks!

On top of Parker's comment, please see the two notes I left.

tests/test_commander_core.py
Meriemi-git: @jonasmalacofilho Thanks for your feedback you're right it seems to be a better solution to leave the dict empty. I tested with your suggestion to get the mode and it works.

.gitignore
Meriemi-git: It's done.

Meriemi-git:
@ParkerMc I updated the write function in test to ensure mode is not set whereas previous was not reset. I had to ensure device initially call a reset because of the command flow modification (reset is not the first command anymore) .
I put the initial reset call in the _wake_device_context function but i hesitated between here and in the __init__ function.

ParkerMc:
> I put the initial reset call in the _wake_device_context function but i hesitated between here and in the init function.

Is that necessary? From my testing, reset is only needed after a write or read.
Hense why I proposed renaming it so it makes more sense 
> It should become: open endpoint(currently called set_mode) -> read and/or write -> close endpoint(currently called reset)

> I don't see any section in protocol documentation about the command flow but i can create a new one to document this changed for all new firmware versions from v2.10.219

This is the proper use of the protocol so it applies to v1 and v2. I just had the flow wrong but it still worked.

Meriemi-git:
@ParkerMc You're right, i remove this unnecessary call and update the protocol doc.

jonasmalacofilho:
@ParkerMc does everything here look good to you?

ParkerMc:
> @ParkerMc does everything here look good to you?

Yeah, looks great. Sorry, I thought I had replied the other day.

jonasmalacofilho:
Great! Thank you both!

PR #508, opened 2022-09-23
Add support for setting fan speed for Aquacomputer Octo

aleksamagicka:
Add support for setting pump and fan speed for the Octo fan contoller in the `aquacomputer.py` driver. Also, refactor generating hwmon `pwmX` and `pwmX_enable` property names to a separate method.

Related: #438 

---

Checklist:

New CLI flag?

New device?

New driver?


liquidctl.8
```diff
@@ -236,7 +236,7 @@ Internal data used by some drivers.
 .SS Aquacomputer Farbwerk 360
 .SS Aquacomputer Octo
 .SS Aquacomputer Quadro
-Cooling channels: (D5 Next): \fIfan\fR, \fIpump\fR;, (Octo, Quadro:) not yet supported; (Farbwerk 360:) not applicable.
+Cooling channels: (D5 Next): \fIfan\fR, \fIpump\fR; (Octo): \fIeight fans\fR, (Quadro:) not yet supported; (Farbwerk 360:) not applicable.
```
jonasmalacofilho, line 239:
```suggestion
Cooling channels: (D5 Next): \fIfan\fR, \fIpump\fR; (Octo): \fIfan[1\-8]\fR; (Quadro:) not yet supported; (Farbwerk 360:) not applicable.
```

liquidctl/driver/aquacomputer.py
```diff
@@ -354,13 +362,25 @@ def set_speed_profile(self, channel, profile, **kwargs):
...
     def _set_fixed_speed_hwmon(self, channel, duty):
-        hwmon_pwm_name = self._device_info["hwmon_ctrl_mapping"][channel]
-        hwmon_pwm_enable_name = f"{hwmon_pwm_name}_enable"
+        hwmon_pwm_name, hwmon_pwm_enable_name = self._fan_name_to_hwmon_names(channel)
 
         # Set channel to direct percent mode
         self._hwmon.write_int(hwmon_pwm_enable_name, 1)
 
+        # Some devices (Octo, Quadro and Aquaero) can not accept reports in quick succession, so slow down a bit
+        time.sleep(0.5)
```
jonasmalacofilho, line 382: Is a delay always necessary?  Sleeping 500 ms would be quite noticiable: could we optimistically try without it, and only do it when necessary?

(Does it matter whether the previous write was redundant (i.e. when  `pwmX_enable` was already set)?)

aleksamagicka: I've set the delay to 0.2s now. I've been spamming it and it still behaved correctly, though 0.5s wasn't _that_ noticeable, at least to me. Setting fan speed now takes about 0.3s as a whole, which feels instantaneous (it was ~0.1s previously).

There's no way to do it optimistically, it may accept setting the speed twice and fail the third time, out of the blue. And we can't read the `pwmX_enable` value either, because that also triggers sending HID reports to the device to read the state (aside from sensor reports, the driver does not cache anything).

jonasmalacofilho: What I previously had in mind was something like:

```py
# pseudo-code, I don't know the exact failure mode when the device isn't ready
try:
    self.device.send_feature_report(ctrl_settings)
except IOError:
    # Maybe the device couldn't handle the quick succession of reports; try one more time after a bit.
    time.sleep(0.5)
    self.device.send_feature_report(ctrl_settings)
```

But a 0.2 s delay is already a lot better, and keeps things simpler, so I actually prefer it.

jonasmalacofilho:
Thanks!

liquidctl/driver/aquacomputer.py
aleksamagicka: For the record, the driver [responds](https://github.com/aleksamagicka/aquacomputer_d5next-hwmon/blob/0f525808c4267e4e1f3a28400a46171dac6f87c4/aquacomputer_d5next.c#L424) with `-ENODATA`.

jonasmalacofilho: Thanks!

PR #529, opened 2022-10-23
Hwmon support for nzxt-kraken3

aleksamagicka:
With the [nzxt-kraken3](https://github.com/liquidctl/liquidtux/blob/master/nzxt-kraken3.c) HWMON driver now having expanded support for NZXT Kraken X53/X63/X73 and Z53/Z63/Z73 devices, allow liquidctl to leverage it when available.

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ subjective "from more to less liquid control-ly" order.
 | AIO liquid cooler  | [NZXT Kraken X31, X41, X61](docs/asetek-690lc-guide.md) | USB | <sup>_LZ_</sup> | 1.9.1 |
 | AIO liquid cooler  | [NZXT Kraken X42, X52, X62, X72](docs/kraken-x2-m2-guide.md) | USB HID | <sup>_h_</sup> | 1.11.1 |
 | AIO liquid cooler  | [NZXT Kraken X53, X63, X73](docs/kraken-x3-z3-guide.md) | USB HID | <sup>_h_</sup> | 1.11.1 |
-| AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_e_</sup> | 1.11.1 |
+| AIO liquid cooler  | [NZXT Kraken Z53, Z63, Z73](docs/kraken-x3-z3-guide.md) | USB & USB HID | <sup>_he_</sup> | 1.11.1 |
 | Pump controller    | [Aquacomputer D5 Next](docs/aquacomputer-d5next-guide.md) | USB HID | <sup>_ehp_</sup> | 1.11.1 |
 | Fan/LED controller | [Aquacomputer Octo](docs/aquacomputer-octo-guide.md) | USB HID | <sup>_ehp_</sup> | 1.11.1 |
 | Fan/LED controller | [Aquacomputer Quadro](docs/aquacomputer-quadro-guide.md) | USB HID | <sup>_ehp_</sup> | 1.11.1 |
--- a/liquidctl/driver/kraken3.py
+++ b/liquidctl/driver/kraken3.py
@@ -203,6 +203,7 @@ class KrakenX3(UsbHidDriver):
             {
                 "speed_channels": _SPEED_CHANNELS_KRAKENX,
                 "color_channels": _COLOR_CHANNELS_KRAKENX,
+                "hwmon_ctrl_mapping": {"pump": "pwm1"},
             },
         ),
         (
@@ -212,14 +213,18 @@ class KrakenX3(UsbHidDriver):
             {
                 "speed_channels": _SPEED_CHANNELS_KRAKENX,
                 "color_channels": _COLOR_CHANNELS_KRAKENX,
+                "hwmon_ctrl_mapping": {"pump": "pwm1"},
             },
         ),
     ]
 
-    def __init__(self, device, description, speed_channels, color_channels, **kwargs):
+    def __init__(
+        self, device, description, speed_channels, color_channels, hwmon_ctrl_mapping, **kwargs
+    ):
         super().__init__(device, description)
         self._speed_channels = speed_channels
         self._color_channels = color_channels
+        self._hwmon_ctrl_mapping = hwmon_ctrl_mapping
 
     def initialize(self, direct_access=False, **kwargs):
         """Initialize the device and the driver.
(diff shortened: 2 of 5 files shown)


liquidctl/driver/kraken3.py
```diff
@@ -311,18 +311,13 @@ def get_status(self, direct_access=False, **kwargs):
         Returns a list of `(property, value, unit)` tuples.
         """
 
-        # no driver currently supports pwm1, so silently fallback to
-        # direct mode if it isn't available; for the same reason, also don't
-        # yet issue a warning when directly accessing the device
-
-        if self._hwmon and not direct_access and self._hwmon.has_attribute("pwm1"):
```
jonasmalacofilho, line 318: The pwm1 check should remain, in case the user is still running the old version of the driver.

And the preceeding comment should be adjusted accordingly.

liquidctl/driver/kraken3.py
```diff
@@ -695,6 +688,32 @@ def get_status(self, **kwargs):
...
+    def _get_status_from_hwmon(self):
+        return [
+            (_STATUS_TEMPERATURE, self._hwmon.read_int("temp1_input") * 1e-3, "°C"),
+            (_STATUS_PUMP_SPEED, self._hwmon.read_int("fan1_input"), "rpm"),
+            (_STATUS_PUMP_DUTY, self._hwmon.read_int("pwm1") * 100.0 / 255, "%"),
+            (_STATUS_FAN_SPEED, self._hwmon.read_int("fan2_input"), "rpm"),
+            (_STATUS_FAN_DUTY, self._hwmon.read_int("pwm2") * 100.0 / 255, "%"),
+        ]
+
+    def get_status(self, direct_access=False, **kwargs):
```
jonasmalacofilho, line 700: If I'm not mistaken, this is an unnecessary override.  Can't the Z class just use the base X class method?

aleksamagicka: Of course, just pushed that commit.

liquidctl/driver/kraken3.py
aleksamagicka: In that case, wouldn't it be better to not disregard the old version of the driver and only read the coolant temp and pump speed from it?

jonasmalacofilho: You're right, that would be better.

aleksamagicka: I've added a warning if it can't be read, similar to the [aquacomputer](https://github.com/liquidctl/liquidctl/blob/main/liquidctl/driver/aquacomputer.py#L375) driver.

aleksamagicka:
While implementing fixed duty writing through hwmon, I just noticed that we [don't set](https://github.com/liquidctl/liquidtux/blob/master/nzxt-kraken3.c#L325) the pwmX_enable value when writing to pwmX directly, we just write it to the device. I think we should cache it and then check if we should write it, like for [the curve](https://github.com/liquidctl/liquidtux/blob/master/nzxt-kraken3.c#L397)?

jonasmalacofilho:
> Can I add the additional `hwmon_ctrl_mapping` parameter [here](https://github.com/liquidctl/liquidctl/blob/main/tests/test_backward_compatibility_14.py#L57), since that's new and not being tested there?

Yes!

README.md
```diff
@@ -125,7 +125,7 @@ subjective "from more to less liquid control-ly" order.
 | AIO liquid cooler  | [NZXT Kraken X31, X41, X61](docs/asetek-690lc-guide.md) | USB | <sup>_LZ_</sup> | 1.9.1 |
 | AIO liquid cooler  | [NZXT Kraken X42, X52, X62, X72](docs/kraken-x2-m2-guide.md) | USB HID | <sup>_h_</sup> | 1.11.1 |
 | AIO liquid cooler  | [NZXT Kraken X53, X63, X73](docs/kraken-x3-z3-guide.md) | USB HID | <sup>_h_</sup> | 1.11.1 |
```
aleksamagicka, line 127: Should this be marked as experimental, since part of hwmon stuff is new?

aleksamagicka:
That should be the majority of work for this PR, now to tie the loose ends if we find any.

README.md
jonasmalacofilho: It's not necessary.

jonasmalacofilho:
At a glance, this LGTM. But I have to admit that I haven't been able to thoroughly review.  I'll give you write access to this repo, and you can merge this if or as soon as you're comfortable with it.

---

Just one question: the 200 ms timeout, why is it required, and is another way to solve that problem?

aleksamagicka:
Thanks for the invite, much appreciated! I'll work on getting these PRs across the finish line as soon as college stuff clears up.

As for the 200ms timeout, [in the driver](https://github.com/liquidctl/liquidtux/blob/master/nzxt-kraken3.c#L401) we apply the whole curve whenever one point of it is changed, if it was already in curve mode. So from liquidctl we set those curve points in succession and trigger the same amount of HID reports sent to the device, from the driver. Setting `pwmX_enable` immediately after causes it to do nothing IIRC, it just gets confused.

Looking at it again, we pause a bit before setting `pwmX_enable` to 2, if it was already 2, which doesn't make much sense, since that curve is now wholly applied anyway. Only setting it to 2 if it's in some other mode should work, without the timeout.

aleksamagicka:
So I investigated this on the Z53 (for now), and what I said above doesn't work.

Scenario 1: `pwmX_enable` is set to 2 after a small timeout, regardless of what it is. Different curves get applied properly, even when spammed in quick succession.

Scenario 2: `pwmX_enable` is set to 2 only if was something else, without any timeout. Order of actions (channel was in fixed mode):

- Setting first curve: it's set correctly, `pwmX_enable` is set to 2
- Setting second curve: nothing happens
- Setting first curve again: nothing seems to happen, it's still on the first curve from step 1 (same curve)

And so on, until it happens that setting the first curve activates the second one, and vice versa. Or that setting a curve has no effect.

Looks like to me that the firmware is finicky when spammed with lots of reports, which happens when 40 temp points are set through the driver in quick succession. Maybe it bails after a few, caches others and reprocesses them at some point when new ones arrive, or discards them after some period... I'm not sure what it does exactly, but it always honors the latest single report that's sent from setting `pwmX_enable` to 2, which sends the final curve (initiated after a small wait so it has time to decide it won't act on the previous reports).

jonasmalacofilho:
@aleksamagicka, if I correctly understood your results, scenario 1 is significantly more robust/reliable and, also, simple. So let's go with it (i.e. keep everything as is).

---

On the kernel side, we might want to document his finicky behavior for other users of the driver.

aleksamagicka:
That's right. I've been meaning to add an expanded note to liquidctl earlier today, but it's getting late now and I'll do it tomorrow (for the kernel driver as well).

I'll test this on both devices tomorrow once again and then we can merge this and plan on trying to upstream the kernel part.

jonasmalacofilho:
Thanks!

PR #558, opened 2023-01-14
Auto matching driver for Corsair AIO using lsusb -v ( hydro-platinum )

st1vms:

If the implementation is not trivial, please also include a short overview.

Fixes: 
Closes: 
Related: 

---

Checklist:

New CLI flag?

New device?

New driver?

*NOTE: black formatted applied to hydro_platinum.py, most of the changes are just code format.


jonasmalacofilho:
Hey, I wish you had given me time to respond to https://github.com/liquidctl/liquidctl/pull/557#issuecomment-1382570721 before starting to work on this.

> From my honest point of view, some of these compatibility issues may be easily fixed with an `lsusb -v` wrapper that parses `VID:PID` pairs looking for products names, using a pre-defined table that also stores the required properties for `_MATCHES`. It may help for for auto-generating udev rules and VID:PID parameter assignment.

First, we already internally do part of what you propose (through PyUSB). Well, technically this specific driver works at the HID (kernel) level, but even then we do a similar thing (through cython-hidapi). The point is that there's no need to parse lsusb output.

Another issue is that we strive to be cross-platform. And depending on (Linux's) lsusb is obviously incompatible with it.

With that said, it seems that you're also suggesting that we match primarily based on the manufacturer and product string descriptors, instead of the VID and PID. While I can see how that might seem reasonable, that's not practical:

- several devices report generic product descriptors, even for different families (e.g. "NZXT USB Device" for both Kraken X62 and Smart Device... no, they don't use the same protocol);
- several devices report misleading product descriptors (e.g.the H115i Pro, not Pro XT, reports  "H115i Platinum"... and, no, those don't all use the same protocol);
- in general, reading the string descriptors results in USB traffic to the device, which we don't want to do while just probing for what's there (some platforms may cache these descriptors, in some instances, but not all do);
- at a systems programming level, USB devices are typically identified by VID, PID and, optionally, bcdDevice (but none of the devices we deal with needs to be handled differently according to the bcdDevice, so we've recently removed it from `_MATCHES`).

---

Sorry you wasted some time on this, but it's a no-go, for the reasons above.

Issue #459, opened 2022-05-19
MSI MPG Coreliquid K240


MaceAus:
The First-party software for this device is awful, and doesn't allow 0RPM mode on some fans. 
If implementation of this into Liquidctl was not too difficult I would greatly appreciate it.

Please let me know if I can provide further information. 

Device type: AIO Liquid Cooler w/integrated controller
Product page: https://www.msi.com/Liquid-Cooling/MPG-CORELIQUID-K240
First-party software: [MSI Center](https://www.msi.com/Landing/MSI-Center) 
Other commonly used tools: None 

The device appears to support the following features:

 - monitoring of temperatures or other environment sensors: Yes
 - monitoring of fan or pump speeds or duty cycles: Yes
 - monitoring of voltages, currents or power: Yes
 - configurable fan or pump speeds:  Yes
 - configurable voltages or current/power limits: No
 - configurable lighting of embedded LEDs: Yes
 - configurable lighting of accessories like RGB fans or LED strips: Yes

The device is physically connected to the host system using:

- USB: Yes (USB 2.0)
- PCI-E: 
- onboard the motherboard: 
- other: Additional PWM header (appears to be for monitoring and perhaps power) 
[further testing suggests this is a 4pin PWM connector with the PWM control pin disabled]

The device uses which protocols to communicate with the host system:

- HID: 
- USB: Yes
- I²C: 
- SMBus: 
- other: I can investigate on request. 

Additional information about the device:
* Asetek gen 7 pump
* PWM header will return pump pwm% as set in Bios (regardless of setting in MSI Center), however pump RPM reads correctly as set in MSI Center. These readings are from Libre Hardware Monitor. 
* Changing the duty cycle applied to the PWM header does not change pump rpm
* integrated fan in pump block - cannot be monitored/controlled without MSI Center 
* Also includes 2.4" LCD display - I do not personally care if this is controllable. 

Finally, I can help with:

- testing changes on Linux: 
- testing changes on Windows: Yes
- testing changes on Mac OS: 
- attempting to capture USB traffic: Yes
- attempting to capture I²C/SMBus traffic: 
- analyzing traffic data: 
- documenting the protocol: 
- implementing the driver: 

assuming documentation and/or assistance will be provided.

jonasmalacofilho:
Hey,

There's some initial work on a driver for these devices in PR #371.  IIRC, the author ended up returning their cooler, and that's why the development stopped.  If anyone can pick up where they left off, that would be awesome!

[...]

PR #564, opened 2023-02-02
Add support for MSI Coreliquid K360 device

aapo-kossi:
Takes over the early work of Tatsh to implement a driver for MSI MPG coreliquid K360 cooler.

Fixes: 
Closes: #371 
Related: #459 

---

Checklist:

    - [x] `liquidctl.8` Linux/Unix/Mac OS man page

New CLI flag?

New device?

- [x] Add entry to the README's supported device list with applicable notes (at least `e`) and `git` MRLV

New driver?


aapo-kossi:
I think the pump liquid temperature sensor data is currently read incorrectly, otherwise status support is now working. I will document the protocol when I have some free time and it should be quite simple to have fan control working soon too. I already implemented RGB control for this device in OpenRGB so that should be relatively simple here too. I am also planning on writing a fan controller script specific for this device as the protocol has support for some preset fan/pump modes that would be nice to have accessible, but all in due time.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,971 @@
...
+                continue
+            if match and match.lower() not in desc.lower():
+                continuei
+            if handle.hidinfo['usage_page'] != 0xff00:
+                continue
+            consargs = devargs.copy()
+            consargs.update(kwargs)
+            dev = cls(handle, desc, **consargs)
+            _LOGGER.debug('%s identified: %s', cls.__name__, desc)
+            yield dev
```
jonasmalacofilho, line 244: Take a look at

https://github.com/liquidctl/liquidctl/blob/3c227276666dbfee2b60491dbe3d87b35598f776/liquidctl/driver/rgb_fusion2.py#L107-L126

for an example of how this might be simplified and also made a bit more robust.

drpars:
K360 cihazım var.  Bende de aynı şekilde gözüküyor. Ayrıca liquidctl'in MPG soğutucusunun iki kopyasını algılamasını düzeltilmiş gözüküyor.
MSI MPG Coreliquid K360
├── Fan 1 speed              1570  rpm
├── Fan 1 duty                 60  %
├── Fan 2 speed              1562  rpm
├── Fan 2 duty                 60  %
├── Fan 3 speed              1595  rpm
├── Fan 3 duty                 60  %
├── Water block speed        2521  rpm
├── Water block duty           50  %
├── Pump speed               2777  rpm
├── Pump duty                 100  %
├── Temperature inlet         0.0  °C
├── Temperature outlet        0.0  °C
├── Temperature sensor 1    125.0  °C
└── Temperature sensor 2    125.0  °C

Tatsh:
All of the code is based on the decompilation of the .NET code. Most likely it was labeled this way in the code. It may not necessarily be functional for every device.

If getting a value for your device is not working you should probably remove it.

As for detecting 2 devices of the same kind, this is an issue with how liquidctl enumerates devices with hidapi. A device may register itself twice with different usages. I did have an incorrect fix here https://github.com/liquidctl/liquidctl/pull/372 but this is not the way to fix it.

aapo-kossi:
> All of the code is based on the decompilation of the .NET code. Most likely it was labeled this way in the code. It may not necessarily be functional for every device.
> 
> If getting a value for your device is not working you should probably remove it.
> 
> As for detecting 2 devices of the same kind, this is an issue with how liquidctl enumerates devices with hidapi. A device may register itself twice with different usages. I did have an incorrect fix here #372 but this is not the way to fix it.

Yes, I looked at the code you decompiled and it is the same. The reason for my confusion is the example result of the status call you added to the documentation for the driver, which implies the temp sensors would have been working unless the numbers were just invented. 

I fixed the duplicate device issue with the extended MPGCooler.probe method.

extra/coreliquid_ctl
```diff
@@ -0,0 +1,276 @@
+#!/usr/bin/env python3
```
jonasmalacofilho, line 1: What specific features does this script offer over `extra/yoda`?

Ideally, we should just improve `yoda` to support these additional features, unless there's something here that's totally unreasonable for any other device.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1076 @@
...
+        self._fan_count = kwargs.pop('fan_count', 5)
+        # the following fields are only initialized in connect()
+        self._data = None
+        self._feature_data = None
+
+    @classmethod
+    def probe(cls, handle, **kwargs):
+        if (handle.hidinfo['usage_page'] != USAGE_PAGE):
+            return
+        yield from super().probe(handle, **kwargs)
```
jonasmalacofilho, line 276: Please keep the same comments as in `RgbFusion2.probe`, since this isn't a trivial piece of code.

jonasmalacofilho (commented):
Please remove the images, it's desirable to keep this repository size as small as possible.

Instead, you can add them to <https://github.com/liquidctl/collected-device-data>.

extra/coreliquid_ctl
aapo-kossi: The control methodology for this device is quite fundamentally different, so I wasn't sure how I could include it in yoda. The device has it's own PID controller, so the control script doesn't need to calculate or send fan speeds at all, it is sufficient to simply send the current temperature to the device and it will automatically control itself by the set profile. I just now realized that there is still an error in the way the current temp is sent and that a limitation from the manufacturer is that this method can only use one single temperature reading (usually cpu) to control all channels.

Another difference is that the device has named built in profiles, some of which have different limitations that I am not sure can be bypassed. For instance, custom fan speeds refuse to lower fan duties below 20%, while the SMART mode enables fan-stop for two of the three fans. I think the named profiles could quite easily be added to yoda, but there is no reason to if the control functions cannot be combined.

liquidctl/driver/msi.py
aapo-kossi: Fixed by https://github.com/liquidctl/liquidctl/pull/564/commits/c4ac9a2677690e84851b1ab642e8d0efa6feeeb9 and https://github.com/liquidctl/liquidctl/pull/564/commits/5282e7776d9cb3299e916c29e26312c9dd5fc2da together

aapo-kossi:
The oled image setting is *kind of* working now, but seems very inconsistent. I will experiment some more with it. Also still need to implement some oled modes.

PetrKlassen:
I tried to run your implementation with S360. Fans changes speed depending on the temperature and mode, but the temperature is not displayed correctly on the monitor. How is this possible? Is there some kind of duplication of temperature information for the sensors and for the monitor?

aapo-kossi:
> I tried to run your implementation with S360. Fans changes speed depending on the temperature and mode, but the temperature is not displayed correctly on the monitor. How is this possible? Is there some kind of duplication of temperature information for the sensors and for the monitor?

Hi, are you using yoda or coreliquid_ctl as the control script? With coreliquid_ctl, there should be no difference between the temperature that controls the fan speeds and the temperature displayed on the screen. With yoda, the sensor temperature is never sent to the device, the control is only performed through sending duty cycle values. I am now coming up with a bit more time to finish this pr so I could combine the functionalities into yoda, but I will need to think about how to build the support without introducing too much complexity. Currently yoda isn't used for this kind of communication.

PetrKlassen:
last time I checked command `python extra/coreliquid_ctl --debug using coretemp.package_id_0 control fans with "silent"` monitor just didn't update information but speed of fans was changing depending on system load.

aapo-kossi:
This is now approaching a state that I can be content with, though I still need to fix some details (image uploading checksums currently do not match even when the file gets successfully transferred, some screen configuration functions need to be cleaned up).

extra/coreliquid_ctl
aapo-kossi: I have now incorporated the necessary changes into yoda, which I think makes it slightly less clean but could be relatively simple to adapt for new devices that work similarly in the future.

aapo-kossi:
So, I think I have documented everything relevant and the driver has been working for me. LCD image uploading is still inconsistent, but its not clear why and usually repeating a command resolves problems. Save slots for custom uploaded images also don't really seem to work, all of them seem to be overwritten no matter which slot is specified for the newest image. A little inconvenient, but I imagine not a big issue as it just slows down switching between custom images because of the need to upload again each time.

jonasmalacofilho:
@aleksamagicka, can you take a look at this when you have some time?

docs/developer/protocol/coreliquid.md
```diff
@@ -0,0 +1,314 @@
...
+| 0x0A-0x11 | Radiator fan 2 config |
+| 0x12-0x19 | Radiator fan 3 config |
+| 0x01A-0x21 | Pump speed config |
+| 0x22-0x29 | Waterblock fan config |
+
+A fan temperature config consists of 8 bit integer values:
+  - Mode index
+  - 7 temperature points in Celsius.
+
+### Get fan speed config - `0x32` (Get) or `0x40` (Set)
```
aleksamagicka, line 204: I think that naming this section `Fan speed config` in line with the previous one would make more sense, as you can both get and set it.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1468 @@
...
+        return self._write(
+            (0xC0, clamp(main, 0, 100), clamp(left, 0, 100), clamp(right, 0, 100)), 0xCC, prefix=1
+        )
+
+    def set_oled_cpu_message(self, message):
+        return self._write([0x90] + list(message[:60].encode("ascii", "ignore")))
+
+    def set_oled_show_hardware_monitor(self, opts, radiator_fan_smart_mode_on_off=True):
+        """
+
```
aleksamagicka, line 980: NIt: unneded blank

extra/yoda
```diff
@@ -155,33 +168,86 @@ def parse_profile(arg, mintemp=0, maxtemp=100, minduty=0, maxduty=100):
...
-            raise ValueError('temperature must be integer between {} and {}'.format(mintemp, maxtemp))
+            raise ValueError(
+                'temperature must be integer between {} and {}'.format(mintemp, maxtemp)
+            )
         if not isinstance(duty, int) or duty < minduty or duty > maxduty:
             raise ValueError('duty must be integer between {} and {}'.format(minduty, maxduty))
     return normalize_profile(val, critx=maxtemp)
 
 
+def auto_control(device, channels, profiles, sensors, update_interval):
```
aleksamagicka, line 197: This is mostly geared towards the specific MSI AIO, leave a comment to note that.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1468 @@
...
+            ("Fan 1 duty", array[0x16] + (array[0x17] << 8), "%"),
+            ("Fan 2 speed", array[4] + (array[5] << 8), "rpm"),
+            ("Fan 2 duty", array[0x18] + (array[0x19] << 8), "%"),
+            ("Fan 3 speed", array[6] + (array[7] << 8), "rpm"),
+            ("Fan 3 duty", array[0x1A] + (array[0x1B] << 8), "%"),
+            ("Water block speed", array[8] + (array[9] << 8), "rpm"),
+            ("Water block duty", array[0x1C] + (array[0x1D] << 8), "%"),
+            ("Pump speed", array[0xA] + (array[0xB] << 8), "rpm"),
+            ("Pump duty", array[0x1E] + (array[0x1F] << 8), "%"),
+            ##    ('Temperature inlet', array[12] + (array[13] << 8), '°C'),
```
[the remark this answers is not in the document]
aapo-kossi: These were a part of the original contributors implementation:
> > All of the code is based on the decompilation of the .NET code. Most likely it was labeled this way in the code. It may not necessarily be functional for every device.
> > If getting a value for your device is not working you should probably remove it.

My device returns constant values instead of anything meaningful: inlet and outlet are always zeroed out, and the sensor temperatures are constant 125C. My conclusion is that either the temperature sensors need to somehow be activated separately or this is a feature only included in some other models (most likely the latter, since the device does not adjust duties without external temperature communication). The thought is that they could be useful in the future if other models are added / the implementation is refined. I can add a brief comment in relation to this.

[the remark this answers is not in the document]
aleksamagicka: Thanks, adding a comment for the future sounds good to me since the code is still there.

docs/msi-mpg-coreliquid-guide.md
```diff
@@ -0,0 +1,165 @@
...
+
+The screen resolution is 320 x 240 px, and custom images uploaded with this driver are resized to fit this requirement. The screen orientation and brightness (0-100) can also be controlled. The only channel available for the K360 model is "lcd".
+
+Maximum length of the displayed banner mesages is 62 ASCII characters. hardware status display functionality is limited, as the displayed data must be communicated to the device. This functionality is implemented in the driver, but currently its usage is limited to yoda, which gpu-unaware so the gpu_freq and gpu_usage parameters will not display correct information without custom update services.
+
+
+| mode name | action | options |
+| --- | --- | --- |
+| hardware | set the screen to display hardware info | up to 3 semicolon delimited keys from the available sensors |
+| image | set the screen to display a custom or preset image | \<type (0=preset,1=custom)\>;\<index\>[;\<filename\>] |
```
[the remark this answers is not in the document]
aapo-kossi: The `<index>` refers to a save slot on the device. It has several preset animations and is supposed to be able to store multiple custom images that can be selected with their uploaded index. This has been the most challenging part to get working, custom images now still seem to overwrite any previous custom images no matter which index is used, but the index at which the custom image can be loaded is also specific, and inconsistent. Often, but not always, the first time after uploading an image, loading its index will instead load a preset, and sometimes, but not always, the custom image will also get viewed from other indices preserved for the custom files in addition to the one it was uploaded to.
Trying to upload an image with the type flag 0 currently raises an error. I thought this would be the clearest option syntax, the layout lends itself more naturally to MSI's GUI where you only have separate rows for preset and custom images. The redundant `1` could be prevented with another flag specifically for `upload-image`, but there are already many enough to be slightly confusing. I am open to alternative cleaner solutions though.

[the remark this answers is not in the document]
aapo-kossi: > Or is it just for `<direction (0-3)>` in case of `settings`?
Yes.

liquidctl.8
```diff
@@ -372,6 +372,68 @@ Channel	Mode	#colors (Platinum)	#colors (Pro XT/Elite RGB)	#colors (Platinum SE)
...
+\fIrandom\fR	0
+\fIrap\fR	0
+\fIstack\fR	1
+\fIvisor\fR	2
+\fIwater drop\fR	1
+.TE
+.PP
+When applicable the animation speed can be set with
+.BI \-\-speed= value ,
+where the allowed values are: \fI0\fR, \fI1\fR and \fI2\fR.
```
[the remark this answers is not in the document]
aapo-kossi: Has now been added.

extra/yoda
aapo-kossi: Thank you for the suggestion, added a note.

liquidctl/driver/msi.py
aapo-kossi: Fixed with overall improvements.

docs/developer/protocol/coreliquid.md
aapo-kossi: Thank you for spotting this.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1468 @@
...
+        for i in self.parse_channel(channel):
+            self._fan_cfg[i] = _FanConfig(_FanMode.CUSTOMIZE.value, *duties)
+            self._fan_temp_cfg[i] = _FanTempConfig(_FanMode.CUSTOMIZE.value, *temps)
+
+        self.set_fan_config(self._fan_cfg)
+        self.set_fan_temp_config(self._fan_temp_cfg)
+
+        _LOGGER.info(
+            "duty profiles on this device require continuous communication, "
+            "setting default speeds to the highest point on the curve."
```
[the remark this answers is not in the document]
aapo-kossi: I see now how the original message could seem misleading, the message was intended to convey exactly what you say I do. The new version should be clearer.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1468 @@
...
+
+        self.set_fan_config(fan_cfg)
+        self.set_fan_temp_config(fan_temp_cfg)
+
+    def parse_channel(self, channel):
+        if channel == "pump":
+            return [4]
+        elif channel == "fans":
+            return range(_RAD_FAN_COUNT)
+        elif channel == "waterblock fan":
```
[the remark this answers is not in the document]
aapo-kossi: Updated.

docs/msi-mpg-coreliquid-guide.md
```diff
@@ -0,0 +1,165 @@
...
+With those out of the way, the pump speed can be configured to a fixed duty value or with a profile dependent on a (temperature) signal that MUST be periodically sent to the device.
+
+Fixed speeds can be set by specifying the desired channel and duty value.
+
+```
+# liquidctl set pump speed 90
+```
+
+| Channel | Minimum duty | Maximum duty |
+| --- | -- | --- |
```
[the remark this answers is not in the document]
aapo-kossi: Updated: radiator fan channels have identical limits so they are still in the same row, but now all are explicitly listed.

docs/msi-mpg-coreliquid-guide.md
```diff
@@ -0,0 +1,165 @@
...
+## The LCD screen
+
+The screen resolution is 320 x 240 px, and custom images uploaded with this driver are resized to fit this requirement. The screen orientation and brightness (0-100) can also be controlled. The only channel available for the K360 model is "lcd".
+
+Maximum length of the displayed banner mesages is 62 ASCII characters. hardware status display functionality is limited, as the displayed data must be communicated to the device. This functionality is implemented in the driver, but currently its usage is limited to yoda, which gpu-unaware so the gpu_freq and gpu_usage parameters will not display correct information without custom update services.
+
+
+| mode name | action | options |
+| --- | --- | --- |
+| hardware | set the screen to display hardware info | up to 3 semicolon delimited keys from the available sensors |
```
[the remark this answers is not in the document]
aapo-kossi: Correct.

liquidctl/driver/msi.py
```diff
@@ -0,0 +1,1468 @@
...
+        "cpu_temp": _OLEDHardwareMonitorOffset.CPU_TEMP,
+        "gpu_freq": _OLEDHardwareMonitorOffset.GPU_MEMORY_FREQ,
+        "gpu_usage": _OLEDHardwareMonitorOffset.GPU_USAGE,
+        "fan_pump": _OLEDHardwareMonitorOffset.FAN_PUMP,
+        "fan_radiator": _OLEDHardwareMonitorOffset.FAN_RADIATOR,
+        "fan_cpumos": _OLEDHardwareMonitorOffset.FAN_CPUMOS,
+    }
+    _MATCHES = [
+        (0x0DB0, 0xB130, "MSI MPG Coreliquid K360", {"fan_count": 5}),
+        (0x0DB0, 0xCA00, "Unknown", {}),
```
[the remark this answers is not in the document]
aapo-kossi: Should they be removed (commented out for future contributions) if it is undesirable to have technically unsupported devices in the udev rules? These were added in the original PR, to start I now at least wrote the more descriptive names.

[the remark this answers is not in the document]
jonasmalacofilho: How confident are we that those suspected IDs will really work properly with the driver?

Depending on that it might be wise to either comment them out (if not much) or leave them in but require an [unsafe flag](https://github.com/liquidctl/liquidctl/blob/e3f91793b35550f1b454e26e37f110c70911b9eb/liquidctl/driver/ddr4.py#L212-L215) for actual use (if quite confident).

---

Please suffix all new device descriptions with ` (experimental)`.

liquidctl.8
```diff
@@ -372,6 +372,68 @@ Channel	Mode	#colors (Platinum)	#colors (Pro XT/Elite RGB)	#colors (Platinum SE)
 \fIled\fR	\fIsuper\-fixed\fR	24	16	48
 .TE
 .
+.SS MSI MPG Coreliquid k360
+Cooling channels: \fIfan1\fR, \fIfan2\fR, \fIfan3\fR, \fIfans\fR, \fIwaterblock fan\fR, \fIpump\fR.
```
aleksamagicka, line 376: `waterblock-fan` now.

docs/developer/protocol/coreliquid.md
```diff
@@ -0,0 +1,312 @@
...
+| ---------- | ----------- |
+|  0x01 | 0x7E |
+|  0x02 | brightness (0-100) |
+|  0x03 | direction (0-3) |
+
+## Image Upload Commands
+
+### Upload Image - `0xC0` (GIF) or `0xD0` (Banner)
+
+File uploads are initiated by a single report, after which the data is transferred in chunks of 60 bytes. Uploaded images must be 240x320 px size, and in the standard 24-bit color BMP format. A short sleep (2 seconds has proven safe) should be pla2 seconds has proven safeced between the transfer initiation and start of the data transfer to make sure the device is ready.
```
aleksamagicka, line 288: `A short sleep (2 seconds has proven safe) should be pla2 seconds has proven safeced` -> `A short sleep (2 seconds has proven safe)`

liquidctl/driver/msi.py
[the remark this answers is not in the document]
aapo-kossi: I am reasonably confident that the driver should be mostly functional for the listed devices, as there are no meaningful differences in the communication protocol compared to the decompiled manufacturer implementation. [Some issues](https://github.com/liquidctl/liquidctl/pull/564#issuecomment-1675871831) clearly still occur, but from that it seems that the fan control does work on the S360. I added unsafe checks for this.

aapo-kossi:
I also implemented the option to use `--unsafe` with [extra/yoda](https://github.com/liquidctl/liquidctl/blob/905d954085c51b2cda2a1c73a9cab693b74fa962/extra/yoda), but should that become a separate pull request instead of polluting this?

docs/msi-mpg-coreliquid-guide.md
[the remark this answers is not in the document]
aapo-kossi: I added some clarification to the docs about this now.

jonasmalacofilho:
> I also implemented the option to use `--unsafe` with [extra/yoda](https://github.com/liquidctl/liquidctl/blob/905d954085c51b2cda2a1c73a9cab693b74fa962/extra/yoda), but should that become a separate pull request instead of polluting this?

Yes, that would be better. Thanks!

jonasmalacofilho (approved):

jonasmalacofilho:
I just noticed it, sorry, but why is it tagged as being Linux-only in the supported devices table?

jonasmalacofilho:
If there's indeed such a limitation, the driver will need to enforce it (at least by default). For now I'll just remove the flag, and we'll see.

By the way, that flag is indeed for driver that _really_ only work on Linux, such those dealing with I2C or SMBus (which we simply haven't implemented for other platforms).

Thanks again!

PR #585, opened 2023-03-31
[CLI] Print the usb port in json status

SuspiciousActivity:
When printing the status of devices in json inside the cli, we get `bus` and `address`, but when `bus` is a `usb`, `address` is quite useless and it would be good to see the `port` to address the correct device using `--usb-port`. As you can see in [this](https://github.com/jmarucha/FanControl.Liquidctl/issues/16#issuecomment-1484169867) comment, the `port` is not shown:

```json
[
    {
        "bus": "usb1",
        "address": 4,
        "description": "Asetek 690LC (assuming EVGA CLC)",
        "status": [...]
    }
]
```

After the patch:

```json
[
    {
        "bus": "usb1",
        "address": 4,
        "port": "11",
        "description": "Asetek 690LC (assuming EVGA CLC)",
        "status": [...]
    }
]
```

The implementation is taken from the [_list_devices_human](https://github.com/liquidctl/liquidctl/blob/c20fbb0f6ae4a0ee6cc6bd091ecd7ad99502cb0e/liquidctl/cli.py#L218) function which already shows the port.

There is one [test](https://github.com/SuspiciousActivity/liquidctl/actions/runs/4572209608/jobs/8071227024) failing that I have no idea why, and to be honest I don't know what it even does and how it has anything to do with my patch. So if someone could give some help on that, that'd be great.

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -248,6 +248,7 @@ def _dev_status_obj(dev, status):
     return {
         'bus': dev.bus,
         'address': dev.address,
+        'port': '.'.join(map(str, dev.port)) if dev.port else None,
         'description': dev.description.replace(' (experimental)', ''),
         'status': [convert(x) for x in status]
     }
--- a/tests/_testutils.py
+++ b/tests/_testutils.py
@@ -268,7 +268,7 @@ class VirtualBusDevice(BaseDriver):
 
     @property
     def port(self):
-        return None
+        return [12, 34]
 
 
 class VirtualBus(BaseBus):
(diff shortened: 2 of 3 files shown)


jonasmalacofilho:
Isn't the port already returned from `list --json`?

Also, we probably don't want to cram every possible device property in the output from `status --json`. Instead, consumers are expected to use the bus and address to consult any additional properties they need from `list --json`.

By the way, the USB port is generally only available for a very small subset of the devices supported by liquidctl. AFAIK, on most platforms the USB port is not available from HIDAPI (and HIDs are the vast majority of the devices we support).

---

What's the real-world use case and issue? Is fetching the USB port from `list --json` too expensive, or otherwise impractical?

PR #605, opened 2023-05-29
Add support for nzxt krakens 2023

brokenmass:
Adds support for new NZXT Kraken AIOs (2023 standard and elite). Protocol is pretty much identical to krakenZ with following differences:
- max bulk size write buffer is 2MB
- new devices do not have color channels
- new devices have different resolutions than kraken Z ( 640x60 for kraken elite, 240x240 for kraken standard)
- when setting fan speed byte 3 and 4 must be `0x1 0x1` instead of `0x0 0x0`
- when setting pump speed byte 3 and 4 must be `0x1 0x0` instead of `0x0 0x0` (but even `0x0 0x0`) works

Closes: #596 

---

Checklist:

- [x] Adhere to the [development process]
- [x] Conform to the [style guide]
- [x] Verify that the changes work as expected on real hardware
- [x] Add automated tests cases
- [x] Verify that all (other) automated tests (still) pass
- [x] Update/add documentation
    - [x] README, with ["new/changed in" notes]
    - [x] applicable `docs/*guide.md` device guides, with ["new/changed in" notes]
    - [x] `liquidctl.8` Linux/Unix/Mac OS man page
    - [x] protocol documentation in `docs/developer/protocol`
- [x] Submit relevant data, scripts or dissectors to https://github.com/liquidctl/collected-device-data

New device?

- [x] Regenerate `extra/linux/71-liquidctl.rules` (instructions in the file header)
- [x] Add entry to the README's supported device list with applicable notes (at least `e`) and `git` MRLV

[development process]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/process.md
[style guide]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/style-guide.md
["new/changed in" notes]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/process.md#newchanged-in-notes

PR #634, opened 2023-11-01
docs: set correct links for kraken models in readme

WebsiteDeveloper:


jonasmalacofilho:
Oh, we missed that in PR #605.

But I think it might be better to rename the guide back to `kraken-x3-z3-guide.md`, so any external links to it remain valid. I'll do that...

Thanks!

PR #637, opened 2023-11-14
Add experimental support for H110i GT

Serphentas:

If the implementation is not trivial, please also include a short overview.

Fixes: n/a
Closes: n/a
Related: #142 #147 

---

Checklist:

New CLI flag?

New device?

- [x] Add entry to the README's supported device list with applicable notes (at least `e`) and `git` MRLV

New driver?


Serphentas:
This is almost a 1:1 version of the original PR with simplified code in some parts.

CRCinAU:
fwiw:

```
$ sudo python -m liquidctl status
Corsair H110i GT (experimental)
├── Liquid temperature    31.7  °C
├── Fan 1 speed            782  rpm
├── Fan 2 speed            784  rpm
└── Pump speed            2353  rpm
```

Serphentas:
Per tests with @CRCinAU there is something broken in the driver when writing the fan/pump settings. The original version works fine.

Serphentas:
Pump mode is now entirely supported with `--pump-mode {quiet, extreme}` but custom fan curves aren't applied properly. Will look into captures with @CRCinAU shortly.

Serphentas:
Custom fan curves now work properly with max RPM per channel being read on demand.

jonasmalacofilho:
That's great, thanks!

What more is left to do here? Is it time for a first review?

Serphentas:
Code-wise we're good, we just need to add protocol documentation/tests/CLI completion. I can look into it today, but we can also share this task (DM me).

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,420 @@
...
+ - Corsair H110i GT
+
+Supported features
+------------------
+
+ - general monitoring
+ - pump speed control
+ - fan speed control
+
+Copyright (C) 2020–2020  Jonas Malaco
```
aleksamagicka, line 15: Perhaps a stupid question: is it necessary to put 2020 twice?

Serphentas: You're right, this is useless if contribution was over a single year.

Serphentas:
We are good for review now, data dumps available to include in the database.

liquidctl/driver/coolit.py
jonasmalacofilho: We no longer include dates in the copyright notices, with only a few exceptions.

```
Copyright Roberto Marques, Serphentas  and contributors
```

Related: 1cc8cf03d78bb269f8614e42aadcae6a1acb350f
Also: [process.md#license-and-copyright](https://github.com/liquidctl/liquidctl/blob/main/docs/developer/process.md#license-and-copyright)

---

I also don't know why I'm listed as the first co-author here... I don't remember having worked on it directly (although this driver has been in the works for a while, so I might just have forgotten about it).

Was there something from me in the old PR?

Serphentas: I think this is an old-style attribution per old PR (see [files](https://github.com/liquidctl/liquidctl/pull/147/files)). I didn't find any commit by you (@jonasmalacofilho) so I'll just roll with myself and @RobertoBiundo.

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,417 @@
...
+        buf[1:] = dataPackage
+
+        return self._send_buffer(buf)
+
+    def _send_buffer(self, buf):
+        LOGGER.debug("write %s", buf.hex())
+        self.device.clear_enqueued_reports()
+        self.device.write(buf)
+        buf = bytes(self.device.read(_REPORT_LENGTH))
+        LOGGER.debug("received %s", buf.hex())
```
aleksamagicka, line 302: Same for this.

docs/coolit-guide.md
```diff
@@ -0,0 +1,51 @@
...
+Corsair H110i GT (experimental)
+└── Firmware version    2.0.0
+```
+
+When (re)initializing the device, it is possible to select the pump mode:
+
+```
+# liquidctl initialize --pump-mode extreme
+Corsair Hydro H100i Pro
+└── Firmware version      2.10.0.0
```
aleksamagicka, line 21: Shouldn't it be just `2.10.0`? The code doesn't output four numbers.

docs/coolit-guide.md
```diff
@@ -0,0 +1,51 @@
+# Corsair Hydro H110i GT AIO liquid cooler
+_Driver API and source code available in [`liquidctl.driver.coolit`](../liquidctl/driver/coolit.py)._
+
+## Initialization
+[Initialization]: #initialization
+
+The AIO does not need to be initialized prior to use, but it will set the pump
+mode to quiet.
```
aleksamagicka, line 8: If _quiet_ is in backticks later on, maybe surround it here as well.

docs/coolit-guide.md
```diff
@@ -0,0 +1,51 @@
...
+Corsair H110i GT (experimental)
+├── Liquid temperature    32.6  °C
+├── Fan 1 speed           1130  rpm
+├── Fan 2 speed           1130  rpm
+└── Pump speed            2831  rpm
+```
+
+## Fan speed control
+
+Fan speeds can be configured either to fixed duty values or profiles.  The
```
aleksamagicka, line 44: Nit: double space before `The`.

docs/coolit-guide.md
```diff
@@ -0,0 +1,51 @@
...
+# liquidctl initialize
+Corsair H110i GT (experimental)
+└── Firmware version    2.0.0
+```
+
+When (re)initializing the device, it is possible to select the pump mode:
+
+```
+# liquidctl initialize --pump-mode extreme
+Corsair Hydro H100i Pro
```
aleksamagicka, line 20: Shouldn't it be `Corsair H110i GT`?

README.md
```diff
@@ -120,6 +120,7 @@ subjective "from more to less liquid control-ly" order.
 | :--                | :-- | --: | :-: |
 | AIO liquid cooler  | [Corsair Hydro H80i GT, H100i GTX, H110i GTX](docs/asetek-690lc-guide.md) | <sup>_Ze_</sup> | 1.9.1 |
 | AIO liquid cooler  | [Corsair Hydro H80i v2, H100i v2, H115i](docs/asetek-690lc-guide.md) | <sup>_Z_</sup> | 1.9.1 |
+| AIO liquid cooler  | [Corsair Hydro H110i GT](docs/coolit-guide.md) | <sup>_e_</sup> | git |
```
aleksamagicka, line 123: We've mostly done away with the `(experimental)` tagging for devices in #663, so `e` is not needed. Same goes for the docs and name of the device in the driver.

aleksamagicka (changes requested):

docs/coolit-guide.md
Serphentas: You're right, corrected.

docs/coolit-guide.md
Serphentas: Corrected

docs/coolit-guide.md
Serphentas: Corrected

docs/coolit-guide.md
Serphentas: Actually `2.0.0` but yeah I changed it

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,417 @@
...
+    def _send_command(self, dataPackage):
+        buf = bytearray(_REPORT_LENGTH)
+
+        buf[0] = len(dataPackage)
+        buf[1:] = dataPackage
+
+        return self._send_buffer(buf)
+
+    def _send_buffer(self, buf):
+        LOGGER.debug("write %s", buf.hex())
```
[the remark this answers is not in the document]
Serphentas: Fixed

liquidctl/driver/coolit.py
Serphentas: Fixed

README.md
Serphentas: Done

Serphentas:
@aleksamagicka Changes done per review. I see macOS tests are failing but it's not related to this PR, cannot rerun them either. Any clue ?

aleksamagicka (approved):

aleksamagicka:
I reran them and the issue persists, I'm not sure why that's happening as I've had very little contact with Mac machines generally. Perhaps Github changed something about their runners.

Serphentas:
If you're OK with this peculiar situation, I suggest we move forward with the PR. We can ask Jonas for additional input if you prefer to be on the safe side.

aleksamagicka:
This doesn't seem to be tied to this PR specifically, I just reran it in [another PR](https://github.com/liquidctl/liquidctl/actions/runs/8571885560/job/24332961390?pr=690) and it fails as well. Seems like libusb or something like it is missing?

@jonasmalacofilho, tagging you since you surely know more about this than me :)

jonasmalacofilho:
The problem isn't related to this PR, but rather to a change in what the `macos-latest` runner points to. Very basically, we're now running on the M1/ARM architecture, and Python's `find_library` and homebrew don't play well there.

Homebrew has a patch for its Python formulas to fix this issue, but we're using `actions/setup-python` to install our Python interpreters, and that provide us with stock CPython. (Possibly even x86-64 Python, according to its documentation... but that would be too weird, so I'm guessing the docs are wrong).

There doesn't seem to be an obvious way to handle this... we want to look in the Homebrew library path, but we also need to decide whether to give it priority or not, and also check that the architecture matches (at least if that isn't the last path tried).

It's quite surprising that something like this was left for downstreams to solve. And it seems that some are resorting to shipping their own copies of the relevant dylibs...

Anyway, this is something I'll try to handle in PyUSB, so it really isn't related to liquidctl at all. In the meantime, @aleksamagicka, maybe we should pin our tests to run on Mac 12 (still Intel)? My reasoning is that outdated tests are better than no tests at all.

We could instead disable the affected test cases, but I think they are worth more than having all the others running on the new runner. The whole point of these affected tests is that they are the only ones that actually test (something) about the platform, all others are almost exclusively using mocks.

aleksamagicka:
> Anyway, this is something I'll try to handle in PyUSB, so it really isn't related to liquidctl at all. In the meantime, @aleksamagicka, maybe we should pin our tests to run on Mac 12 (still Intel)? My reasoning is that outdated tests are better than no tests at all.

I think we should, no reason not to let them run. (And as I understand, that's what they've been running on anyway up to now?)

jonasmalacofilho:
And they did detect a real issue as soon as they ran on the affected platform. : )

(Too bad it's something that, in hindsight, PyUSB me should have fixed a long time ago... I probably underestimated the number of Mac users running non-brew Python: if that was truly as rare as I thought, actions/setup-python would probably use brew Python too).

aleksamagicka:
> And they did detect a real issue as soon as they ran on the affected platform. : )

Touché :)

Maybe, when this is fixed, we can leave macos-12 in with macos-latest (at least, until macos-12 is deprecated) and cover both.

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,415 @@
...
+    @classmethod
+    def _missing_(cls, value):
+        LOGGER.debug("falling back to QUIET for _PumpMode(%s)", value)
+        return _PumpMode.QUIET
+
+
+def _sequence():
+    """Return a generator that produces valid protocol sequence numbers.
+
+    Sequence numbers are random.
```
jonasmalacofilho, line 88: This doesn't match the implementation.

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,415 @@
...
+    """
+    num = 1
+    while True:
+        yield (num % 31) + 1
```
jonasmalacofilho, line 92: Is this really needed for this device? In the protocol docs you said a simple counter is sufficient, and this is slightly more complex that that.

liquidctl/driver/coolit.py
```diff
@@ -0,0 +1,415 @@
...
+            0x1B1C,
+            0x0C04,
+            "Corsair H110i GT (experimental)",
+            {"has_pump": True, "fan_count": 2, "rgb_fans": False},
+        ),
+    ]
+
+    def __init__(self, device, description, fan_count, rgb_fans, **kwargs):
+        super().__init__(device, description, **kwargs)
+        self._component_count = 1 + fan_count * rgb_fans
```
jonasmalacofilho, line 125: This field and the `rgb_fans` argument are currently unused. How do you plan to use them in the future? And are there really any devices from this generation with RGB fans?

docs/developer/protocol/coolit.md
```diff
@@ -0,0 +1,84 @@
...
+
+The device expect packets in the following order/format:
+
+1. Request ID byte (usually a counter that rolls over)
+2. Opcode byte
+3. Command byte
+4. Optional arguments
+
+See the `_build_data_package` function for data package construction.
+
```
jonasmalacofilho, line 23: Can you add a quick description on the incoming/response packets?

It might also be good to mention that some commands/opcodes can be combined in a single packet (e.g. in `get_status`), and how that affects the response packet as well.

docs/coolit-guide.md
```diff
@@ -0,0 +1,51 @@
+# Corsair Hydro H110i GT AIO liquid cooler
+_Driver API and source code available in [`liquidctl.driver.coolit`](../liquidctl/driver/coolit.py)._
+
```
jonasmalacofilho, line 3:
```suggestion
_Driver API and source code available in [`liquidctl.driver.coolit`](../liquidctl/driver/coolit.py)._

_New in git._<br>

```

jonasmalacofilho:
Just a few minor notes (see above).

> We are good for review now, data dumps available to include in the database.

Please added the data dumps to <https://github.com/liquidctl/collected-device-data>.

jonasmalacofilho:
@aleksamagicka,

> Maybe, when this is fixed, we can leave macos-12 in with macos-latest (at least, until macos-12 is deprecated) and cover both.

Ack. I think they might not be necessary with the fix being made (and tested) in PyUSB. But since Intel Macs are much more relevant to our project than most (because of Hackintoshes), maybe the extra CI jobs are worth it.

liquidctl/driver/coolit.py
Serphentas: Per [product page](https://www.corsair.com/us/en/p/cpu-coolers/cw-9060026-ww/hydro-series-h110i-280mm-extreme-performance-liquid-cpu-cooler-cw-9060026-ww#tab-overview), there is at least one RGB channel that can be controlled. This feature is not currently supported, so I think it can be left for future implementation.

liquidctl/driver/coolit.py
Serphentas: Fixed, description matches behaviour.

README.md
```diff
@@ -120,6 +120,7 @@ subjective "from more to less liquid control-ly" order.
 | :--                | :-- | --: | :-: |
 | AIO liquid cooler  | [Corsair Hydro H80i GT, H100i GTX, H110i GTX](docs/asetek-690lc-guide.md) | <sup>_Ze_</sup> | 1.9.1 |
 | AIO liquid cooler  | [Corsair Hydro H80i v2, H100i v2, H115i](docs/asetek-690lc-guide.md) | <sup>_Z_</sup> | 1.9.1 |
+| AIO liquid cooler  | [Corsair Hydro H110i GT](docs/coolit-guide.md) | | git |
```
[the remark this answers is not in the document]
Serphentas: Fixed

docs/coolit-guide.md
Serphentas: Fixed

Serphentas:
@jonasmalacofilho 
> Please add the data dumps to https://github.com/liquidctl/collected-device-data.

Done in https://github.com/liquidctl/collected-device-data/pull/13

docs/developer/protocol/coolit.md
Serphentas: Not going to lie, I'm quite unsure as to how the exact reply works. I mostly adapted the existing driver from previous PR...

Added mention for multiple data packages though.

liquidctl/driver/coolit.py
jonasmalacofilho: The fact that random sequence numbers didn't work well suggests that some jumps in sequence numbers aren't acceptable to the device. Thus restarting the sequence at every invocation should also sometimes cause issues, and persisting the sequence in `storage` may be necessary after all.

While numbers modulo 32 seem to work, why not using the simpler full byte range (i.e. modulo 256), as seen in the captures? Well, actually, the full range except `0`, which doesn't appear to be used. Something like

```py
yield (num % 255) + 1
```

liquidctl/driver/coolit.py
jonasmalacofilho: Ack.

docs/developer/protocol/coolit.md
jonasmalacofilho: That's ok.

liquidctl/driver/coolit.py
Serphentas: We could use that but it implies testing the device again. Perhaps a quicker method (that's known to work) is to use the [old implementation](https://github.com/liquidctl/liquidctl/pull/147/files#diff-96821345dcd62e3b3b04ad2e571f1e2e53d79f4de38446c9ab40ef9f298f3b3cR91).

jonasmalacofilho: Fair point. And since this appears to work, and from sanmai's comment there that in general any per-command sequence seems to work, let's just keep things as is. At least for now/until bugs are reported...

jonasmalacofilho:
Thanks!

Serphentas:
Finally ! Thank you all, especially @CRCinAU for the many hours spent reversing and debugging :)

PR #649, opened 2023-12-03
CoolerMaster MM730 mouse LED change support

sophipl:
CoolerMaster MM730 mouse LED support.
This is my first github pull request, please be kind.
If there is the will to add RGB mouse support to liquidctl, I will fulfill the checklist. 

Checklist:

- [ ] Adhere to the [development process]
- [ ] Conform to the [style guide]

### DIFF
--- a/liquidctl/cli.py
+++ b/liquidctl/cli.py
@@ -24,6 +24,7 @@ Device selection options (see: list -v):
 
 Animation options (devices/modes can support zero or more):
   --speed <value>                    Abstract animation speed (device/mode specific)
+  --brightness <value>               LED brightness (device/mode specific)
   --time-per-color <value>           Time to wait on each color (seconds)
   --time-off <value>                 Time to wait with the LED turned off (seconds)
   --alert-threshold <number>         Threshold temperature for a visual alert (°C)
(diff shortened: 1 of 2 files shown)


aleksamagicka:
> That's a valid concern. I'm honestly not sure if supporting mice isn't going a bit too for liquidctl. Arguably we already push things with LED-only support for some GPUs and motherboards, but at least for those we could _theoretically_ support hardware monitoring as well.

As far as code maintainership is concerned, it's just a HID device (yay for a common ground!), so not much different from pumps and the like. I personally don't have a problem with that. But mice could be viewed as stretching the _liquid_ in liquidctl.

Looking over the devices we support, all go into a PC case, perhaps that's where we shoud draw the line.

> have you considered using/contributing to the piper/ratbag projects instead? If you did, why would you prefer to control the mouse through liquidctl?

Python (and by extend, liquidctl) is much easier to extend than C/C++ projects, at least at first. The feedback loop is quick and you don't have to think about lots of stuff C/C++ force you to.

I don't want to drive away effort, quite the opposite - I think that this PR has merit (the protocol is outlaid) and reverse engineering is never easy. But I feel that piper/OpenRGB might be a better place for this kind of support. Unfortunately, I haven't worked on either - if I have, I'd already be typing up a skeleton to help :)

sophipl:
I suppose I will close this PR then.
I noticed that OpenRGB has a solution for this mouse, but it doesn't work in all respects.
Probably new firmware version.
I has however ability to set separate scroll and logo colors, which I utilized too.
This ability is not present in manufacturer software.

PR #651, opened 2023-12-08
readme: Add a note to the Git install step

aleksamagicka:
Add a note to the Git install step to refer the reader that they should set up the udev rules, as outlined in the following section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,8 @@ python -m pip install git+https://github.com/liquidctl/liquidctl#egg=liquidctl
 [git]: https://git-scm.com/
 [pip]: https://pip.pypa.io/en/stable/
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -363,6 +363,8 @@ python -m pip install git+https://github.com/liquidctl/liquidctl#egg=liquidctl
 [git]: https://git-scm.com/
 [pip]: https://pip.pypa.io/en/stable/
 
+To avoid getting permission related errors, you'll need to set up udev rules, as outlined in the next section.
```
jonasmalacofilho, line 366:
```suggestion
To set up rootless access to devices on Linux and BSDs, install documentation and completions, or for auxiliary scripts, continue reading for the next few sections.
```

PR #659, opened 2024-01-03
Fix various issues with Unix file permission handling in `liquidctl.keyval`

intelfx:
The immediate issue with `liquidctl.keyval` is that it hardcodes Unix permission bits and thus does not play nice with (default) POSIX ACLs, if any are set on `/run/liquidctl`. This, in turn, makes it impossible to use liquidctl as more than one user (even if the necessary permissions are granted on the device nodes).

Normally, `/run/liquidctl/*/*` (e. g. `/run/liquidctl/vid1b1c_pid0c10/loc2`) is created with restricted 0o1700 permission bits, which is reasonable. However, if one desires to grant read/write permissions to these files to more than one user, it is impossible to do so with default ACLs (e. g. `setfacl -d g:liquidctl-users:rwX /run/liquidctl`) because the 0o1700 bits are forcibly set with `os.chmod()`, clearing the POSIX ACL mask value (which is overlaid with group permissions) and invalidating any permission-granting ACLs.

To fix this, we need to work with the permission bits in a semantically correct way. We apply the access restriction by adjusting the process umask value (which is overridden by default ACLs, if any are applicable) prior to file creation. Then, to set the sticky bit, we OR it into the resulting file's permissions without assuming their value.

Additionally, this fixes broken access flags handling in `_open_with_lock()` (as written, it causes all files to be opened with `w+` access string) and sets the mode to `0o666` to avoid creating executable files (as per above, we do not set `0o644` or `0o600` directly in the `os.open()` call, instead letting umask and default ACLs do their job).

---

Fixes: (multiple commits; see individual commit messages)

---

Checklist:

### DIFF
--- a/liquidctl/keyval.py
+++ b/liquidctl/keyval.py
@@ -71,6 +71,9 @@ def _open_with_lock(path, flags, *, shared=False):
 
         yield f
 
+        if access != os.O_RDONLY:
+            f.flush()  # ensure flushing before automatic unlocking
+
 
 if {'umask', 'stat', 'chmod'} <= os.__dict__.keys() \
         and {os.stat, os.chmod} <= os.supports_fd:
@@ -165,7 +168,6 @@ class _FilesystemBackend:
 
         with _open_with_lock(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC) as f:
             f.write(data)
-            f.flush()  # ensure flushing before automatic unlocking
 
         _LOGGER.debug('stored %s=%r (in %s)', key, value, path)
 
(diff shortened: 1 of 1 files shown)


intelfx:
@jonasmalacofilho fixed win32 compat, please take a look again

intelfx:
@jonasmalacofilho gentle ping

jonasmalacofilho:
@intelfx, hey, thanks for working on this, and sorry I haven't been able to look at it. Unfortunately I'm dealing with some family matters and still don't really have the time to look into it.

@aleksamagicka, any chance you can add this to your to-review list? If not, it's ok, I'll get to it eventually... I can see you're swamped too (unless I missed it, I think you haven't even submitted those last kernel changes upstream yet).

intelfx:
No worries at all. Take care!

aleksamagicka:
> I can see you're swamped too (unless I missed it, I think you haven't even submitted those last kernel changes upstream yet).

Yeah, I really should get around to that.

I'll try to review this as well, might have a few questions along the way as I haven't used this part of liquidctl before - doesn't seem too long.

aleksamagicka (approved):
From what I see, LGTM, thanks

jonasmalacofilho:
Thanks!

PR #683, opened 2024-03-18
add yoda support for --unsafe flag

aapo-kossi:
This is a simple addition that allows `extra/yoda.py` to be used with devices behind the `--unsafe` check. Currently relevant to my knowledge are only two MSI coreliquid devices.

`extra/yoda` currently doesn't seem to have any tests, but I have now also verified that `--unsafe` actually functions as intended on real hardware, both with the cli and `extra/yoda` by temporarily adding the unsafe requirement to my K360 in my local environment.

The small driver change is needed so that the user input unsafe option doesn't accidentally override the device setting from MATCHES. This previously caused any non-empty `--unsafe` option to pass instead of needing the intended `experimental_coreliquid_cooler`. This is unrelated to yoda (sorry), but I noticed it during the hardware testing mentioned above.

Related: #564
---

Checklist:

### DIFF
--- a/extra/yoda.py
+++ b/extra/yoda.py
@@ -48,7 +48,7 @@ Options:
   --bus <bus>              Filter devices by bus
   --address <address>      Filter devices by address in bus
   --usb-port <port>        Filter devices by USB port in bus
-  --unsafe <flags>         Specify comma separated unsafe flags for use with unstable drivers (some coreliquid coolers)
+  --unsafe <features>         Comma-separated bleeding-edge features to enable
   -v, --verbose            Output additional information
   -g, --debug              Show debug information on stderr
   --legacy-690lc           Use Asetek 690LC in legacy mode (old Krakens)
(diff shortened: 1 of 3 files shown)


extra/yoda.py
```diff
@@ -48,6 +48,7 @@
   --bus <bus>              Filter devices by bus
   --address <address>      Filter devices by address in bus
   --usb-port <port>        Filter devices by USB port in bus
+  --unsafe <flags>         Specify comma separated unsafe flags for use with unstable drivers (some coreliquid coolers)
```
jonasmalacofilho, line 51: While it's being ported to yoda because of the untable coreliquid coolers, the option is (or may be) useful with other devices too. I think the description from the CLI is good/better:

```suggestion
  --unsafe <features>         Comma-separated bleeding-edge features to enable
```

jonasmalacofilho (commented):
Sorry for the delay. Functionally it looks good, just one comment about the description in the `--help` message.

jonasmalacofilho:
Thanks!

Issue #624, opened 2023-08-18
Please add support for Kraken 240 


marcomistretta:
### Device type

AIO liquid cooler, Kraken 240 (no rgb version)

### Product page

https://nzxt.com/product/kraken-240

### First-party software

NZXT CAM

### What monitoring functionality does the device support?

temperatures, fan/pump speeds or duty cycles

### What configuration functionality does the device support?

fan/pump speeds or duty cycles, lighting of embedded LEDs, lighting of accessories like RGB fans or LED strips

### Can you help with implementing or testing the changes needed for liquidctl to support this device?

testing changes on Linux

aleksamagicka:
You didn't list the USB product ID, so I'll guess that it's `0x300E`. Support for that landed in https://github.com/liquidctl/liquidctl/pull/605. Can you please try and see if it works for you?

Silverwolf-3D:
Does this help I did a lsusb 
`Bus 001 Device 002: ID 1e71:2012 NZXT NZXT RGB Controller`
`Bus 003 Device 002: ID 1e71:300e NZXT NZXT Kraken Base`
I got one also, but I am a linux newbie so hope this helps.

lanceofwhichwedream:
Going to +1 this thread, looks like on windows the Kraken 240 presents itself like this

> USB\VID_1E71&PID_300E&MI_00
Vendor id 1e71
Product id 300e

I'd be super down to test on windows and I can provide any additional information you might need @aleksamagicka

aleksamagicka:
Support for that Kraken in liquidctl has been merged in [October](https://github.com/liquidctl/liquidctl/commit/846ac5f5aa9d5484e8ed1d549fd2478bbf51fdd8). Please retry with the [Git version](https://github.com/liquidctl/liquidctl#installing-from-pypi-or-github) for the time being.

lanceofwhichwedream:
Will do, thank you for the info. I'll report back if it was successful.

renhiyama:
@lanceofwhichwedream can you report back whether you got it working successfully or not? I need to buy a water cooler, and this cooler costs $30 less than kraken x53, so if its successful, I could be getting a better product with more features (lcd) and $30 cheaper 😅

naggie:
I can confirm it works on fb697e05c00a9ed8818d88a25d6202f46782d811 (latest master as of yesterday)

...however, after overnight using the python library to control pump and fan speed I get a "device has no longid" error

[...]

Issue #631, opened 2023-10-12
Firmware 2.0.1 breaks Kraken 2023 support


jonasmalacofilho:
### Describe the bug

The Kraken 2023 (non elite, non rgb) has had a firmware update (2.0.1 as reported by CAM) .   (I think the previous version was 1.2 if I remember correctly.)  CAM automatically upgraded itself without asking, so there was no choice.

Since then, setting an image using liquidctl on linux has given this error: 
[ERROR] (cli) (log): NZXT Kraken 2023: unexpected error: AssertionError('missing messages (attempts=12, missing=1)')

All other functions of liquidctl seem to be unaffected.

_Originally posted by @sossw in https://github.com/liquidctl/liquidctl/issues/605#issuecomment-1745536722_

### Commands executed

N/A

### Output of all relevant commands with `--debug` flag

N/A

### Affected device

Kraken 2023 (non elite, non rgb)

### Operating system and version

N/A

### Installation method

N/A

### Version of liquidctl

one of the final states of PR #605

jensmeindertsma:
i have arch installed on a system with this cooler but haven't been able to install this PR😬 not familiar enough with AUR anyone got me some instructions?

aniongithub:
I have this cooler with the 2.0.1 firmware (running on Ubuntu 22.04) and have encountered this bug.

I can run/debug liquidctl, see my initial comments about this issue [here](https://github.com/liquidctl/liquidctl/pull/605#issuecomment-1739802897) and [here](https://github.com/brokenmass/AIOLCDUnchained/issues/4#issue-1920339497)

Happy to help however I can!

jklint:
[debug.log](https://github.com/liquidctl/liquidctl/files/13110863/debug.log)

Hitting this as well with 2.0.0 firmware on Manjaro. Attached debug log for reference.

Serphentas:
@jklint what command were you trying to run during the capture ? If possible, include the image you were trying to upload.

Any other dumps welcome as well !

sophipl:
Hi

I have firmware version 2.0.0 and have the same problem
I tried gifs, static images (png and jpeg, stored from gimp), multiple images, multiple gifs, nothing works

```
python3 -m liquidctl --match kraken initialize
NZXT Kraken 2023 (experimental)
├── Firmware version    2.0.0  
├── LCD Brightness         40  %
└── LCD Orientation         0  °
```

[...]

PR #688, opened 2024-04-04
Add firmware warning after experience with Kraken 240

naggie:
This PR adds a suggestion to update the firmware before raising a bug to the
readme.

Related: #624


aleksamagicka:
This is a kind of a double-edged sword, IMO, as we have examples of (incidentally) Kraken devices behaving badly/being unsupported with newer/latest firmware versions (#631).

jonasmalacofilho:
I agree with @aleksamagicka. 

What about having the driver try to detect the issue and, in that case, make the suggestion to the user?

PR #689, opened 2024-04-05
kraken3: Improve LCD screen setting checks

aleksamagicka:
#673 erroneously disallowed Kraken 2023 Elite devices from setting the LCD screen. Also, it introduced a bug because `self.fw` is available during initialization and not populated after. This PR fixes the overall logic.

Related: #631

### DIFF
--- a/liquidctl/driver/kraken3.py
+++ b/liquidctl/driver/kraken3.py
@@ -772,10 +772,10 @@ class KrakenZ3(KrakenX3):
             self.orientation = msg[0x1A]
 
         # Firmware 2.0.0 and onwards broke the implemented image setting mechanism for Kraken 2023
-        # (standard and elite). In those cases, show an error until issue #631 is resolved.
+        # (non-elite). In those cases, show an error until issue #631 is resolved.
         def unsupported_fw_version():
             device_product_id = self.bulk_device.product_id
-            if device_product_id in (0x300C, 0x300E) and self.fw[0] == 2:
+            if device_product_id == 0x300E and self.fw[0] == 2:
                 _LOGGER.error(
                     "setting images is not supported on firmware 2.X.Y, please see issue #631"
                 )
--- a/liquidctl/driver/kraken3.py
+++ b/liquidctl/driver/kraken3.py
@@ -26,7 +26,7 @@ if sys.platform == "win32":
     from winusbcdc import WinUsbPy
 
 from liquidctl.driver.usb import PyUsbDevice, UsbHidDriver
-from liquidctl.error import NotSupportedByDevice
+from liquidctl.error import NotSupportedByDevice, NotSupportedByDriver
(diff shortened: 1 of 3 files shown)


aleksamagicka:
Should I add a note [in the docs](https://github.com/liquidctl/liquidctl/blob/main/docs/kraken-x3-z3-guide.md#the-lcd-screen-only-z-and-2023-models)?

aleksamagicka:
@maxawake, here it is, please test on your side.

liquidctl/driver/kraken3.py
```diff
@@ -647,6 +647,12 @@ def _find_winusb_device(self, vid, pid, serial):
                 return True
         return False
 
+    def _get_fw_version(self, clear_reports=True):
+        if clear_reports:
+            self.device.clear_enqueued_reports()
+        self._write([0x10, 0x01])  # firmware info
+        self._read_until({b"\x11\x01": self.parse_firm_info})
+
```
jonasmalacofilho, line 655: Thinking about the API use case (in opposition to the more common CLI use case), I think it would be nice to store the (first) result and make sure that we wont keep fetching the firmware version from the device over and over again.

liquidctl/driver/kraken3.py
```diff
@@ -772,14 +778,16 @@ def parse_lcd_info(msg):
...
-                    "setting images is not supported on firmware 2.X.Y, please see issue #631"
-                )
-                return True
+            device_product_id = self.device.product_id
+            if device_product_id == 0x300E:
+                self._get_fw_version()
+                if self.fw[0] == 2:
+                    _LOGGER.error(
+                        "setting images is not supported on firmware 2.X.Y, please see issue #631"
+                    )
```
jonasmalacofilho, line 789: I think I missed it the previous time, but I think a `NotSupportedByDriver` error would be better.

liquidctl/driver/kraken3.py
aleksamagicka: Makes sense, but what if the firmware is upgraded in the meantime? It's an uncommon situation, just wondering.

jonasmalacofilho: That's a good point. But I think (even though it's probably not guaranteed anywhere) that our HID/USB handles would all become invalid anyway once the device is rebooted into upgrade mode.

aleksamagicka: Yeah, that's what I'm thinking as well, the devices usually disconnect. Let's cache it, then.

jonasmalacofilho: Yeah, because it's not only that the device reboots, but also that it generally in upgrade mode it generally comes back with completely different VID, PID and other descriptors.

jonasmalacofilho (approved):

PR #692, opened 2024-04-07
Kraken230 firmware 2 image support

sophipl:
Added ability to set images for Kraken with firmware version 2

Fixes: #631 
Closes: 
Related: 

---

Checklist:


sophipl:
So I wrote the solution to send the image to the AIO LCD.
However after usb reset/dissconnect, the image needs to be sent twice, without it no effect.
One solution is to send it twice all the time.

The manufacturer software sends it twice too

liquidctl/driver/kraken3.py
```diff
@@ -788,8 +789,15 @@ def check_unsupported_fw_version():
                 self._get_fw_version()
                 if self.fw[0] == 2:
                     raise NotSupportedByDriver(
-                        "setting images is not supported on firmware 2.X.Y, please see issue #631"
+                        "gif images is not supported on firmware 2.X.Y, please see issue #631"
```
aleksamagicka, line 792: `is` -> `are` with this sentence.

Also, #631 would be resolved by merging this PR. Is GIF setting not supported at all, or just not implemented right now? Either way, the message should be changed if that issue is to be closed.

liquidctl/driver/kraken3.py
```diff
@@ -631,6 +631,7 @@ def __init__(
         # 0 = Normal, 1 = +90 degrees, 2 = 180 degrees, 3 = -90(270) degrees
         self.orientation = 0
         self.brightness = 50  # default 50%
+        self.fw = None
```
aleksamagicka, line 634: `self._fw` is already set to None in KrakenX3, and was renamed to that in https://github.com/liquidctl/liquidctl/commit/255821125c07eabfdfde71e14b7ea0d7e629dd7d. Please rebase this PR, as I did that a few days after you here.

liquidctl/driver/kraken3.py
```diff
@@ -885,6 +923,35 @@ def prepare_frames(frames):
 
         return result_bytes.getvalue()
 
+    def _send_data_fw2(self, data, bulkInfo):
```
aleksamagicka, line 926: I'd add an explicit comment that this is for firmware 2.X, and rename to `_send_elite_data_fw2()` or similar. (Yes, it's obvious from the name now, but still.)

liquidctl/driver/kraken3.py
```diff
@@ -788,8 +789,15 @@ def check_unsupported_fw_version():
                 self._get_fw_version()
                 if self.fw[0] == 2:
                     raise NotSupportedByDriver(
-                        "setting images is not supported on firmware 2.X.Y, please see issue #631"
+                        "gif images is not supported on firmware 2.X.Y, please see issue #631"
                     )
+        
+        def is_fw_version2():
```
aleksamagicka, line 795: This does the same thing as check_unsupported_version(), minus the logging. Since you're now using that in only one place, I'd suggest to replace that call with this and log + return if necessary.

Also the name is ambigious, I'd call it `_is_elite_fw_version2()` or similar (but prefixed with an underscore, since it's private).

liquidctl/driver/kraken3.py
```diff
@@ -851,6 +867,28 @@ def _prepare_static_file(self, path, rotation):
...
+            Image.open(path)
+            .resize(self.lcd_resolution)
+            .rotate(rotation * -90)
+            .convert("RGB")
+            .getdata()
+        )
+        result = []
+        pixelDataIndex = 0
+        for pixelDataIndex in range(0, len(data)):
+            dr = (data[pixelDataIndex][0] >> 3)
```
aleksamagicka, line 886: Run `black` on the file to fix the linter CI.

liquidctl/driver/kraken3.py
```diff
@@ -806,9 +814,16 @@ def check_unsupported_fw_version():
...
             return
         elif mode == "static":
-            check_unsupported_fw_version()
-            data = self._prepare_static_file(value, self.orientation)
-            self._send_data(data, [0x02, 0x0, 0x0, 0x0] + list(len(data).to_bytes(4, "little")))
+            if is_fw_version2():
+                data = self._prepare_static_file_rgb16(value, self.orientation)
+                self._send_data_fw2(data, [0x06, 0x0, 0x0, 0x0] + list(len(data).to_bytes(4, "little")))
+                # sending it twice is only required once after initialization
+                # the same behaviour is observed in manufacturer at init software
```
aleksamagicka, line 821: Nit: `at software init`?

liquidctl/driver/kraken3.py
```diff
@@ -821,6 +836,7 @@ def check_unsupported_fw_version():
         elif mode == "liquid":
             self._switch_bucket(0, 2)
             return
+
```
aleksamagicka, line 839: Line break probably not required, but I guess `black` will resolve.

aleksamagicka (changes requested):

aleksamagicka:
Also:

- Docs should be updated to note that GIFs are not supported on 2.X on 2023 models
- I guess `test_krakenz3_screen_not_totally_broken_part2()` should be updated to cover this case? (Although it's broken for other reasons right now)

Thanks!

liquidctl/driver/kraken3.py
sophipl: More firmware init, anytime usb is disconnected or device rebooted

liquidctl/driver/kraken3.py
sophipl: I did not manage to trigger gif handing in manufacturer software, it sends frames

liquidctl/driver/kraken3.py
sophipl: This is to determine how to send the image data, so throwing an exception would exclude firmware 1.X.Y, unless caught

sophipl:
I added potential fix for `test_krakenz3_screen_not_totally_broken_part2` case
However this testcase is not related to this pull request since it is about gif handling

tests/test_krakenz3_response.py
```diff
@@ -1,4 +1,38 @@
...
+    img.save(
+        result_bytes,
+        format="GIF",
+        interlace=False,
+        save_all=True,
+        append_images=list(frames),
+        loop=0,
+    )
+    return result_bytes.getvalue()
+
```
aleksamagicka, line 29: Are these line breaks needed?

liquidctl/driver/kraken3.py
aleksamagicka: I was referring to [this part](https://github.com/liquidctl/liquidctl/blob/3aba011146eb0432b9db786ecd2fcc94e5f16fc8/liquidctl/driver/kraken3.py#L833). Now that you have `_is_2023_fw_version2()`, use it there instead of `check_unsupported_fw_version()` and raise the same exception. That way, the [fw checking code ](https://github.com/liquidctl/liquidctl/blob/3aba011146eb0432b9db786ecd2fcc94e5f16fc8/liquidctl/driver/kraken3.py#L784-L800) is not duplicated.

tests/test_krakenz3_response.py
```diff
@@ -1,4 +1,38 @@
 # fmt:off
+from PIL import Image, ImageSequence
+import os
+import io
+""" 
+Required for different library versions
+"""
+def load_rgb_gif():
```
aleksamagicka, line 8: This is doing the same thing as the driver, right? And then verifying if it's the same - IMO this defeats the purpose of the test. It should contain the hardcoded values, so we're sure that those are generated and sent. (But that got disrupted by pillow changing the way they generate the output.)

@jonasmalacofilho, thoughts?

sophipl: There is another way of thinking about it, everything is static except the data that library is producing, and we are not testing library.
Granted, the way that we are using the library might be wrong, but it is proven to work that way.

liquidctl/driver/kraken3.py
sophipl: I see

tests/test_krakenz3_response.py
sophipl: I can revert this if you like.
But even if we generate gif file, there is no guarantee that gif will be same between versions.

jonasmalacofilho: Actually, what @sophipl just wrote gave me an idea...

During the test, let's inject an already appropriately massaged image into the driver at the right place. This way, when testing the driver, pillow (or whatever we choose to use as a image manipulation library) doesn't matter.

Ideally we also want to check that the driver _tries_ to resize the image, so maybe do something like:

- the driver calls a "normalize image" function, which normally uses pillow, and that can be easily tested in isolation (supply it with a pillow-generated GIF of the wrong size, and have its output match the pillow-generated equivalent GIF of the right size);
- when testing the rest of the driver, specifically the `set_screen` functionality, the "normalize image" is monkeypatched to just return an already well formed GIF that the driver can just split and send to the (mock) device.

This would mean retaining a binary blog in the repository, which (as mentioned before) is not ideal. So let's keep this binary blob as simple as possible: like, 3 frames, each of a solid base color. Something that's really simple to audit; it doesn't have to look good, it just has to produce a corresponding golden sample of traffic that's useful enough for us to have some confidence in the driver.

sophipl: I will revert this commit then.
It seems that bigger test rewrite is in order with mocking the library if I understand the idea.
Unfortunately new firmware does not support gifs, as far as I know, so the test is related to the functionality added.

sophipl:
Anything else I need to do?

sophipl:
@aleksamagicka @jonasmalacofilho please tell me what I need to do to merge this change?
I believe I addressed all @aleksamagicka points.
Sorry, I'm new to this :)

liquidctl/driver/kraken3.py
```diff
@@ -807,12 +803,26 @@ def check_unsupported_fw_version():
...
-            self._send_data(data, [0x02, 0x0, 0x0, 0x0] + list(len(data).to_bytes(4, "little")))
+            if _is_2023_fw_version2():
+                data = self._prepare_static_file_rgb16(value, self.orientation)
+                self._send_2023_data_fw2(
+                    data, [0x06, 0x0, 0x0, 0x0] + list(len(data).to_bytes(4, "little"))
+                )
+                # sending it twice is only required once after initialization
+                # the same behaviour is observed in manufacturer at init
+                # some soft of framebuffer swapping?
+                self._send_2023_data_fw2(
```
aleksamagicka, line 814: There's probably no way to reliably track if we're here just after an init, right?

Otherwise, I think it LGTM.

sophipl: It pains me to say this, but I didn't found a way to track this.
When manufacturer software loads, it sends the image twice at load.
After changing to different image, it sends it twice.
I tried to find any pattern that would indicate that second attempt is needed, but I failed.

sophipl:
I don't know if I'm supposed to do that, but I checked all items in the task, since LGTM received.

aleksamagicka:
No need to select things that are not applicable to your PR, for example, updating the man pages if it's not needed (and I don't think it is, here).

docs/kraken-x3-z3-guide.md
```diff
@@ -172,6 +172,8 @@ _New in git._<br>
 
 Adds support for NZXT Kraken 2023 Standard, Elite
 
+*For firmware version 2 gif screen mode is no longer supported*
```
aleksamagicka, line 175: Maybe link the issue #631 here, and do say that this only applies to 2023 models.

sophipl:
> No need to select things that are not applicable to your PR, for example, updating the man pages if it's not needed (and I don't think it is, here).

So edit PR and remove items now?

docs/kraken-x3-z3-guide.md
sophipl: How do you link issues? just hashtag in code?

aleksamagicka: Yes, something like `*On 2023 models (standard and ELite), GIF screen mode is no longer supported for firmware versions 2.X (see #631).*

aleksamagicka:
> > No need to select things that are not applicable to your PR, for example, updating the man pages if it's not needed (and I don't think it is, here).
> 
> So edit PR and remove items now?

I don't think there's a hard and fast rule... I just leave them unchecked if they do not apply.

docs/kraken-x3-z3-guide.md
aleksamagicka: Actually wait, I think the syntax is different.

sophipl: Well, I copied this already, and it's linked
I can change it of course

aleksamagicka: This works:

```
*On 2023 models (standard and Elite), GIF screen mode is no longer supported for firmware versions 2.X (see [#631][`issue-631`]).*
```

With this at the end of the file (it already has a reference to liquidtux there, so below that):

```
[`issue-631`]: https://github.com/liquidctl/liquidctl/issues/631
```

Sorry, I forgot that it doesn't automatically do that for md files.

> Well, I copied this already, and it's linked

In the PR view it is, but not when you access the file [normally](https://github.com/liquidctl/liquidctl/blob/c77f20306b071d97c2b92069f7c6850bca953770/docs/kraken-x3-z3-guide.md#the-lcd-screen-only-z-and-2023-models).

swensorm:
Is this waiting for the comment on line 205 of the md guide to be removed, or trying to find a solution to the double image send?

Looking forward to this getting merged to main 🙏

sophipl:
> Is this waiting for the comment on line 205 of the md guide to be removed, or trying to find a solution to the double image send?
> 
> Looking forward to this getting merged to main 🙏

I think we are waiting for @jonasmalacofilho review?

jonasmalacofilho:
LGTM, thanks!

jonasmalacofilho (approved):

PR #715, opened 2024-06-15
Draft: Avoiding excessively read/write in the device when using yoda

vandersonmr:
I have a icue h150i elite and when using yoda to control the pump and fans based on CPU temperature, the RGB blinks all the time. 

It blinks every time the device is touched (read_sensors or apply_duty). I did some small changes that made it blink much less in such way that it is acceptable.

1) Just read sensors in device with the sensors we want to read are from the device (In my use case, it comes from the CPU so I don't need to have the blink from reading the device sensors).

2) Just apply duty when the duty changed. If the temperature of the sensor is stable, for example, there is no need to stay setting the same fan speed all the time. 

Obs.: This was useful to me. I just built my PC today and tried the script. I'm creating this PR as a draft to discussion, as I didn't check yet the style guide... or added tests...

### DIFF
--- a/extra/yoda.py
+++ b/extra/yoda.py
@@ -105,12 +105,19 @@ elif sys.platform.startswith("linux") or sys.platform.startswith("freebsd"):
         psutil = None
 
 
-def read_sensors(device, **kwargs):
+devices_sensors = None
+def read_sensors(device, sensors_to_read, **kwargs):
+    global devices_sensors
+    if not devices_sensors:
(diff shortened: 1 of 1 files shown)


vandersonmr:
@jonasmalacofilho do you think this is useful?

jonasmalacofilho:
I haven't checked for corner cases yet, but I think both optimizations are worthwhile.

PR #743, opened 2024-11-14
usb: PyUsbDevice addresses need to be strings

nmalaguti:
The cli treats addresses as strings, but the USB address is an integer. It needs to be converted to a string for the comparison to work.

This also changes the output from `--json status`.

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/usb.py
+++ b/liquidctl/driver/usb.py
@@ -376,7 +376,7 @@ class PyUsbDevice:
 
     @property
     def address(self):
-        return self.usbdev.address
+        return str(self.usbdev.address)
 
     @property
     def port(self):


jonasmalacofilho:
Nice catch.

Maybe

```diff
diff --git a/liquidctl/driver/usb.py b/liquidctl/driver/usb.py
index 27ff346f0f0a..fecacc283eae 100644
--- a/liquidctl/driver/usb.py
+++ b/liquidctl/driver/usb.py
@@ -376,7 +376,7 @@ class PyUsbDevice:
 
     @property
     def address(self):
-        return self.usbdev.address
+        return str(self.usbdev.address)
 
     @property
     def port(self):
```

could be a better fix for it, though.

Is it still enough to fix the real world issue for you?

nmalaguti:
> Is it still enough to fix the real world issue for you?

I'll test it out, but looking at the code almost certainly yes (and that's a better fix too).

nmalaguti:
Confirmed the fix worked. It might be a small breaking change because the `--json` output now outputs the address as a string instead of a number.

jonasmalacofilho:
> It might be a small breaking change because the `--json` output now outputs the address as a string instead of a number.

Forgot about that... and it also breaks the public API. And even though the change is small, it will likely require  downstream changes anyway, because listing and filtering devices is a very basic feature.

I'll take your original fix to avoid the unnecessary breaking changes. Thanks!

(I'll manually merge your original commit).

PR #769, opened 2025-03-02
Fix 'wings' mode in SmartDevice

pferreir:
Using `wings` mode currently fails with an `IndexError`. I didn't run the automated tests since it's just two lines and it's broken in its current state anyway.

---

Checklist:

- [x ] Adhere to the [development process]
- [x ] Verify that the changes work as expected on real hardware

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/smart_device.py
+++ b/liquidctl/driver/smart_device.py
@@ -643,7 +643,7 @@ class SmartDevice2(_BaseSmartDevice):
             for [g, r, b] in colors:
                 self._write([0x22, 0x10, cid])  # clear out all independent LEDs
                 self._write([0x22, 0x11, cid])  # clear out all independent LEDs
-                color_lists = [] * 3
+                color_lists = [0] * 3
                 color_lists[0] = [g, r, b] * 8
                 color_lists[1] = [int(x // 2.5) for x in color_lists[0]]
                 color_lists[2] = [int(x // 4) for x in color_lists[1]]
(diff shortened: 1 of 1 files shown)


liquidctl/driver/smart_device.py
```diff
@@ -643,7 +643,7 @@ def _write_colors(self, cid, mode, colors, sval, direction='forward',):
             for [g, r, b] in colors:
                 self._write([0x22, 0x10, cid])  # clear out all independent LEDs
                 self._write([0x22, 0x11, cid])  # clear out all independent LEDs
-                color_lists = [] * 3
+                color_lists = [0] * 3
```
jonasmalacofilho, line 646: Looking back into how the broken line of coded ended up here (mea culpa), it seems that the intention was to create a list of *four* empty lists:

```suggestion
                color_lists = [[]] * 4
```

The first three lists would be filled in, but the forth would be left empty.

liquidctl/driver/smart_device.py
```diff
@@ -653,7 +653,7 @@ def _write_colors(self, cid, mode, colors, sval, direction='forward',):
                             0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x06,
                             0x05, 0x85, 0x05, 0x85, 0x05, 0x85, 0x00, 0x00,
                             0x00, 0x00, 0x00, 0x00])
-                    self._write(msg + color_lists[i % 4])
+                    self._write(msg + color_lists[i % 3])
```
jonasmalacofilho, line 656: For reasons unknown to me (don't have this particular device), it seems that the modulo 4 indexing was intentional, and that some writes should send an empty color list.

```suggestion
                    self._write(msg + color_lists[i % 4])
```

jonasmalacofilho (commented):
Can you please check if the two changes I suggested still work on the real hardware?

If they do, could you also please try to add a minimal test case so that we don't break this mode again in the future?

PR #794, opened 2025-06-03
Asus Rog Ryuo I 240 (first generation) basic PWM fan control

Bloodhundur:
Hi, I added a driver that supports basic fixed PWM fan control for the [Asus Rog Ryuo 240 AIO](https://rog.asus.com/ca-en/cooling/cpu-liquid-coolers/rog-ryuo/rog-ryuo-240-model/) (1st generation). It can also return the firmware version.

I couldn't figure out how to make it return other values such as liquid temperature as those sensor options weren't showing in armoury crate (possibly due to my method of connecting an external windows laptop to the AIO via USB). 

Just for the record I attempted to use the USB codes of the Ryujin driver to return those values without any luck as they all reported 0. I’m sharing this as a working driver for the ASUS ROG Ryuo I 240. I haven’t added tests or full documentation yet, but fan speed and firmware reading work. Hopefully helpful to anyone else who needs it.

PWM was the main focus since without said software to control the fans they would blast at 100% nonstop.

Image support is also unimplemented.

Fixes: 
Closes: 
Related: 

---

Checklist:

    - [x] `liquidctl.8` Linux/Unix/Mac OS man page

New CLI flag?

New device?

- [x] Add entry to the README's supported device list with applicable notes and `git` MRLV

New driver?

### DIFF
--- a/liquidctl/driver/asus_ryuo.py
+++ b/liquidctl/driver/asus_ryuo.py
@@ -30,15 +30,14 @@ class AsusRyuo(UsbHidDriver):
 
     def initialize(self, **kwargs):
         msg = self._request(*_REQUEST_GET_FIRMWARE)
-        fw_string = bytes(msg[2:]).split(b'\x00')[0].decode('ascii', errors='ignore')
+        fw_string = bytes(msg[2:]).split(b"\x00")[0].decode("ascii", errors="ignore")
         return [
-            (_STATUS_FIRMWARE, fw_string, ''),
+            (_STATUS_FIRMWARE, fw_string, ""),
         ]
 
     def get_status(self, **kwargs):
         _LOGGER.info("No status available")
 
-       
     def set_fixed_speed(self, channel, duty, **kwargs):
         if channel not in ("fans", "fan"):
             raise ValueError("Only 'fans' channel is supported")
--- a/tests/test_asus_ryuo.py
+++ b/tests/test_asus_ryuo.py
@@ -1,13 +1,15 @@
 import pytest
 
-from _testutils import MockHidapiDevice 
+from _testutils import MockHidapiDevice
 from liquidctl.driver.asus_ryuo import AsusRyuo
+
 
 @pytest.fixture
 def mockRyuo():
     device = AsusRyuo(_MockRyuoDevice(), "Mock Asus Ryuo I")
     device.connect()
-    return device 
+    return device
+
 
 class _MockRyuoDevice(MockHidapiDevice):
     def __init__(self):
@@ -40,19 +42,21 @@ class _MockRyuoDevice(MockHidapiDevice):
 
         return buf[:length]
 
+
 def test_initialize(mockRyuo):
     (firmware_status,) = mockRyuo.initialize()
 
     assert firmware_status[1] == "AURO0-S452-0205"
 
+
 @pytest.mark.skip(reason="TODO: implement expected status values")
 def test_status(mockRyuo):
     actual = mockRyuo.get_status()
 
-    expected = [
-   ]
+    expected = []
 
     assert sorted(actual) == sorted(expected)
+
 
 def test_set_fixed_speeds(mockRyuo):
     mockRyuo.set_fixed_speed(channel="fans", duty=40)


liquidctl/driver/asus_ryuo.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class AsusRyuoTest(UsbHidDriver):
+    """Driver for ASUS ROG Ryuo I 240 (fan control only)."""
+
+    _MATCHES = [
+        (0x0B05, 0x1887, "ASUS ROG Ryuo I 240", {}),
+    ]
+
+    feature_set_fixed_speed = True
```
jonasmalacofilho, line 31: This field is currently unused.

liquidctl/driver/asus_ryuo.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def initialize(self, **kwargs):
+        _LOGGER.info("No initialization required")
+
+    def get_status(self, **kwargs):
+        msg = self._request(*_REQUEST_GET_FIRMWARE)
+        fw_string = bytes(msg[2:]).split(b'\x00')[0].decode('ascii', errors='ignore')
+        return [
+            (_STATUS_FIRMWARE, fw_string, ''),
+        ]
```
jonasmalacofilho, line 41: Since the firmware version doesn't change often, it's better to return it from `initialize` instead of `get_status`.

jonasmalacofilho (commented):
Hey, thanks for the PR!

It's odd that Armory Crate isn't showing any status information. Perhaps there's more than one USB interface you need to pass through to the VM.

You could also try to look whether any part of the USB traffic looks like sensible values for temperatures or fan speeds/duty cycles.

Can you also add/update the docs?

jonasmalacofilho:
Running Armory Crate on the same machine shouldn't matter for the data coming in through the USB connection.[^1]

But after checking the lsusb data you linked (thanks for reminding of that thread) and the pinouts from the installation manual, I still feel that there should be at least fan speed present in the USB data.

I know you already checked it, but can you show me what's available in the "Fan Control" tab?

[^1]: I mean, I can see how Armory Crate could maybe disable related parts of itself if it couldn't "see" the pump tachometer data sent to the motherboard AIO_PUMP header, but that's quite unlikely to be done in practice, plus the cooler should still work in non-ASUS motherboards.

Bloodhundur:
I decided to just grab the SSD from my laptop and throw it into my PC and from what I can tell I don't think the rog ryuo actually reports any data from itself (other then the firmware) or its fans. Everything shown in the hardware monitor is from my motherboard values (I don't have an AIO Header only CPU Fan and CPU_OPT).

rogryuo3
rogryuo2
image

I don't see any rpm values anywhere for the fans connected to the AIO either, only the duty percent and the chart that converts duty to RPM. Additionally from what I remember the only USB data I found related to the fans was it sending out an payload % code to set the fan duty % every second.

But I'll double check the USBPcap data just to be sure.

Bloodhundur:
While checking I noticed armoury crate service is required to set the AIO speed via CPU temperature as it will just run at the last set fan speed without it.

And I still haven't found any report of AIO fan speed rpm anywhere.

Bloodhundur:
>
>You could try contacting ASUS and asking about it.
>

I've contacted ASUS support on the matter, I should be getting a response in 2 business days.

>
>Which brings to another question: doesn't ASUS have a hardware monitoring/control software anymore? Armory Crate seems to only focus on aesthetics...
>

There is a LiveDash software that does the same thing as Armoury Crate but in a much more minimal format. However, I can't get that to work. According to online forum the newest (1.05.02) version of it is broken and I've been trying to find an older working version with no luck.
image

>
>Also, what other options does the other dropdown menu (currently set to "Temperature") show?
>

I'm connected via laptop USB currently but even on PC the other options — besides temperature and fans (pc only) — are just Voltage and Frequency

image
image
image

Bloodhundur:
Update: I got livedash working but no liquid sensors are available

image

ghtyrant:
Thanks for your PR @Bloodhundur, that really saves my ears! Tested it on my machine, setting the fan speed works fine.

I booted into Windows to check Armoury Crate and I get the same results as you, no temperature values for the pump - or really any values.

jonasmalacofilho:
Yeah, sure, I'm comfortable merging it in its current state.

Thanks!

PR #746, opened 2024-11-22
Rudimentry support for the Kraken Elite 2024 RGB 

hyenatown:
Because the 2024 model Kraken Elite functions pretty similar to the Z models, most of what needed to be added was the device ID. This managed to allow pump/fan speed setting and changes to the LCD, but not much else. This simply makes the device work with liquidctl. Why "RGB" - is due to the additional changes to the pump housing and fan casing, which have more LEDs from what I can glean.

Closes: #745

---

Checklist:

- [x] Adhere to the [development process]
- [x] Conform to the [style guide]
- [x] Verify that the changes work as expected on real hardware
- [x] Verify that all (other) automated tests (still) pass
- [x] Update/add documentation
    - [x] README, with ["new/changed in" notes]
    - [x] applicable `docs/*guide.md` device guides, with ["new/changed in" notes]
    - [x] `liquidctl.8` Linux/Unix/Mac OS man page

New device?

- [x] Regenerate `extra/linux/71-liquidctl.rules` (instructions in the file header)
- [x] Add entry to the README's supported device list with applicable notes and `git` MRLV

[development process]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/process.md
[style guide]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/style-guide.md
["new/changed in" notes]: https://github.com/liquidctl/liquidctl/blob/main/docs/developer/process.md#newchanged-in-notes

PR #803, opened 2025-07-02
NZXT Kraken 2024 Elite RGB | Set RGB Channels

hyenatown:
I left out the ability to address the RGB on this unit initally. Once again, it's got overlap with the other Kraken models so it was a drop-in fix. (edit: ha ha ha) Differences between them is that with the lack of a logo channel, the address for the ring and radiator are different, but simple as swapping the strings around.

Fixes: N/A
Closes: N/A
Related: #746

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/kraken3.py
+++ b/liquidctl/driver/kraken3.py
@@ -80,6 +80,7 @@ _COLOR_CHANNELS_KRAKENX = {"external": 0b001, "ring": 0b010, "logo": 0b100, "syn
 _COLOR_CHANNELS_KRAKENZ = {
     "external": 0b001,
 }
+
 _COLOR_CHANNELS_KRAKEN2023 = {"ring": 0b001, "external": 0b010, "sync": 0b111}
 
 _HWMON_CTRL_MAPPING_KRAKENX = {"pump": 1}


hyenatown:
Good point. I'm mulling it over and what I've done previously wouldn't have an effect on any other model, but on doing _this_, I'm at a point where I don't like it. I'm going to clean it up and it's going to be more consistent when finished.

hyenatown:
I have an unknown thing happening.
When I posted this PR, it worked just fine! But at some point, initialization stopped working a short time after. So I'm also kind of lost because this prevents me from making other changes. I set it aside for a while because I couldn't figure out what was wrong, but now I'm looking at it again and it still doesn't make sense.
venv is 3.13.3. all deps are satisfied. I'm not sure what I'm overlooking.

When `_COLOR_CHANNELS_KRAKEN2023` was declared but empty (i.e. the way it was before I made any changes), it worked just fine.

```
[INFO] (cli) (log): detailed error: NZXT Kraken 2024 Elite RGB: unexpected error: AttributeError("'NoneType' object has no attribute 'value'")
  File "./liquidctl/cli.py", line 458, in main
    status = dev.initialize(**opts)
  File "./liquidctl/driver/kraken3.py", line 710, in initialize
    self._read_until({b"\x21\x03": self.parse_led_info})
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./liquidctl/driver/kraken3.py", line 468, in _read_until
    func(msg)
    ~~~~^^^^^
  File "./liquidctl/driver/kraken3.py", line 275, in parse_led_info
    found_logo = find(2, 0) == Hue2Accessory.KRAKENX_GEN4_LOGO
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./liquidctl/util.py", line 76, in __eq__
    return self.value == other.value
                         ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'value'
```

jonasmalacofilho:
@hyenatown, re-reading your original PR description, the failure also makes sense. There is no logo channel. Maybe the device sometimes leaves some garbage data in the places of the report where there are no (valid) channels, and by chance this had previously matched a valid HUE2 accessory ID?

hyenatown:
That's a start. I'll see what I can do.

hyenatown:
Oh yes, `acc_id` returns 0 instead of a valid ID when checking `found_logo`.
In that case, shouldn't `return Hue2Accessory(acc_id) if acc_id else None` return None, and run through `_missing_` like how it does with `found_ring`?

Anyway, both `found_ring` and `found_logo` would likely still be False, but getting to that result to begin with is borked.

[Note: It took me this long to understand what the traceback was trying to say.]

hyenatown:
It should be True, yes. I guess I meant that one snag was leading to the other ultimately, and prevented even a correct (failing) result, but I see how these are related.

...I've got a bit of reading up to do.  Thanks for the help.

jonasmalacofilho:
> EDIT: The header is not `0x21 0x01` like it should.
> 
> ```
> [DEBUG] (usb) (read): read 64 bytes: 21:03:4f:63:94:82:0e:98:30:15:00:00:00:00:02:1e:00:00:00:00:00:1b:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00
> ```

I don't remember the specifics of this device/driver to parse this in my head, but it could be simply an unrelated message. The cooler sends any different messages during initialization, that's why we read until we see the particular one we're interesting to parse.

Do we still get the `21:01:...` we're looking for? (Probably: yes, otherwise in #803 the execution wouldn't have reached that point).

jonasmalacofilho:
In case it's useful, and because it's something not everyone knows about: Python has a fairly usable debugger.

https://docs.python.org/3/library/pdb.html

PR #838, opened 2025-12-01
Aura LED minor fix

codifryed:
Fixes an error that happens when the Aura LED device returns an unexpected response:
```
list.append() takes exactly one argument (3 given)
```

The cause:
```
status.append("Unexpected reply for config", "", "")
```
which is missing the status tuple parenthesis.

This error only happens when the Aura LED device returns an unexpected response, such as sometimes right after boot.
Additionally a couple very small cleanup changes have been made.

---

Checklist:

New CLI flag?

New device?

New driver?

### DIFF
--- a/liquidctl/driver/aura_led.py
+++ b/liquidctl/driver/aura_led.py
@@ -200,11 +200,15 @@ class AuraLed(UsbHidDriver):
             num = 6  # number of bytes per record
             count = 1
             while start_index + num < _READ_LENGTH:
-                status.append((
-                    "Device Config: " + str(count),
-                    ", ".join("0x{:02x}".format(x) for x in data[start_index: start_index + num]),
-                    "",
-                ))
+                status.append(
+                    (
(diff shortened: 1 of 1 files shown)


jonasmalacofilho (commented):
I think the original code takes the wrong approach. In both cases, it seems that the unexpected reply is, well, unexpected, and indicates some type of failure, either on the device or the driver. So maybe an assertion/exception would be more appropriate than returning some type of placeholder data point. 

Additionally, one of the reasons for these unexpected replies might be that the driver doesn't clear the kernel-side incoming buffer before writing to the device and trying to read the response back (so the response to some other command may be read from the buffer). The API to clear the kernel-side buffer is `device.clear_enqueued_reports`.

Can you try to incorporate these other fixes into your PR? (If they indeed make sense).

Thanks for catching this, btw.

jonasmalacofilho (approved):
Everything looks great. And the test is not noise and very welcomed.

Thanks again!

Issue #259, opened 2021-01-03
Linux with hidapi-0.10.1 is showing RGB Fusion controller twice


aravance:
#### System details
* Distro: Arch Linux
* `uname -r`
> 5.10.3-zen1-1-zen
* ``pacman -Q liquidctl hidapi `pacman -Qi liquidctl | awk -F'[:<=>]' '/^Depends/ {print $2}'` ``
> liquidctl 1.4.2-4
> hidapi 0.10.1-1
> python 3.9.1-1
> python-docopt 0.6.2-9
> python-pyusb 1.1.0-3
> python-hidapi 0.10.1-1

### Steps to reproduce
`liquidctl list`

### Expected
> Device ID 0: Gigabyte RGB Fusion 2.0 8297 Controller (experimental)
> Device ID 1: Corsair H115i Platinum (experimental)

### Actual
> Device ID 0: Gigabyte RGB Fusion 2.0 8297 Controller (experimental)
> Device ID 1: Gigabyte RGB Fusion 2.0 8297 Controller (experimental)
> Device ID 2: Corsair H115i Platinum (experimental)

### Workaround
I downgraded to hidapi-0.9.0-2 from package cache, and the devices now show up only once.

## Hardware Configuration
* x570 Aorus Master rev1.1
* Corsair h115i RGB Platinum

jonasmalacofilho:
Can you please run

```
$ python -c "import hid; print('\n'.join(map(repr, hid.enumerate())))"
```

and post the output?

aravance:
Worth note, hidapi 0.10.1 introduced a change to handling of usages and pages: https://github.com/libusb/hidapi/releases

I reported a similar issue with OpenRGB here: https://gitlab.com/CalcProgrammer1/OpenRGB/-/issues/924

### 0.9.0-2

> {'path': b'/dev/hidraw0', 'vendor_id': 1165, 'product_id': 33431, 'serial_number': '', 'release_number': 3, 'manufacturer_string': 'ITE Tech. Inc.', 'product_string': 'ITE Device(8595)', 'usage_page': 51456, 'usage': 13354, 'interface_number': 0}
{'path': b'/dev/hidraw1', 'vendor_id': 4152, 'product_id': 4781, 'serial_number': '', 'release_number': 281, 'manufacturer_string': 'SteelSeries ', 'product_string': 'SteelSeries Arctis 7', 'usage_page': 12080, 'usage': 12336, 'interface_number': 5}
[...]

PR #901, opened 2026-06-15
msi: fix Coreliquid probe dropping the control interface

cne27:
The MSI MPG Coreliquid driver fails to detect the cooler on hidapi
backends that populate HID usage information (for example recent hidraw
on Linux): `liquidctl list` does not show the device, and direct
commands return "no device matches available drivers and selection
criteria".

`MpgCooler.probe()` skipped every handle on usage page 0x0001 (Generic
Desktop) to ignore redundant top level usages. On these backends,
however, the cooler exposes a single handle, and it is the control
interface itself, reported as usage page 0x0001 / usage 0x0000, so it
was dropped and the device disappeared.

Genuine extra top level usages carry a nonzero usage id (for example
0x0080 system control, 0x0006 keyboard), while the primary control
collection reports usage 0x0000. This qualifies the exclusion with
`usage != 0`, mirroring the precise (page, usage) approach already used
in `rgb_fusion2.py`: the redundant handles are still ignored, but the
control interface is kept.

Verified on real hardware (MPG Coreliquid K360, Linux/hidraw): before
the change `liquidctl list` omits the cooler; after it, the cooler is
detected and status, lighting, and fan commands work. The full test
suite passes (534 passed, 2 preexisting skips).

I could only test on Linux. The change widens which handles are accepted
on usage page 0x0001, so confirmation from a Windows user with an MSI
Coreliquid would be welcome. The HID usage tables and the existing
rgb_fusion2 handling both suggest the redundant Windows handles carry a
nonzero usage.

Related: #259

### DIFF
--- a/liquidctl/driver/msi.py
+++ b/liquidctl/driver/msi.py
@@ -26,6 +26,7 @@ from liquidctl.util import RelaxedNamesEnum, check_unsafe, clamp, u16le_from
 _LOGGER = logging.getLogger(__name__)
 
 EXTRA_USAGE_PAGE = 0x0001
+_CONTROL_USAGE = 0x0000
 _MAX_DATA_LENGTH = 185
 _PER_LED_LENGTH = 720
 _REPORT_LENGTH = 64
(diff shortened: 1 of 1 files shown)


cne27:
Pushed a fixup: The lint failure was just a black formatting nit in the if condition; black --check passes locally now.

jonasmalacofilho:
LGTM, thanks!

cne27:
Thanks for the review and quick merge!

PR #903, opened 2026-06-16
Add driver for Enermax Liqtech XTR pump-head LCD (USB 2e3c:0a12)

sungwookson:
## Summary

The Enermax Liqtech XTR AIO cooler has a numeric LCD on the pump head that accepts live data over USB HID. No vendor documentation or prior Linux/liquidctl support existed; the protocol was fully reverse-engineered by iterative black-box probing of the HID interface on a Proxmox Linux host (the Windows companion app could not be used as a reference because GPU temperatures were passed through to a VM, leaving the screen blank).

**Protocol (reverse-engineered):**
- HID report ID `0x20`, 64 data bytes
- Values are 16-bit big-endian, displayed verbatim (no unit conversion by device)
- Three writable display fields:

  | Field | Command | Byte offset | Display capacity |
  |-------|---------|-------------|------------------|
  | CPU temp | `0x01` | 5–6 | 4 digits |
  | GPU temp | `0x04` | 8–9 | 4 digits |
  | Pump RPM | `0x10` | 11–12 | 4 digits |

- A secondary CPU field (`0x02` @6–7) exists but its hardware renders only ~2 digits; the driver mirrors the primary CPU value there silently.
- Screen blanks if any field is not refreshed within ~2 seconds.
- No sensor data is readable from the device over USB (`get_status` returns `[]`).

## New file: `liquidctl/driver/enermax_liqtech_xtr.py`

Exposes three channels via `set_screen`:

```bash
liquidctl set cpu screen temperature 45
liquidctl set gpu screen temperature 2065   # GPU #2 at 65 °C  (id × 1000 + temp_c)
liquidctl set pump screen rpm 2900
```

The `id × 1000 + temp_c` convention for the GPU field encodes which GPU is currently shown in the leading digit, enabling multiple-GPU cycling from a daemon without needing a separate field.

`set_color`, `set_fixed_speed`, and `set_speed_profile` raise `NotSupportedByDriver` — the device has no USB-controlled LEDs and pump speed is PWM-controlled via the motherboard header.

## New file: `tests/test_enermax_liqtech_xtr.py`

30 unit tests covering all channels, byte offsets, report length, boundary values, validation errors, and unsupported-method errors.

## Tested on

- Proxmox VE 9 / Linux 6.x, Enermax Liqtech XTR 360 (USB `2e3c:0a12`)
- All three display fields confirmed working at 1 s refresh cadence

### DIFF
--- a/liquidctl/driver/__init__.py
+++ b/liquidctl/driver/__init__.py
@@ -43,6 +43,7 @@ from liquidctl.driver import nzxt_epsu
 from liquidctl.driver import rgb_fusion2
 from liquidctl.driver import smart_device
 from liquidctl.driver import ga2_lcd
+from liquidctl.driver import enermax_liqtech_xtr
 
 if sys.platform == 'linux':
     from liquidctl.driver import ddr4
--- a/liquidctl/driver/enermax_liqtech_xtr.py
+++ b/liquidctl/driver/enermax_liqtech_xtr.py
@@ -1,0 +1,151 @@
+# Copyright 2024  liquidctl contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""liquidctl driver for Enermax Liqtech XTR series liquid coolers.
+
+Protocol reverse-engineered from the USB traffic on Linux; no vendor
+documentation was available.  The device is a USB HID peripheral that
+controls the LCD display on the pump head of the Liqtech XTR AIO cooler.
+It does not expose any readable sensor data over USB.
+
+Copyright 2024  liquidctl contributors
+SPDX-License-Identifier: GPL-3.0-or-later
+"""
+
+import logging
+
+from liquidctl.driver.usb import UsbHidDriver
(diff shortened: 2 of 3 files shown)


Btrc4t:
I have an Enermax AquaFusion II ARGB 120 which also identifies over USB as 2e3c:0a12

The new functionality is very good, there are just a few issues:
* When displaying the CPU temperature, the *C annotation is most times missing, it however appears briefly in rare occasions. The annotation is displayed all of the time with the GPU temperature.
* This description does not match the observed behaviour on my device:  
```
The id × 1000 + temp_c convention for the GPU field encodes which GPU is currently shown in the leading digit, enabling multiple-GPU cycling from a daemon without needing a separate field.
```
On my device if I send 1080, I get 1080 displayed as the GPU temperature.
Edit: maybe I misunderstood that and you didn't mean to say that the screen somehow marks the different GPUs - is it only for our reference?

I'm open to try to help you with anything you suggest. Thanks for this!

Issue #864, opened 2026-02-23
Kraken PLUS 360: image upload fails because _delete_bucket expects 0x33:02 but device responds with 0xff:02


breadbored:
### Is there already an issue for this?

- [x] ~~There is no relevant open or closed issue for this~~

Apologies, I just saw #631 actually, but I'll drop more information about what does and doesn't work. Let me know if I can provide more info, or how I could contribute.

### Is this a request for support of a new device?

- [x] This is not a new device request

### Affected liquidctl version

liquidctl v1.16.0.dev62+gf9f0025d1 (Linux-6.18.9-arch1-2-x86_64-with-glibc2.43)

### Affected device

Listed as NZXT Kraken 2024 Plus, though this may be a different revision

### Is the device supported by that liquidctl version?

- [x] The device is listead in the README as supported by my liquidctl version

### Describe the bug

Image uploading fails. More info in the "Commands Executed" section

### Commands executed

`liquidctl` is able to run the following commands successfully:
- `liquidctl initialize`
- `liquidctl status`
- `liquidctl list`
- `liquidctl set lcd screen orientation 270`
- `liquidctl set lcd screen brightness 80`

The error happens with these commands:
- `liquidctl set lcd screen static test.png`
- `liquidctl set lcd screen gif test.gif`

The [`test.gif` can be found here](https://media3.giphy.com/media/v1.Y2lkPTc5MGI3NjExazh1bzFjbDBzb3k0d2wwbXFpNnl3bnBpdXBtZm40MXBvODduNnZxZyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9cw/y0BBlgKGFgDJK/giphy.gif), though any gif thus far has had the same result. [`test.png` as well](https://ezgif.com/save/ezgif-401ebd92d33471db.png)

### Output of all relevant commands with `--debug` flag

Animated GIF

```
$ liquidctl set lcd screen static ~/Downloads/mudkip-pokemon.gif --debug
[DEBUG] (cli) (_log_env_infos): script: /home/brad/.local/bin/liquidctl
[DEBUG] (cli) (_log_env_infos): version: 1.16.0.dev62+gf9f0025d1
[DEBUG] (cli) (_log_env_infos): platform: Linux-6.18.9-arch1-2-x86_64-with-glibc2.43
[...]

PR #912, opened 2026-07-15
fix: [#908] kraken 2024 plus lcd static

### LINKED ISSUE #908: NZXT Kraken 2024 Plus (1e71:3014) LCD static/gif upload fails with bucket delete timeout
### Is there already an issue for this?

- [x] There is no relevant open or closed issue for this

### Is this a request for support of a new device?

- [x] This is not a new device request

### Affected liquidctl version

liquidctl v1.17.0.dev5+g292dc5613

### Affected device

NZXT Kraken 2024 Plus

### Is the device supported by that liquidctl version?

- [x] The device is listead in the README as supported by my liquidctl version

## Device

NZXT Kraken 2024 Plus

USB ID:
0x1e71:0x3014

Firmware:
1.0.0

OS:
SteamOS / Linux 6.16.12-valve

liquidctl:
v1.17.0.dev5+g292dc5613

## Working

The following commands work:

- initialize all
- status
- lcd brightness
- lcd orientation
- lcd liquid

Examples:

liquidctl -m Kraken set lcd screen brightness 50

liquidctl -m Kraken set lcd screen orientation 90

liquidctl -m Kraken set lcd screen liquid

## Broken

Uploading static images fails:

liquidctl -m Kraken set lcd screen static /home/deck/test.png

Error:

AssertionError('missing messages (attempts=12, missing=1)')

## Debugging

The failure happens in kraken3.py during bucket handling.

_delete_bucket() always fails:

AssertionError('missing messages (attempts=12, missing=1)')

Even manual calls fail:

for bucket 0-15:
_delete_bucket(bucket)

_query_buckets() works and returns valid bucket data.

However:

_setup_bucket(0,1,[0x00,0x00],[0x01,0x00])

returns:

False

It looks like the Kraken 2024 Plus firmware accepts LCD control commands but rejects the bucket protocol used for uploading custom images.

### Commands executed

see discussion

### Output of all relevant commands with `--debug` flag

na

### liquidctl installation method

clone from git

### Operating system and version

SteamOS / Linux 6.16.12-valve

krupupakku:
Fixes the LCD static image upload failure on the NZXT Kraken 2024 Plus (1e71:3014). On this
device, the bucket-based transfer protocol used by other LCD-capable Krakens is not implemented
by the firmware: bucket queries and deletes never receive a real response, only continuous
unsolicited status broadcasts, which previously surfaced as an opaque
`AssertionError('missing messages...')`.

Static images are now sent through the same bucket-free transfer protocol already used for
Kraken 2023 units running firmware 2.X.Y, bypassing the non-functional bucket protocol entirely.
GIF mode remains unsupported on this device, for the same reason it is unsupported on 2023 units
running firmware 2.X.Y.

Verified working on real Kraken 2024 Plus hardware (SteamOS).

Fixes: #908
Closes: #908
Related: #864

---

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## Unreleased
+
+### Changes since 1.16.0
+
+Fixed:
+
+- Kraken X3/Z3: fix LCD static image upload on the Kraken 2024 Plus, whose firmware does not
+  respond to the bucket-based transfer protocol used by other LCD-capable Krakens; static images
+  are now sent through the same bucket-free protocol already used for Kraken 2023 units on
+  firmware 2.X.Y (liquidctl#908, liquidctl#864)
+
 ## [1.16.0] – 2026-03-03
 
 ### Changes since 1.15.0
--- a/docs/kraken-x3-z3-guide.md
+++ b/docs/kraken-x3-z3-guide.md
@@ -38,6 +38,13 @@ The functionality of the 2024 RGB AIO is identical to the 2023 model, retaining 
 
 The functionality of the 2024 RGB AIO is identical to the 2023 Standard model, with a 240x240 LCD.
 
+On this device, the bucket-based upload protocol used by other LCD-capable Krakens is not
+functional: bucket queries never receive a real response from the firmware, only continuous
+unsolicited status broadcasts (see [#864][`issue-864`] and [#908][`issue-908`]).  Because of this,
+`static` images are instead sent through the same bucket-free transfer path used by 2023 units on
+firmware 2.X.Y.  **GIF mode is not supported on this device** for the same reason it isn't
+supported on 2023 units running firmware 2.X.Y.
+
 
 ## Initialization
 
@@ -193,7 +200,8 @@ The LCD screen can be configured in a few different modes.
 Images and GiFs are automatically resized and rotated to match the device orientation.
 
 *Note that, on the 2023 models (Standard and Elite), the GIF screen mode is not currently supported
-on firmware versions 2.X (see [#631][`issue-631`]).*
+on firmware versions 2.X (see [#631][`issue-631`]).  The GIF screen mode is also not supported on
+the Kraken 2024 Plus (see [#864][`issue-864`] and [#908][`issue-908`]).*
 
 
 ## Interaction with Linux hwmon drivers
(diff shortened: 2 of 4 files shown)


ppk-museum:
Hit the same thing on a Kraken 2023 Elite (1e71:300c) on firmware 2.1.0, tested against 292dc56.

This device spams unsolicited status frames with prefix 75 02 on the response endpoint. If I do 12 reads with no command pending, I get 12 of those. So the single blind _read() in _write_then_read will happily hand back a broadcast instead of the reply to whatever was just written. You only notice when there's a gap between HID exchanges, and _prepare_gif_file() gives you a big one right before the bucket handshake.

Repro: do what set_screen does first (30 01 + read), sleep 2s to stand in for the image prep, then query buckets:

blind single read (current behaviour):
    queried 0 -> STATUS BROADCAST 7502, not a bucket reply
    queried 1 -> STATUS BROADCAST 7502, not a bucket reply
    queried 2 -> 3104 reports bucket index 0   <-- MISMATCH (expected 2)
    queried 3 -> 3104 reports bucket index 1   <-- MISMATCH (expected 3)

skipping broadcasts:
    queried 0..3 -> 3104, reports bucket index 0..3, all aligned

Worth noting byte 14 of a 31 04 reply carries its own bucket index, so you can see the shift directly instead of guessing at it.

What that breaks: _query_buckets ends up pairing bucket N's reply with N+1, and _setup_bucket / _switch_bucket pull response[14] out of a broadcast and decide they failed. The bulk transfer actually went fine, so the asset is sitting on the device, just never activated, and you get a black LCD. Since both of those are _LOGGER.error and not raised, set lcd screen gif exits 0 with literally no output unless you pass -v. Took me a while to figure that out. The broadcasts also eat the _read_until attempt budget, which I'm fairly sure is the AssertionError('missing messages (attempts=12, missing=1)') from #631.

I'm running basically your _read_skipping_broadcasts plus clear_enqueued_reports() before the handshake, with one extra bit:
_read_until and _read_until_first_match skip broadcast frames too without burning an attempt, otherwise 12 attempts are gone before the reply I want shows up. With that, set lcd screen gif works on the 2023 Elite on 2.1.0. 640x640, 35 frames, uploads and animates fine. That's the mode the docs and the (broken) tag both say isn't supported on 2.X. kraken3 and smart_device tests pass unchanged.

Caveat: one device, one firmware, so no idea about 300e / 3014 or whether 75 02 is the same prefix elsewhere. Also the ERROR lines come and go depending on queue state, the misalignment above is the reliable part. Can test patches on this hardware if useful.

PR #916, opened 2026-07-22
kraken3: match HID responses to the requests that asked for them

creativeheadz:
## The problem

`_write_then_read()` returns whatever report arrives next, with no check that it answers the command just sent:

```python
def _write_then_read(self, data):
    self._write(data)
    return self._read()
```

Two things routinely leave other reports in the queue: several call sites send with a bare `_write()` and never consume the reply, and the device also emits unsolicited status reports.

The two meet inside `_send_data()`:

```python
self._write([0x36, 0x02])          # end transfer — reply left queued
if not self._switch_bucket(...):   # reads that leftover instead
```

so `_switch_bucket()` decides success from byte 14 of the stale `0x37 0x02`. Tracing every exchange of a single `set lcd screen static` on a Kraken 2023 Elite (`1e71:300c`, firmware 1.1.0, Linux):

```
ok        cmd=0x32 0x01  expected=0x33 0x01  got=0x33 0x01  byte14=0x01
ok        cmd=0x36 0x01  expected=0x37 0x01  got=0x37 0x01  byte14=0x01
MISMATCH  cmd=0x38 0x01  expected=0x39 0x01  got=0x37 0x02  byte14=0x01
```

That byte happened to be `0x01`, so the bucket switch reported success without having been verified at all.

`_query_buckets()` reads the same blind way, and because all sixteen queries answer `0x31 0x04`, a single misread shifts the entire table — `_get_bucket_memory_offset()` then sizes a transfer using another bucket's address. In one capture the table came back as `b0@1507+8995` followed by every subsequent entry displaced by one slot.

The user-visible symptom is an LCD push that reports success and does not display, intermittently, with no error on any path — which makes it look like a panel quirk rather than a driver bug.

## The fix

Clear the queue before sending, then check the prefix of what comes back.

The clear is what actually fixes it. The prefix check is a backstop and **cannot** by itself disambiguate replies that share a prefix, as the bucket queries do — worth stating explicitly, since prefix matching alone looks sufficient and is not.

Reads that find no matching report fall back to the first report read, preserving the previous behaviour rather than raising.

## Tests

Added a regression test that queues a stale `0x37 0x02` whose byte 14 reads as success, ahead of a genuine `0x39 0x01` reporting failure, and asserts `_switch_bucket()` returns `False`. It fails on `main` with `assert True is False` and passes with this change.

This required making the mock faithful: it previously answered bucket-setup and bucket-switch with a zeroed `reply[0:2]` header that no real device sends, so those paths could not be tested at all. Prefixes are now set for the bucket-setup, bucket-switch, bucket-query and data-transfer commands.

Full suite: 518 passed, 2 skipped (both pre-existing). `black --check` clean.

## Scope

Independent of #915 and branched from `main`, so the two can land in either order. #915 needs a malformed serial descriptor to bite; this one affects every Kraken Z and Elite, since `_write_then_read()` is unvalidated for all of them.


giggio:
I just want to give my 2c.
I pulled both this patch and #915 together 2 days ago into my liquidctl installation and my Kraken 3 Elite is working much better. The device's screen would go black and it would cause issues that are not happening anymore.
I did not review the code, but behaviorally it is working much better.

Issue #907, opened 2026-07-05
find_liquidctl_devices() crashes with ValueError when a device without string descriptors is on the bus (e.g. NZXT bootloader 1e71:3011)


relative23:
### Is there already an issue for this?

- [x] There is no relevant open or closed issue for this

### Is this a request for support of a new device?

- [x] This is not a new device request

### Affected liquidctl version

liquidctl v1.15.0 (Linux-7.0.0-27-generic-x86_64, Ubuntu 26.04); the affected comparison is unchanged on current main

### Affected device

NZXT Kraken 2024 Elite RGB — but discovery breaks for all devices on the system

### Is the device supported by that liquidctl version?

- [x] The device is listead in the README as supported by my liquidctl version

### Describe the bug

While an NZXT Kraken is stuck in its bootloader mode (enumerates as `1e71:3011 "NZXT BOOTarea"`, no string descriptors), device discovery crashes entirely — for *all* devices, not just the affected one.

`KrakenZ3`'s bulk-device matching compares `serial_number` of every candidate USB device; reading the string descriptor of the bootloader device raises, and the `ValueError` escapes `find_liquidctl_devices()`. Unlike #618 / #645 (permission problems on the target device itself), this makes discovery unusable as long as *any* descriptor-less device is attached — monitoring tools lose access to every other supported device on the system, and the wedged device cannot even be *detected* gracefully.

How to get a Kraken into this state (sadly reproducible): see my comment on #774 — repeated image uploads with the stock error handling wedged the device into the bootloader twice.

### Suggested fix

Wrap the serial comparison in `try/except ValueError` and skip candidates whose descriptors are unreadable. Discovery then keeps working with the remaining devices, and a wedged device can be detected instead of taking discovery down with it. Happy to send a PR.

### Commands executed

Any operation that enumerates devices fails while a `1e71:3011` device is attached, e.g.:

```
$ liquidctl list
ValueError: The device has no langid (permission issue, no string descriptors supported or device error)
```

Expected: the bootloader device is skipped (or listed as unsupported) and all other devices remain usable.
Observed: discovery raises and nothing is listed at all. The same crash hits programmatic users of `find_liquidctl_devices()` — in our case a monitoring daemon's reconnect path.

### Output of all relevant commands with `--debug` flag

Captured from our daemon's log while the wedged device was on the bus (identical with the CLI); the relevant frames:

```
Traceback (most recent call last):
[...]

Issue #774, opened 2025-03-09
Failed to switch active bucket" error on NZXT Kraken Elite 2024 LCD update


kaan9700:
### Describe the bug

I am using a custom script that generates a 640x640 image for the LCD display of my NZXT Kraken Elite 2024 (v2) and uploads it via liquidctl. The image is an infographic showing temperatures. Therefore it runs in a loop that refreshes the image every n seconds.
From time to time 2 errors occur:

1. `failed to switch active bucket`:
The initial monitoring of the octopus is displayed for a short time (liquid temperature) 

2. `failed to setup bucket for data transfer`: 
The entire LCD display of the octopus goes black for a short time until it is replaced by the new image at the next period.

The errors are from the `kraken3.py` file, more precisely from the `_send_data` function.

### Commands executed

This is the code i am executing:
```
def set_kraken_screen(image_path):
    result = subprocess.run([
        "/home/user/liquidctl-venv/bin/liquidctl",
        "--match", "Kraken",
        "set", "lcd", "screen",
        "static", image_path
    ], capture_output=True, text=True)
    print(result)



def main():
    script_dir = os.path.dirname(os.path.abspath(__file__)) 
    image_path = os.path.join(script_dir, "infograph_custom.png")

    while True:
        cpu_usage = psutil.cpu_percent(interval=1.0)
        cpu_temp = get_cpu_temp()
        gpu_temp, gpu_usage = get_nvidia_stats()

        img = create_custom_layout(gpu_temp, cpu_temp, cpu_usage)
        img.save(image_path)

        try:
            set_kraken_screen(image_path)

        except Exception as e:
            print(">> Kraken failed to update")
            print(e)

        time.sleep(5)

```

### Output of all relevant commands with `--debug` flag

I started the script by executing this command:
```
sudo /home/user/liquidctl-venv/bin/python ~/monitoring/main.py
```

The logs looks like this:

```
[INFO] Starting data transfer on bucket index 3
...
[ERROR] Failed to switch active bucket
...
[ERROR] Failed to setup bucket for data transfer

### Affected device

NZXT Kraken Elite 2024

### Does your version of liquidctl support the device in question?

Yes, my version supports it

### Operating system and version

Ubuntu 24.04.2 LTS

### Installation method

Installation via GitHub Repository

### Version of liquidctl

liquidctl v1.15.0.dev16+ge1d0b60 (Linux-6.11.0-19-generic-x86_64-with-glibc2.39)

[...]

PR #927, opened 2026-09-11
kraken3: abort the upload when the device refuses the bucket setup

relative23:
`KrakenZ3._send_data` logs a refused bucket setup and then streams the whole image anyway. The data cannot land, the bucket switch fails afterwards, and `set_screen()` returns as if it had worked.

On a Kraken 2024 Elite RGB (`1e71:3012`) the firmware refuses the setup sporadically even with free image memory. Streaming regardless, under a periodic upload workload, is the situation in which our device wedged into its bootloader (`1e71:3011`) twice, recoverable only by cutting standby power (#774, #907).

This raises `ExpectationNotMet` before any bulk data is sent. The bucket-switch failure after a successful stream keeps its log line: nothing more is sent at that point, and making it raise would need a change to `MockKraken`'s switch reply.

**Tested:** `tests/test_kraken3.py` passes (31 passed, 1 pre-existing skip), one regression test added. On the 2024 Elite RGB (firmware 1.2.0) the same abort has been running in a `KrakenZ3` subclass since July in [relative23/kraken-lcd](https://github.com/relative23/kraken-lcd); refusals surface as exceptions the caller can handle instead of silent failed uploads.

**Not addressed:** the sporadic refusals themselves. Only the harmful reaction to them.

Refs #774


relative23:
One correction to the description: "firmware 1.2.0" is what liquidctl reads from the firmware-info report (bytes 0x11–0x13). NZXT CAM shows the same firmware as 1.2.12 (its update history: 1.2.1 → 1.2.8 → 1.2.12), I don't know which command CAM uses for the patch level. See also my update on #774: the abort makes a failed upload visible, it did not stop the wedges on this device.

